PR #6, opened 2015-04-28
evaluate core programs

amosr:


test/Icicle/Test/Arbitrary.hs
```diff
@@ -24,7 +26,7 @@ instance Arbitrary Attribute where
 
 instance Arbitrary DateTime where
   arbitrary =
-    DateTime <$> elements muppets -- FIX replace with an actual DateTime
+    dateOfYMD <$> oneof_vals [2010..2014] <*> oneof_vals [1..12] <*> oneof_vals [1..28]
```
FrontierOtr: If you have a look at DateTimeUtils in ivory you'll find a days number to date function which would be better than this (and cover all edges, leap years...).

amosr: oh - that's a much better idea. I'll hunt around

amosr:
Merging because there's an emoticon somewhere

PR #10, opened 2015-05-06
Rip out and generalise Exp

amosr:
Because the different languages will all probably have expressions, we might as well generalise Core.Exp.
It is now parameterised by the type of primitives, and type checking is parameterised by the "language fragment" it's in, which defines the types of primitives and which extra invariants need to be upheld.


doc/fragments.md
```diff
@@ -0,0 +1,90 @@
...
+In Core and Avalanche, all primitives must be fully applied.
+I don't imagine this is the case in Icicle, but it will be easy to eta expand on the way to Core.
+
+Some Core expressions allow functions at the top level: worker functions for stream filters and so on.
+But other than that, precomputations and postcomputations must be values, so functions aren't allowed there.
+By the time we get to Avalanche, all expressions should be values, not functions.
+
+It'd be super simple to disallow name shadowing for all stages, but I fear this might be burdensome for Icicle.
+
+Disallowing lets or requiring something like A-normal form would be other useful invariants.
```
markhibberd: :100:

src/Icicle/Avalanche/Eval.hs
```diff
@@ -0,0 +1,83 @@
...
+
+evalLoop
+        :: Ord n
+        => EvalPrim n p
+        -> Loop n p
+        -> (AccumulatorHeap n, XV.Heap n p)
+        -> AsAt (BubbleGumFact, BaseValue)
+        -> Either (RuntimeError n p) (AccumulatorHeap n, XV.Heap n p)
+evalLoop evalPrim  loop (ah, xh) input
+ = ...
```
markhibberd: The dots make it fast I guess :speedboat:

markhibberd:
Assuming there is a little bit more to make this compile? but on the surface this look good. 

It would be really nice to start getting more people across this and thinking about how it hangs together (this will help with review, make it tractable for moar people to help out later on, etc...). Given we are so busy at the moment, makes it a bit hard, but maybe we should come up with a plan to do an engineering talk in the first week of june. This would run through everything so far, where things are at, and where things are headed. Also, it might be nice to start raising issues for little tasks that aren't critical and you are leaving to the side for latter - the idea would be if we could build up a little pool of these, then I could get people to help out for a couple of hours here and there to work on these issues - hopefully this would help get people up to speed and let you focus on the important bits.

markhibberd:
:dromedary_camel:

amosr:
Yeah, still a fair bit to do to get executable code out, but the structure is there. Pretty printing the actual loop structure from Avalanche should be easy, but I still need to convert array and map folds and whatnot in the expressions to actual loops.

Doing a talk in June sounds good, and I'll create some tickets now before I forget them

amosr:
Oh, I misunderstood! Yeah, the evaluator doesn't compile yet, that was just a high-level sketch

PR #17, opened 2015-05-12
Add DateTime to language

amosr:
We can now implement "days since latest".

Some DateTime primitives such as days difference between two DateTimes.
Each stream now has the fact value and the fact date available.

For Core, only postcomputations are able to access the current date. For Avalanche, any part can access the current date, because window checks need the current date.

### DIFF
--- a/doc/core.md
+++ b/doc/core.md
@@ -67,7 +67,7 @@ All expressions here are worker functions, only able to access the precomputatio
 
 ```
 Stream  ::= Source                              : Stream (concrete feature type)
-          | Take   (Exp Int) (Stream a)         : Stream a
+          | SourceWindowed Days                 : Stream (concrete feature type)
           | Filter (Exp (a -> Bool)) (Stream a) : Stream a
           | Map    (Exp (a -> b))    (Stream a) : Stream b
 ```
@@ -75,7 +75,7 @@ Stream  ::= Source                              : Stream (concrete feature type)
 These are not strictly necessary from an expressiveness standpoint, as a fold on each of these could be rewritten as a fold on the original data.
 However, making these explicit makes composition a lot easier than with folds.
 
-I had Scan here before, but a) I'm not convinced it's necessary, and b) it complicates lookbehind computation.
+I had Scan and Take here before, but a) I'm not convinced they are necessary, and b) they complicate lookbehind computation.
 
 Similarly, Drop could be added, but it also complicates lookbehind.
 
@@ -106,6 +106,11 @@ Postcomputations
 Finally, the result of the reductions might need to be modified somehow.
 Postcomputations can be any expression, referring to precomputations, the result of reductions and previous postcomputations.
 Postcomputations cannot refer to streams.
+
+Postcomputations are also the only place that the current or snapshot date can be read and used.
+By disallowing reductions and stream transformers from using the current date, we ensure determinism:
+the stream transformers have the same output each time they are run on the same data.
+This is very important for computing lookbehind and so on, since we know that only same values that are needed today will be needed tomorrow.
 
 
 Types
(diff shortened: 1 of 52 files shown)


src/Icicle/Dictionary.hs
```diff
@@ -185,14 +204,43 @@ program_count_unique
...
+                        (X.XValue (T.MapT T.IntT T.IntT) $ N.VMap $ Map.empty)
+                        (N.Name "inp2"))]
+ , P.postdate   = Nothing
  , P.postcomps  = [(N.Name "size", X.XPrim (P.PrimFold (P.PrimFoldMap T.IntT T.IntT) T.IntT) @~ (lam T.IntT $ \a -> lam T.IntT $ \_ -> lam T.IntT $ \b -> a +~ b) @~ constI 0 @~ var "uniq")]
  , P.returns    = var "size"
  }
  where
   mT = T.MapT T.IntT T.IntT
+
+program_days_since_latest :: P.Program Text
```
charleso: Is there a test/validation for this somewhere? Apologies if that doesn't make sense yet.

amosr: mm no, not really.
maybe I'll try doing some of these
https://github.com/ambiata/ivory/blob/master/ivory-cli/src/test/bin/tests/window/input/dictionary.psv

test/Icicle/Test/Dictionary/Types.hs
```diff
@@ -0,0 +1,42 @@
...
+
+check_attributes (Dictionary attrs)
+ = and <$> mapM check attrs
+ where
+  check (_, ConcreteDefinition _)
+   = return True
+  check (_, VirtualDefinition virtual)
+   = check_virtual (program virtual)
+
+tests :: IO Bool
```
amosr: Is there a better way to do a non-property test?

charleso: > Is there a better way to do a non-property test?

[once](http://hackage.haskell.org/package/QuickCheck-2.8.1/docs/Test-QuickCheck-Property.html#v:once)

prop_foo = once . something something

amosr: neat, thanks

src/Icicle/Core/Exp/Combinators.hs
```diff
@@ -54,50 +61,57 @@ infixl 6 -~
...
+       -- in the expression and it isn't going to fit in memory.
+       -- In that case, we might as well die anyway.
+       v    = head_error free
+   in  XLam v t (f $ XVar v)
+
+ where
+  head_error (x:_)
+   = x
+  head_error []
+   = error "Icicle/Core/Exp/Combinators.hs: lam: this is impossible; taking the head of a nearly-infinite list should not fail"
```
amosr: I got rid of unsafePerformIO but had to add an error. arghgh

doc/core.md
```diff
@@ -142,7 +147,12 @@ Postcomputations cannot access stream values, but we can assume that all reducti
 What about "numflips" or a sum over all time?
 Since it's simply a reduce over all previous data, and the previous data isn't going to change, we should be able to just use the previous reduce value etc.
 
-So in summary, each fold reduction should output its current value as the lookbehind, and each latest reduction should output the values it used.
+There are three types of reductions: 
+
+- latest N: the used facts are stored
```
markhibberd: Not sure if it worth differentiating between `latest n` and `latest where <some condition>` (or is that what the N is for here?). This latest where case is interesting complicated because `<some condition>` might be something like `first`. That sounds a bit weird reading it back - but I think it makes sense - grab me tomorrow if it doesn't.

markhibberd:
:+1: Looks good.

doc/core.md
amosr: Ah no, I hadn't thought of that really - the n is just the number of elements here.
So for implementing something like "first" I was thinking we could use a fold over all time:
```fold(\acc elem -> Just elem) Nothing```
and then we'd store the value of that reduce ie (Just Foo) in the history.

For latest, I'm only allowing conditions like filters and windows right now. I had "take" like firstN before, but it actually complicates lookbehind a bit, and I figured it could be implemented with a fold.

markhibberd: Ok, makes sense.

PR #22, opened 2015-05-29
A few more examples

tmcgilchrist:
http://docs.datomic.com/query.html#outline-container-5-17-1

### DIFF
--- a/doc/icicle-concrete.md
+++ b/doc/icicle-concrete.md
@@ -292,13 +292,6 @@ Datomic/datalog (not sure if the :with is correct)
 [:find ?cat (count ?cat) .
  :with ?cat
  :in [[?cat :feat]]] 
- 
-;; rather than this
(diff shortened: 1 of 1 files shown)


doc/icicle-concrete.md
```diff
@@ -498,6 +491,12 @@ if (feat != feat_end) {
 return days;
 ```
 
+Datalog
+```
+;; Basically a the min of the max 5
```
amosr: ? why 5?

amosr:
:+1: more examples always more good

doc/icicle-concrete.md
tmcgilchrist: Might have misunderstood the example, isn't it asking for the nth day since the Earliest, where n is 5 here?

amosr: oh right, no - it's asking how many days ago the earliest entry is

PR #23, opened 2015-06-04
Icicle Source language initial parser and pretty printer

amosr:
We have a parser, a pretty printer and an evaluator for the source language.
I'd say that's a good start.


amosr:
yeah nah, that's a ridiculously large pull request; nobody's going to read that

markhibberd:
woot! Commits seem somewhat organized, I will try to read through as much as I can (but maybe not til the weekend.

charleso:
Looks amazing @amosr . :100:

Would love to do some hands-on with this and play with the results at some point. Even doing it with just `core` would be a good learning exercise. Not sure if we should organise a mini-workshop or something? I could even imagine a mini kata-like thing where you have to implement some expressions at each level.

PR #27, opened 2015-06-16
Icicle Source REPL

tranma:
REPL for Icicle surface language.

Commands:

  * `:load <filepath>`
  * `:set show-type` -- show the expression type
  * `:set show-core` -- show core output
  * `:set show-eval` -- show evaluated result

Given an Icicle expression, e.g. `feature salary ~> sum value`, show the result type and core output -- plus the result value if a data set is given as well.

### DIFF
--- a/main/icicle-repl.hs
+++ b/main/icicle-repl.hs
@@ -1,0 +1,114 @@
+{-# LANGUAGE ViewPatterns #-}
+
+import           Control.Monad
+import           Data.Text                   (Text)
+import qualified Data.Text                   as T
+import qualified System.Console.Haskeline    as HL
+import qualified Text.PrettyPrint.Leijen     as PP
+
+import           Icicle
+import qualified Icicle.Core.Program.Program as CP
+import qualified Icicle.Repl                 as SR
+import qualified Icicle.Source.Parser        as SP
+import           System.Environment
+
+main :: IO ()
+main = runRepl
+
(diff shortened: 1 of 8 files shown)


amosr:
:bangbang:

domdere:
you can put WIP in the title, and remove it when you think its ready

main/icicle.hs
```diff
@@ -47,8 +47,7 @@ main = getArgs >>= \args -> case args of
   _ ->
```
amosr: I think we can entirely replace main/icicle.hs with the repl at some stage

main/icicle-repl.hs
```diff
@@ -0,0 +1,113 @@
...
+data Set
+   = ShowType
+   | ShowCore
+   | ShowEval
+
+-- | REPL commands
+data Command
+   = CommandBlank
+   | CommandSet  Set
+   | CommandLoad FilePath
```
amosr: I don't know whether it makes sense, but I wonder whether it would be useful having any of this stuff in the library proper? would it make it easier to write eg web versions?

tranma: They can belong in the Repl library module, yeah

amosr:
:banana: :back:

tranma:
this depends on topic/icicle-initial now so I'll close this PR to make a new one

PR #28, opened 2015-06-17
Icicle conversion to core

amosr:


amosr:
gonna merge cause I can

PR #30, opened 2015-06-17
missing changes

tranma:
these are supposed to be merged into icicle-initial before that one goes into master, @amosr


charleso:
FYI we also have a very rudimentary cli testing "framework" (read: shell scripts) to test cli applications. Example if you're interested:

https://github.com/ambiata/arm/tree/master/test/cli

PR #31, opened 2015-06-18
Minor repl changes; minor conversion fixes

amosr:


main/icicle-repl.hs
```diff
@@ -39,7 +39,7 @@ runRepl
     loop state
       = do line <- HL.getInputLine "> "
            case line of
-             Nothing      -> loop state
+             Nothing      -> return ()
```
tranma: with this, can `CommandBlank` (where user doesn't input anything) ever happen? if not I'll remove the case.

amosr: hmm, I think empty line will still do CommandBlank

amosr: the "Nothing" case is for EOF by the looks of it

tranma: cool, I'll leave CommandBlank in then.

see https://github.com/ambiata/icicle/pull/32

tranma:
woah, we can have iREPL history too!!! let's?

amosr:
iREPL history? I don't know what you mean

tranma:
like ghci history, you can get commands from past sessions

amosr:
if that's easy to do, it'd be super useful

src/Icicle/Source/ToCore/Base.hs
```diff
@@ -91,6 +94,77 @@ freshly f
...
+   | anyPossibles
+   , Definitely <- universePossibility $ universe returns
+   = return $ mkSome $ X.makeApps f args
+   -- Otherwise, we have seen no possibles or the raw function
+   -- itself returns possibles, so we can leave the return as-is.
+   | otherwise
+   = return $          X.makeApps f args
+
+  -- For each entry of xts..
+  go ((x,u):xts') args anyPossibles
```
amosr: am I a bad functional programmer because I wrote this with recursion instead of a "fold"?

amosr: (I don't actually care)

tranma: a fundamentalist functional programmer!

amosr:
gonna merge this. there's some cool stuff here: converting from Icicle Source to Core mainly

PR #33, opened 2015-06-18
WIP - small step evaluator

tranma:
small step dynamics for Core

  * still fully evaluates primitives, but we need to pass the return type along
  * otherwise, standard call-by-value

### DIFF
--- a/src/Icicle/Common/Exp/Eval/Small.hs
+++ b/src/Icicle/Common/Exp/Eval/Small.hs
@@ -1,0 +1,90 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Icicle.Common.Exp.Eval.Small
+
+where
+
+import qualified Data.Map as M
+
+import Icicle.Internal.Pretty
+import Icicle.Common.Base
+import Icicle.Common.Type
+import Icicle.Common.Value
+import Icicle.Common.Exp.Exp
+import Icicle.Common.Exp.Eval (EvalPrim)
+import qualified Icicle.Common.Exp.Eval as Big
+
+import P
+
+import Data.Either.Combinators (mapLeft)
+
+
+data RuntimeError n p
+  = RuntimeErrorPrimBadArgs p [Value n p]
+  | RuntimeErrorUnboundVar  (Name n)
+  | RuntimeErrorPrimBad p
+  | RuntimeErrorPrimEval (Big.RuntimeError n p)
+  deriving (Show, Eq)
+
+--------------------------------------------------------------------------------
+
+-- | Small step evaluation. Call-by-value for simplicity.
+--
+eval :: Eq n
+     => EvalPrim n p -> Exp n p
+     -> Either (RuntimeError n p) (Exp n p)
+eval evalPrim xx = case xx of
+  -- * Variables must have been substituted away.
+  XVar n
+    -> Left (RuntimeErrorUnboundVar n)
+
+  -- * Values are fully evaluated.
+  XValue{}
+    -> return xx
+
+  -- * App where left is lambda and right is value: substitute.
+  XApp (XLam n t1 e) (XValue t2 v)
+    | t1 == t2
+    -> subst n v t1 e
+
+  -- * App where left is lambda: evals right.
+  XApp e1@XLam{} e2
+    -> fmap (XApp e1) (go e2)
+
+  -- * App where left is other exps: evals left.
+  XApp e1 e2
+    -> fmap (flip XApp e2) (go e1)
+
+  -- * Primitives with no arguments.
+  XPrim p
+    -> do x <- mapLeft RuntimeErrorPrimEval $ evalPrim p []
+          v <- getBaseValue (RuntimeErrorPrimBad p) x
+          t <- typeOfValue  (RuntimeErrorPrimBad p) x
+          return (XValue t v)
+
+  -- * Lambdas cannot go any further.
+  XLam{}
+    -> return xx
+
+  -- * Let where definition is a value: substitute.
+  XLet n (XValue t v) e2
+    -> subst n v t e2
+
+  -- * Let where definition is something else.
+  XLet n e1 e2
+    -> do e <- go e1
+          return (XLet n e e2)
+
+  where go = eval evalPrim
+
+subst :: Eq n
+      => Name n -> BaseValue -> ValType -> Exp n p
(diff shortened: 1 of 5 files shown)


src/Icicle/Common/Exp/Eval/Small.hs
```diff
@@ -0,0 +1,90 @@
...
+
+  -- * App where left is lambda: evals right.
+  XApp e1@XLam{} e2
+    -> fmap (XApp e1) (go e2)
+
+  -- * App where left is other exps: evals left.
+  XApp e1 e2
+    -> fmap (flip XApp e2) (go e1)
+
+  -- * Primitives with no arguments.
```
amosr: Icicle.Common.Exp.Compounds.takePrimApps might be useful for primitives with arguments

src/Icicle/Common/Exp/Eval/Small.hs
```diff
@@ -0,0 +1,90 @@
...
+    -> subst n v t e2
+
+  -- * Let where definition is something else.
+  XLet n e1 e2
+    -> do e <- go e1
+          return (XLet n e e2)
+
+  where go = eval evalPrim
+
+subst :: Eq n
```
amosr: Ah, also in Icicle.Common.Exp.Compounds, there's a "subst" and "substMaybe".
"subst" needs a fresh name generator for capture avoiding substitution, so you'd need to put eval in the "FreshT" monad transformer.
or you can use "substMaybe" which only performs substitution if no renaming is necessary - that would be fine too. and if renaming is necessary, the small step just wouldn't take a step

test/Icicle/Test/Core/Exp/Eval.hs
```diff
@@ -44,14 +45,28 @@ prop_prefixlet =
...
+-- =====================
+
+prop_small_transition =
+ withTypedExp $ \x _ -> iterateEval x == hush (eval0 evalPrim x)
+ where iterateEval x
+         | Right z  <- ES.eval evalPrim x, z /= x = iterateEval z
+         | XValue{} <- x = hush $ eval0 evalPrim x
+         | otherwise = Nothing
+       hush (Left _)  = Nothing
+       hush (Right x) = Just x
```
charleso: We have our own [rightToMaybe](https://github.com/ambiata/p/blob/265cd5c3d5a91a332aedfd950d80bc1f9e9fc5f2/src/P/Either.hs#L17) if you prefer. :)

domdere: or [`head`] (https://github.com/ambiata/p/blob/master/src/P/Foldable.hs#L16-L17) even!

src/Icicle/Common/Exp/Eval/Small.hs
```diff
@@ -0,0 +1,90 @@
...
+
+-- | Small step evaluation. Call-by-value for simplicity.
+--
+eval :: Eq n
+     => EvalPrim n p -> Exp n p
+     -> Either (RuntimeError n p) (Exp n p)
+eval evalPrim xx = case xx of
+  -- * Variables must have been substituted away.
+  XVar n
+    -> Left (RuntimeErrorUnboundVar n)
```
amosr: I think this should just return "Right xx" - so we can partially evaluate open expressions

src/Icicle/Common/Exp/Eval/Small.hs
```diff
@@ -0,0 +1,90 @@
...
+    -> return xx
+
+  -- * App where left is lambda and right is value: substitute.
+  XApp (XLam n t1 e) (XValue t2 v)
+    | t1 == t2
+    -> subst n v t1 e
+
+  -- * App where left is lambda: evals right.
+  XApp e1@XLam{} e2
+    -> fmap (XApp e1) (go e2)
```
amosr: Argh, I made a mistake and asked you to implement small step evaluation, but it turns out what I really want is a partial evaluator.
Whether it's small step or big step doesn't really matter after all, and evaluation order doesn't matter, but it should go down and perform any simplifications like beta reduction and primitives applied to just XValues.
(We don't have recursion or fixpoints so we can afford to evaluate anything)

tranma: sounds like we just want simplifier then?

so, if an expression can be evaluated in several ways, e.g. `plus# ((λ x. x) 1) ((λ x. x) 2)`, should the user be able to specify which redex to use, or do we just apply beta reduction as much as possible?

src/Icicle/Common/Exp/Eval/Small.hs
tranma: oh okay, I was dreading having to think about that. Cool beans, I'll probably use FreshT :+1:

src/Icicle/Common/Exp/Eval/Small.hs
tranma: does HERMIT do this somehow?

amosr: yeah, I think just do it all. actually we already have beta reduction implemented (Icicle.Common.Exp.Simp.Beta), so prim-of-XValue and Core program specific things are probably all that need doing

amosr: I think hermit has some idea of a focus, so you can say "go to application of plus#, then second arg, then simp" but I'm not sure

tranma: woohoo, Coq tactic style! (simpl, simpl, simpl!!)

src/Icicle/Common/Value.hs
```diff
@@ -32,6 +37,35 @@ getBaseValue :: e -> Value n p -> Either e BaseValue
...
+  VBool _      -> return BoolT
+  VDateTime _  -> return DateTimeT
+
+  VPair v1 v2  -> PairT   <$> typeOfBaseValue e v1 <*> typeOfBaseValue e v2
+
+  VSome v1     -> OptionT <$> typeOfBaseValue e v1
+  VNone        -> OptionT <$> pure IntT -- I don't know, do we have bottoms?
+
+  VArray vs
+   | ts <- L.map (typeOfBaseValue e) vs, t:_ <- ts, allSame ts
```
amosr: ```{-# LANGUAGE -XPatternGuards #-}```

tranma:
will be subsumed by a simplifier PR

PR #36, opened 2015-06-21
simple Core simplifier

tranma:
a very boring core simplifer, does some constant folding and beta whenever possible.

todo:
- I'm not too sure about just simplifying each bindings in Core Program (see below), that doesn't sound too helpful
- perhaps we want to run the simplifier before generating Avalanche
- same with pretty printing Core (for REPL)

### DIFF
--- a/src/Icicle/Core/Exp/Simp.hs
+++ b/src/Icicle/Core/Exp/Simp.hs
@@ -45,6 +45,9 @@ simpX isValue = go
         | Just (p, as) <- takePrimApps xx
         -> makeApps (XPrim p) (fmap go as)
 
+      XApp p q
+        -> XApp (go p) (go q)
+
       XLam n t x1
         -> XLam n t (go x1)
 
--- a/src/Icicle/Core/Exp/Simp.hs
+++ b/src/Icicle/Core/Exp/Simp.hs
@@ -43,10 +43,17 @@ simpX isValue = go
       -- * beta reduce primitive arguments
       XApp{}
         | Just (p, as) <- takePrimApps xx
-        -> makeApps (XPrim p) (fmap beta as)
+        -> makeApps (XPrim p) (fmap go as)
 
-      -- * leaves everything else alone
-      _ -> xx
+      XLam n t x1
+        -> XLam n t (go x1)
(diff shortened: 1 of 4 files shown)


src/Icicle/Core/Exp/Simp.hs
```diff
@@ -0,0 +1,139 @@
...
+        -> bool2 args (||)
+      M.PrimLogical M.PrimLogicalNot
+        -> bool1 args not
+
+      -- * leaves baked-in constants and datetime alone
+      M.PrimConst    _ -> Nothing
+      M.PrimDateTime _ -> Nothing
+
+    bool1 args f
+      | [(_, VBool x)] <- args
```
amosr: partial pattern match?! this is very naughty.
we should probably return an eval error in this case, but I think at the very least returning the original expression would be better because later typechecking or real eval would give a meaningful error

src/Icicle/Core/Program/Simp.hs
```diff
@@ -0,0 +1,64 @@
...
+simpProgram p
+  = do pres <- forall simp (precomps p)
+       poss <- forall simp (precomps p)
+       ss   <- forall simpStream (streams p)
+       rs   <- forall simpReduce (reduces p)
+       return p { precomps  = pres
+                , streams   = ss
+                , reduces   = rs
+                , postcomps = poss}
+  where forall f = sequenceA . fmap (sequenceA . fmap f)
```
amosr: mapM f ?
/codegolf

src/Icicle/Core/Exp/Simp.hs
tranma: it's not eval, just rewriting so I wouldn't want to give an EvalRuntimeError or similar. I think just not rewrite the original expression is fine too

amosr: yep, I agree

amosr:
:seedling: shall I merge this?

tranma:
should be fine, you wanna plug it in core -> avalanche conversion yourself?

amosr:
yep, will do

PR #38, opened 2015-06-25
Double-sided windows

amosr:
It might be useful to express things like "window newer than 30 days and older than 5 days".
Evaluation of history still needs to be fixed, as in the example above the history must include entries newer than 5 days, but the value and any folds do not use those entries.


amosr:
:+1: from myself

PR #42, opened 2015-07-03
Topic/conversion

amosr:
Simplifier tweaks, Array_map primitive


src/Icicle/Core/Exp/Simp.hs
```diff
@@ -62,125 +60,27 @@ simpX isValue = go
...
-      = let t = functionReturns (M.typeOfPrim pp)
-            v = f x y
-        in  return $ XValue t (VInt v)
-      | otherwise = Nothing
-
-
-takeValue :: Exp n p -> Maybe (ValType, BaseValue)
-takeValue (XValue a b) = Just (a, b)
-takeValue _            = Nothing
+simpP :: Ord n => Prim -> [Value n Prim] -> Maybe (C.Exp n)
```
tranma: originally this primitive simplifier `simpP` is "total", the Maybe just means it doesn't have anything to simplify so the exp simplifier `simpX` can return the original expression. Using the evaluator here instead of rewriting means ignoring runtime error (just return Nothing as if we have nothing to simplify), right? That should be ok, I think....

amosr: Yeah - a runtime error in this case should only really happen (on a well typed program) if we try something like evaluating
```
Array_map (\x . x + some_outside_thing) [1,2,3]
```
 which will give a runtime error of "no such variable some_outside_thing".

A total rewrite / simplifier would be ideal, but we'd need some way of only defining the evaluation rules once - otherwise there's too much risk of them not lining up (and just too much avoidable effort to add new primitives)

amosr: Actually the main advantage that I see of doing "real partial evaluation" rather than calling evalPrim would be that it could rewrite the above Array_map. Maybe evalPrim could be generalised to support both cases, I'm not sure

amosr:
struct fields work!
```
> feature injury ~> group location ~> sum severity
- Result:
[homer, [(arm,4),(head,3),(torso,4)]]
```

tranma:
oh wow :trophy:

amosr:
the stars say yes :8ball:

PR #43, opened 2015-07-14
Exception handling and conversion from Source to Core.

amosr:
Adding exceptions makes the Core language a little less "clean", but it makes the conversion so much simpler.
Conversion works on let-folds inside groups, and "latest 5 ~> value" returns an array of the 5 latest values now.


amosr:
:secret: :mailbox:

amosr:
I deleted this but had something committed locally so probably shouldn't have

PR #47, opened 2015-07-15
Topic/conversion

amosr:


doc/java.md
```diff
@@ -0,0 +1,298 @@
...
+    }
+  }
+  read conv$5 = acc$conv$5;
+  return conv$5;
+}
+```
+
+I'm imagining the Java would be something like this.
+```
+void Compute(IcicleMagic icicle)
```
amosr: wouldn't mind someone's opinion on whether this looks reasonable for java equivalent feature @charleso

doc/java.md
```diff
@@ -0,0 +1,298 @@
...
+
+  }
+  read conv$4 = acc$conv$4;
+  return conv$4;
+}
+```
+
+Java:
+```
+void Compute(IcicleMagic icicle)
```
amosr: again, feedback on this java pseudocode would be nice

doc/java.md
```diff
@@ -0,0 +1,298 @@
...
+    // return conv$5;
+    icicle.Output<Integer>("CountAboveTen", new Integer(conv_5));
+}
+```
+
+And the very very rough bytecode.
+This is not quite right syntax but the idea is there.
+It could also be improved a bit like not storing variables before using them just once.
+```
+// void Compute(IcicleMagic icicle)
```
amosr: the bytecode probably isn't worth reading. it was more for myself

doc/java.md
```diff
@@ -0,0 +1,298 @@
...
+  return conv$5;
+}
+```
+
+I'm imagining the Java would be something like this.
+```
+void Compute(IcicleMagic icicle)
+{
+    // gen$date = DATE
+    DateTime gen_date = icicle.SnapshotDate();
```
charleso: FWIW this took me a few seconds to parse because the convention is that functions start with lower case, so I thought you were creating a new `SnapshotDate`.

doc/java.md
```diff
@@ -0,0 +1,298 @@
...
+            if (anf_17 < 7) {
+                icicle.KeepFactInHistory();
+            }
+        }
+    }
+
+    // read conv$4 = acc$conv$4;
+    int conv_4 = acc_conv_4
+    // return conv$4;
+    icicle.Output<Integer>("CountLastWeek", new Integer(conv_4));
```
charleso: You probably know that the `new Integer(...)` here is boxed and on the heap. The only other option is we generate a extra/specific data type that has a `int countLastWeek` field, which then also knows how to serialize/deserialize to thrift/whatever. Given that we already do the boxed map thing in scala right now it's not a problem, but just thinking about moar optimisations.

michaelneale: Welcome to java have a nice stay.

charleso:
Java seems pretty solid. :+1: :+1: :+1:

doc/java.md
michaelneale: yes Integer is at least an extra 4 bytes over 4 bytes of int if it isn't eliminated at compile time (not sure when that happens, or doesn't) - so just something for memory usage considerations.

amosr: yeah, good plan. I don't think this particular one is a big deal because it's outside the loop, but the one above (line 58) is probably a lot more important.
I'm guessing accessing an int field from an ivory struct will allocate, but maybe we can have specialised versions for accessing int struct fields etc.

charleso: > I'm guessing accessing an int field from an ivory struct will allocate

The values from the structs are allocated when we deserialise from thrift, so by the time it reaches icicle there's no allocations just to read the value, only to write back again. I think what you have is fine - and matches what we have at the moment.

doc/java.md
michaelneale: Looks good to me. Charles pointed out the convention of lower camel case for method names which is different to what you have. (Curiously C# uses upper camel case for method names, mostly out of spite)

(and looks like you have switched $ for _ which seems more idiomatic)

doc/java.md
michaelneale: And short lived heap things are swept out of Eden GC region really quickly so in practice there can be very little impact sometimes

java/icicle/MemoryState.java
```diff
@@ -0,0 +1,186 @@
+import java.util.*;
+import java.time.LocalDate;
+
+class MemoryState<T> implements IcicleState<T>
```
amosr: writing java is really hard

java/icicle/MemoryState.java
```diff
@@ -0,0 +1,186 @@
...
+        for (ArrayList<Map.Entry<Integer,Object>> ixs : _latests.values()) {
+            for (Map.Entry<Integer,Object> kv : ixs) {
+                _kept_in_history.add(kv.getKey());
+            }
+        }
+
+        _latests.clear();
+        _outputs.clear();
+
+        _read_from_history = (HashSet<Integer>)_kept_in_history.clone();
```
amosr: "Note: java/icicle/MemoryState.java uses unchecked or unsafe operations."

charleso:
```java
_read_from_history = new HashSet<Integer>(_kept_in_history)
```

java/icicle/MemoryState.java
michaelneale: you should try writing bash that writes php (I still stand by my decision, it was more pleasant than writing php directly).

amosr: oh man. I wrote a php backend for the Disciple compiler. I'm still waiting for the right opportunity to really use it though
https://github.com/DDCSF/ddc/blob/master/packages/ddc-core-babel/DDC/Core/Babel/PHP.hs

amosr:
hacky and unfinished, but the idea is there:
```java
> feature salary ~> filter value > 10 ~> sum value / count

- Java:
class Feature {
  public void compute(IcicleState<Integer> icicle) {
    int gen$date = icicle.now();

     {
      int ACCUMULATOR$acc$conv$5 = 0;
      int ACCUMULATOR$acc$conv$9 = 0;
      icicle.startNew();
      while (icicle.nextRow()) {
        Pair<Integer, Day> gen$fact = icicle.currentRow();
         {
          int anf$3 = gen$fact.fst();
          if ((anf$3 > 10)) {
            int acc$conv$5 = ACCUMULATOR$acc$conv$5;
            ACCUMULATOR$acc$conv$5 = (acc$conv$5 + anf$3);
            int acc$conv$9 = ACCUMULATOR$acc$conv$9;
            ACCUMULATOR$acc$conv$9 = (acc$conv$9 + 1);
          }
          else {

          }

        }

      }

      int conv$5 = ACCUMULATOR$acc$conv$5;
      int conv$9 = ACCUMULATOR$acc$conv$9;
      int conv$11 = (conv$5 / conv$9);
      return conv$11
    }

  }

}
```

charleso:
:godmode:

jystic:
You probably know already, but beware that generics are boxed :)

The memory layout of your `ArrayList<Pair<Integer, Day>>` will be:

```
ArrayList<>
+---------------------+---------+
| mark word           | 8 bytes |
| class reference     | 8 bytes |
+---------------------+---------+
| n_elems : int       | 4 bytes |
| padding             | 4 bytes |
| elems   : Object[]* | 8 bytes |
+---------------------+---------+

Object[]
+-----------------+------------------+
| mark word       | 8 bytes          |
| class reference | 8 bytes          |
| length : int    | 4 bytes          |
| padding         | 4 bytes          |
+-----------------+------------------+
| data            | 8 bytes * length |
+-----------------+------------------+

Pair<>
+-----------------+---------+
| mark word       | 8 bytes |
| class reference | 8 bytes |
+-----------------+---------+
| _fst : Object*  | 8 bytes |
| _snd : Object*  | 8 bytes |
+-----------------+---------+

Integer
+-----------------+---------+
| mark word       | 8 bytes |
| class reference | 8 bytes |
+-----------------+---------+
| value : int     | 4 bytes |
| padding         | 4 bytes |
+-----------------+---------+

Day
+-----------------+---------+
| mark word       | 8 bytes |
| class reference | 8 bytes |
+-----------------+---------+
| day : int       | 4 bytes |
| padding         | 4 bytes |
+-----------------+---------+
```

So storing nothing in your `ArrayList` will take 56 bytes, and storing 16 pairs will take 56 + 2560 bytes.

amosr:
oh, yeah. so hopefully we'll be able to figure a way around this, maybe specialised versions of IcicleState for primitives.
I'll get rid of the Day class later and replace it with an unboxed int, but for now it's just so I don't mix up ints and ints accidentally.
the Pair<T,Day> is no good, but I'll have to do a bit of rejigging before I can get rid of that.

for now, I'm just doing the simplest possible thing that could possibly work!

amosr:
And the generated Java code actually compiles, except for the "return" statement.
It is a shame about the ``Latest<Pair…>`` but I can fix that.

```
feature salary 
~> let x = 5
~> latest 3 
~> (sum value + 1 + count + x, x)
==>
```

```java
class Feature
{
  public void compute(IcicleState<Integer> icicle)
  {
    int gen$date = icicle.snapshotDate();

    Latest<Pair<Integer, Integer>> ACCUMULATOR$acc$conv$2 = icicle.<Pair<Integer, Integer>>makeLatest(3);
    icicle.startHistory();
    while (icicle.nextRow())
    {
      int elem$gen$fact = (icicle.currentRow()).intValue();
      int elem$gen$date = icicle.currentRowDate();
      Pair<Integer, Integer> gen$fact = Pair.create(Integer.valueOf(elem$gen$fact), Integer.valueOf(elem$gen$date));
      icicle.pushLatest(ACCUMULATOR$acc$conv$2, gen$fact);
    }
    icicle.startNew();
    while (icicle.nextRow())
    {
      int elem$gen$fact = (icicle.currentRow()).intValue();
      int elem$gen$date = icicle.currentRowDate();
      Pair<Integer, Integer> gen$fact = Pair.create(Integer.valueOf(elem$gen$fact), Integer.valueOf(elem$gen$date));
      icicle.pushLatest(ACCUMULATOR$acc$conv$2, gen$fact);
    }
    ArrayList<Pair<Integer, Integer>> conv$2 = icicle.readLatest(ACCUMULATOR$acc$conv$2);
    int ACCUMULATOR$flat$0$simp$20$simp$22$simp$24$simp$26 = 0;
    int ACCUMULATOR$flat$0$simp$20$simp$22$simp$25$simp$30 = 0;
    for (int flat$1 = 0; flat$1 < conv$2.size(); flat$1++)
    {
      int flat$0$simp$36$simp$38$simp$40$simp$42 = ACCUMULATOR$flat$0$simp$20$simp$22$simp$24$simp$26;
      int flat$0$simp$36$simp$38$simp$41$simp$46 = ACCUMULATOR$flat$0$simp$20$simp$22$simp$25$simp$30;
      Pair<Integer, Integer> flat$2 = conv$2.get(flat$1);
      int simp$10 = (flat$2.fst()).intValue();
      int simp$11 = flat$0$simp$36$simp$38$simp$40$simp$42 + simp$10;
      int simp$14 = flat$0$simp$36$simp$38$simp$41$simp$46 + 1;
      ACCUMULATOR$flat$0$simp$20$simp$22$simp$24$simp$26 = simp$11;
      ACCUMULATOR$flat$0$simp$20$simp$22$simp$25$simp$30 = simp$14;
    }
    int flat$0$simp$52$simp$54$simp$56$simp$58 = ACCUMULATOR$flat$0$simp$20$simp$22$simp$24$simp$26;
    int flat$0$simp$52$simp$54$simp$57$simp$62 = ACCUMULATOR$flat$0$simp$20$simp$22$simp$25$simp$30;
    int anf$28 = flat$0$simp$52$simp$54$simp$56$simp$58 + 1;
    int anf$29 = anf$28 + flat$0$simp$52$simp$54$simp$57$simp$62;
    int anf$30 = anf$29 + 5;
    Pair<Integer, Integer> conv$4 = Pair.create(Integer.valueOf(anf$30), Integer.valueOf(5));
    return conv$4
  }
}
```

michaelneale:
yes it is returning something when it expects void.

amosr:
gonna merge :police_car:

PR #52, opened 2015-07-21
Replace time/date

tranma:
replace the combo of `time` and `dates` libraries with `thyme`. It should build on 7.10 now?


amosr:
perhaps we could actually use Delorean for this, instead?

tranma:
this is so far behind, does it still apply? in any case the `dates` library should really be replaced

markhibberd:
:+1: For thyme. Don't think delorean really captures the right type of thing.

amosr:
yep, if you want to do this go for it. if you don't want to, it can certainly wait. @HuwCampbell is touching this date stuff right now though, so it might be worth throwing away this branch and starting again

nhibberd:
@amosr @HuwCampbell @tranma Is this still relevant?

amosr:
@nhibberd nah I think I'll just delete it

PR #53, opened 2015-07-24
Topic/conversion

amosr:


java/icicle/Pair.java
```diff
@@ -15,4 +17,27 @@ public Pair(A a, B b)
     {
         return new Pair<A,B>(a, b);
     }
+
+    public String toString()
+    {
+        return "(" + _fst.toString() + ", " + _snd.toString() + ")";
+    }
+
+    public boolean equals(Object o)
```
amosr: @charleso would appreciate your thoughts on this. 
"icicle/Pair.java uses unchecked or unsafe operations"
is there any way to implement equals without "unsafe operations"?

charleso: http://stackoverflow.com/questions/1130433/how-to-avoid-unchecked-conversion-warning-in-java-if-you-use-legacy-libraries

Yah Java

src/Icicle/Common/Type.hs
```diff
@@ -43,24 +45,26 @@ import              P
 import qualified    Data.Map as Map
 
 
--- | Real values.
+-- | Types of real values.
 -- No functions here, because we don't want higher order functions in the generated code.
 -- This restriction should simplify code generation, because we won't need to
 -- deal with lambda lifting arbitrary functions.
-data ValType =
+data ValType' t =
```
amosr: this is hell.
I have two different kinds of types: these, which are used in Core, but in Source I need more stuff like type variables, foralls, etc.
so I'm trying to "untie the recursive knot" here so I can reuse this for both cases, but it's just not particularly fun.
any suggestions of a better way to do this?

amosr:
we are getting closer.

amosr:
We have functions, we have types, we have passing tests!

PR #58, opened 2015-08-05
Dictionary doc proposal

HuwCampbell:
@charleso @raronson You guys might be interested in this.


domdere:
Looks interesting :+1:

tranma:
will the output dictionary be more or less the same...?

domdere:
@tranma there might not be a need to distinguish between virtual and normal features in the output dictionary, and it will have to be readable by R, so it might be more convenient for the output dictionary to  be a similarly structured JSON document, but looks like it has to be discussed

HuwCampbell:
There is an R toml library, so potentially something similar could be used.

src/Icicle/Dictionary/Parse/Prisms.hs
```diff
@@ -0,0 +1,82 @@
...
+import           Data.Time
+
+import qualified Text.Parsec.Pos as Pos
+
+import           GHC.Float
+
+import           Icicle.Dictionary.Parse.Types
+
+
+_NTValue :: Prism' Node TValue
```
amosr: can you use template haskell to generate this? (I am not suggesting you change it, I'm just curious)

domdere: Its probably `makePrisms` friendly, not going to going to weigh in on the `TemplateHaskell` issue at all though

src/Icicle/Dictionary/Parse/Types.hs
```diff
@@ -0,0 +1,91 @@
...
+emptyNTable = NTable M.empty
+
+
+-- | Inserts a table ('Table') with name ('[Text]') which may be part of
+-- a table array (when 'Bool' is 'True') into a 'Table'.
+-- It may result in an error ('Text') on the 'Left' or a modified table
+-- on the 'Right'.
+insert :: ([Text], (Node, Pos.SourcePos)) -> Table -> Either Text Table
+insert ([], _)           _    = error "FATAL: Cannot call 'insert' without a name."
+insert (_ , (NTValue _, _)) _ = error "FATAL: Cannot call 'insert' with a TValue."
```
amosr: whoa! if these are "impossible" errors, I'd like some explanation of why they won't happen in the comments

domdere: Is it possible to construct a more appropriate type so the errors aren't necessary at all?

HuwCampbell: This was from htoml, I will put it in the left so we don't get anything super silly happening.

amosr:
:+1: a cli test or something would be good though

src/Icicle/Dictionary/Parse/Types.hs
domdere: It looks like the types were poorly designed, the `Node` type represents some final representation, while in this context it should be a type that represents constructs that can be parsed at the top level as a unit from the config file (I.e only Tables, which may or not have table/table array header).

but the author was too lazy to separate them out properly

amosr: I agree to some extent, but I don't think you would gain terribly much by separating it out. the parser has to check for malformed input and error at some stage, regardless

src/Icicle/Dictionary/Parse/Prisms.hs
domdere: incidentally, this was duplicated from here: https://github.com/ambiata/x/blob/master/x-htoml/src/X/Text/Toml.hs

so this was probably faster than doing the template haskell

HuwCampbell: Reason being that the types have changed because we need the position, so the other prism's can't apply anymore.

src/Icicle/Dictionary/Parse/Types.hs
domdere: yeah it should be doing it at the parse stage though, as early as possible.

don't get me wrong what @HuwCampbell is suggesting is OK, given that he is modifying/migrating some pre-existing base, but if this was being designed from the ground up, it would be better to get all of that out of the way in the parsing stage which has an `EitherT` specifically for "malformed input" errors.  once its parsed you want to be dealing wiht malformed input as little as possible

domdere: i realise there can be some real finicky manifestations of "malformed input" that can creep through still, but these particular errors can definitely be caught at the parsing stage quite easily and naturally with applicative parsers

HuwCampbell:
I think I have to close this PR and reopen it onto the conversion branch.

amosr:
I just merged this into topic/conversion

amosr:
OK if I delete this branch? I guess I could have deleted it since I merged it, but I didn't want to step on toes

PR #62, opened 2015-08-18
Annotate feature not found errors

HuwCampbell:
I was loading items from the dictionary and there wasn't a line error when the feature wasn't found.
Hence this.


amosr:
fantastic :bomb:

PR #65, opened 2015-08-21
Inline functions when loading the dictionary

HuwCampbell:
The functions sourceCheck and sourceInline are more or less taken from Repl, so should probably be unified somewhere else.

This makes the virtual definitions the dictionary produces be "complete", in that they don't need any more actions performed on them before they be evaluated with the source evaluator or compiled to core.

### DIFF
--- a/src/Icicle/Storage/Dictionary/Toml.hs
+++ b/src/Icicle/Storage/Dictionary/Toml.hs
@@ -18,8 +18,11 @@ import qualified Icicle.Source.Type                 as ST
 import qualified Icicle.Source.Checker.Checker      as SC
 import qualified Icicle.Source.Checker.Error        as SC
 import qualified Icicle.Source.Checker.Function     as SC
+import qualified Icicle.Source.ToCore.Context       as SC
 
+import qualified Icicle.Source.Query                as SQ
 import qualified Icicle.Source.Parser               as SP
+import qualified Icicle.Source.Transform.Inline     as STI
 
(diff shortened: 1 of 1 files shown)


src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -102,20 +105,19 @@ loadDictionary' parentFuncs parentConf parentConcrete fp
...
-       )
+    <- (\(a, q) ->
+          (<$>)
+         (\(q', _) -> DictionaryEntry a (VirtualDefinition (Virtual q')))
+        $ (=<<)
+         (\q' -> sourceCheck d'' q' )
+        $ (<$>)
+         (\(q', _) -> sourceInline d' q')
+         ( sourceCheck d'' q )
+       ) `traverse` virtualDefinitions'
```
amosr: I think this would be clearer with explicit do notation or something

tranma: @HuwCampbell plays master-level golf

domdere: lol

amosr: not in MY backyard

HuwCampbell: Ok, ok.

amosr:
groovy :purple_heart:

PR #66, opened 2015-08-21
Updated cabal-extras

jystic:


amosr:
whoa (whoa) :whoa:

jystic:
What's this all about:
```bash
The Glorious Glasgow Haskell Compilation System, version 7.6.3
```

domdere:
I think thats fine, `7.8.3` gets installed later, but i think this is a problem:

```
$ cabal --version
cabal-install version 1.18.0.2
```

jystic:
Gah, can't win:
```
cabal: You need to re-run the 'configure' command. The version of Cabal being used has changed (was Cabal-1.18.1.3, now Cabal-1.22.4.0). Additionally the compiler is different (was ghc-7.8, now ghc-7.10) which is probably the cause of the problem.
```

jystic:
Yay travis:
```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).
The build has been terminated.
```

bin/ci.branches
```diff
@@ -9,5 +9,4 @@ git submodule update
 
 ./cabal update
 ./cabal build
-./cabal test --log=/dev/stdout
-
+./cabal testci --log=/dev/null
```
jystic: Not sure what was the story behind this, from what I can tell the log is always the stdout from the tests anyway, so this just causes duplicate test output to be written in Jenkins.

amosr:
:+1:

PR #72, opened 2015-08-25
Fix date formating to have leading zeros on month and day

HuwCampbell:
I was actually considering changing the date to LocalTime from Data.Time. But this seems fine for now.


amosr:
beautiful
:cactus:

PR #73, opened 2015-08-25
Add annotation to Icicle.Common.Exp.Exp.Exp

jystic:


jystic:
:sob: sorry for the enormous PR!

src/Icicle/Avalanche/Simp.hs
```diff
@@ -17,22 +17,22 @@ import              Icicle.Avalanche.Program
...
-         >>= nestBlocks
-         >>= thresher
+simpAvalanche :: (Show n, Show p, Ord n, Eq p) => a -> Program a n p -> Fresh n (Program a n p)
+simpAvalanche a_fresh p
+ = do p' <- transformX return (simp a_fresh) p
+      s' <- (forwardStmts a_fresh $ pullLets $ statements p')
+         >>= thresher     a_fresh
+         >>= forwardStmts a_fresh
+         >>= nestBlocks   a_fresh
+         >>= thresher     a_fresh
```
jystic: I considered that for situations like this, perhaps we should stash an annotation in the state of `Fresh n` to avoid plumbing it around.

amosr:
this is good
:ambulance:

PR #75, opened 2015-08-26
Topic/cases

amosr:
why can't automatically merge


amosr:
Simple merges are supported, but only very simple ones. I need to add another pass that converts cases with nested constructors into multiple nested cases on single constructors.

amosr:
oh, I forgot to get someone to review it before merging. whoops.

PR #76, opened 2015-08-27
Thought on tombstones

HuwCampbell:
Just writing up some of our thoughts from yesterday, and an idea on implementation (which is one extra construct).

### DIFF
--- a/doc/design/tombstones.md
+++ b/doc/design/tombstones.md
@@ -1,0 +1,73 @@
+On Tombstones
+=============
+Machine learning processes need to be able to support sparse data, indicating a lack of knowledge or genuinely nonsense and unset states.
+It is also required that one can reset this state, so that one's current (or latest) state is the same as if it were never set.
+
+As an example, I may know the country in which Ames lives, and use it as a feature. Let's say I `ingest` this into icicle when they
+moved to their new home.
+```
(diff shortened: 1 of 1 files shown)


amosr:
this looks pretty excellent :moneybag:

jystic:
:money_with_wings:

doc/design/tombstones.md
```diff
@@ -0,0 +1,73 @@
...
+--------
+Any expression which `uses` a tombstone during its evaluation returns a missing value, however, a primitive will be provided to either
+filter or map tombstones, such that they can be handled explicitly.
+
+```
+feature x ~> filter (not tombstone) ~> count
+```
+will filter out all tombstone values and allow count to run.
+```
+feature x ~> let xx = if (tombstone) 0 else value ~> mean xx
```
markhibberd: Naive question here, do you want to be able to fold over these, or at least use context of previous value. For example, perhaps you want to be able keep the value the same as the last "known" value or similar.

markhibberd:
:+1:

doc/design/tombstones.md
amosr: yep, you'll be able to use this tombstone primitive inside a fold, so it should be fine

PR #80, opened 2015-08-27
Java code generator uses the type annotations on the AST

jystic:
This is part of the refactoring to be done to make the C codegen easier to write.

### DIFF
--- a/src/Icicle/Common/Type.hs
+++ b/src/Icicle/Common/Type.hs
@@ -252,17 +252,7 @@ valueMatchesType v t
 -- Pretty printing ---------------
 
 instance Pretty ValType where
- pretty IntT            = text "Int"
- pretty DoubleT         = text "Double"
- pretty UnitT           = text "Unit"
- pretty BoolT           = text "Bool"
- pretty DateTimeT       = text "DateTime"
- pretty StringT         = text "String"
- pretty (ArrayT t)      = parens (text "Array " <> pretty t)
- pretty (MapT k v)      = parens (text "Map" <+> pretty k <+> pretty v)
- pretty (OptionT a)     = parens (text "Option" <+> pretty a)
- pretty (PairT a b)     = text "(" <> pretty a <> text ", " <> pretty b <> text ")"
- pretty (StructT fs)    = parens (pretty fs)
+ pretty = ppTop
 
 instance Pretty StructType where
  pretty (StructType fs) = text "Struct" <+> pretty (Map.toList fs)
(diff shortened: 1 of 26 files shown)


test/cli/repl/run
```diff
@@ -7,7 +7,7 @@ do
     echo " --- Test: $dir"
     if [ $update_expected -eq 1 ]
     then
-        dist/build/icicle-repl/icicle-repl ":import data/libs/prelude.icicle" ":load test/cli/repl/data.psv" < $dir/script | diff $dir/expected -
+        dist/build/icicle-repl/icicle-repl ":import data/libs/prelude.icicle" ":load test/cli/repl/data.psv" < $dir/script | diff -u $dir/expected -
```
amosr: does this give nicer output or something?

src/Icicle/Common/Type.hs
```diff
@@ -272,6 +263,29 @@ instance Pretty FunType where
...
+    BoolT      -> text "Bool"
+    DateTimeT  -> text "DateTime"
+    StringT    -> text "String"
+    ArrayT t   -> parens' (text "Array " <>  ppSub t)
+    MapT k v   -> parens' (text "Map"    <+> ppSub k <+> ppSub v)
+    OptionT a  -> parens' (text "Option" <+> ppSub a)
+    PairT a b  -> parens  (ppTop a <> text ", " <> ppTop b)
+    StructT fs -> parens' (pretty fs)
+  where
+    parens' | needParens = parens
```
amosr: hooray!

amosr:
yep, this is definitely worth it. thanks for convincing me to do it properly :seedling:

test/cli/repl/run
jystic: Yeah it makes it the same as a `git diff` so you get a bit more context:

**file.a**
```
let x = read# [Int] ptr
let y = read# [Int] ptr

foreach (e : es) {
  printf# "f(x, y, e) = %d" (f x y e)
}

let z = read# [Int] ptr
```

**file.b**
```
let x = read# [Int] ptr
let y = read# [Int] ptr

foreach (e : es) {
  printf# "g(x, y, e) = %d" (g x y e)
}

let z = read# [Double] ptr
```

**diff file.a file.b**
```diff
5c5
<   printf# "f(x, y, e) = %d" (f x y e)
---
>   printf# "g(x, y, e) = %d" (g x y e)
8c8
< let z = read# [Int] ptr
---
> let z = read# [Double] ptr
```

**diff -u file.a file.b**
```diff
--- file.a	2015-08-28 05:11:08.000000000 +1000
+++ file.b	2015-08-28 05:11:12.000000000 +1000
@@ -2,7 +2,7 @@
 let y = read# [Int] ptr

 foreach (e : es) {
-  printf# "f(x, y, e) = %d" (f x y e)
+  printf# "g(x, y, e) = %d" (g x y e)
 }

-let z = read# [Int] ptr
+let z = read# [Double] ptr
```

PR #83, opened 2015-08-31
Add an avalance runner to the simulator

HuwCampbell:
Adds a version of evaluateVirtualValue which takes an avalanche program instead of a core one.

### DIFF
--- a/src/Icicle/Avalanche/Eval.hs
+++ b/src/Icicle/Avalanche/Eval.hs
@@ -1,8 +1,10 @@
 -- | Evaluate Avalanche programs
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE OverloadedStrings #-}
 module Icicle.Avalanche.Eval (
     evalProgram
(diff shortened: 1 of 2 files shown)


src/Icicle/Avalanche/Eval.hs
```diff
@@ -55,6 +58,26 @@ data RuntimeError a n p
...
+ pretty (RuntimeErrorAccumulator p)
+  = pretty p
+ pretty (RuntimeErrorLoop p)
+  = pretty p
+ pretty (RuntimeErrorLoopAccumulatorBad n)
+  = "Bad look accumulator:" <> pretty n
+ pretty (RuntimeErrorIfNotBool p)
+  = "Value should be a bool but isn't" <> (pretty p)
+ pretty (RuntimeErrorForeachNotInt p p')
+  = "Foreach not ints:" <> pretty p <> pretty p'
```
amosr: Probably want "<+>" here instead of <>

amosr:
excellent. will be very curious to see the performance differences between Core and Avalanche (though I don't expect either to be good)
:tada:

Issue #79, opened 2015-08-27
Folds over groups


amosr:
Groups must end in folds, but there is currently no way to fold over all the values in the group.

Suppose we want to group by days, then find the day with the maximum number of entries:
```
feature feat
~> group fold (key,value) = (group date ~> count)
~> maximum value
```

We could probably simplify the syntax a bit; there's no point of a "group fold" over something that isn't a group.
Keeping them separate might be useful for function reuse though.

The type is something like
```
group fold (N_key, N_value) = ( |- Q : Aggregate (Group key value))
~> (N_key : Element key, N_value : Element value |- Aggregate a)
 : Aggregate a
```
that is, the bindings (key,value) become elements bound in the rest.

When converting this to Core, the group itself becomes a fold that constructs a Map.
The group fold should just to a Map_fold at the end as a postcomputation, with N_key and N_value.

See also the conversion to Core for Distinct and Latest.

@tranma

tranma:
I was wondering whether key is needed, seeing as if you want to do anything with the key, you'd better use `distinct`

amosr:
hm. I don't think it hurts to include both; it should require the same effort.
I can't think of any interesting uses of folding over the keys right now, but I don't see any reason to rule it out.
distinct is really just a degenerate case of group fold, I guess

PR #87, opened 2015-09-09
Add group-fold to Source

tranma:
This adds the ability to fold over values in group-by's. See https://github.com/ambiata/icicle/issues/79

Plus two new prelude functions, min and max.


amosr:
Arbitrary instance would be super

amosr:
this is perfect :blush:

tranma:
mergification?

jystic:
:8ball:

PR #88, opened 2015-09-09
Minor REPL enhancements

tranma:
* .icicle file!!!
* handle ctrl-c (Max would be proud)

### DIFF
--- a/main/icicle-repl.hs
+++ b/main/icicle-repl.hs
@@ -1,9 +1,10 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE PatternGuards     #-}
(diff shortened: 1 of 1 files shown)


jystic:
:100:

wondering if it would make sense to load an `.icicle` file from the current directory as well?

tranma:
oh yeah, also should make it optional so if the user doesn't have a .icicle file, it's still good

main/icicle-repl.hs
```diff
@@ -61,18 +61,22 @@ main
...
-             , autoAddHistory = True}
+    settings home
+      = return $ HL.defaultSettings
+          { historyFile    = Just $ home <> "/.icicle-repl.history"
+          , autoAddHistory = True}
+    dotfile home
+      = lines <$> readFile (home <> "/.icicle")
     loop state
-      = do line <- HL.getInputLine "> "
+      = handleInterrupt (loop state)
```
jystic: already better than `sbt` :laughing:

tranma: it still doesn't handle ctrl-c??

jystic:
:honey_pot:

PR #90, opened 2015-09-09
Add some hand written translations of Avalanche -> C

jystic:


doc/c-backend/examples/latest.c
```diff
@@ -0,0 +1,82 @@
...
+    int64_t     output_repl;
+
+    const char *error;
+};
+
+void compute(struct icicle_state *s)
+{
+    // init [Mutable] [Option Int] acc$conv$6 = None : Option Int;
+    bool    acc_conv_6_set = false;
+    int64_t acc_conv_6_val = 0;
```
tranma: ah, I see what you meant when you said options needed to be converted to (bool, val) in Avalanche now. Doing it in Core would be easier and still makes sense, I think....

jystic: I think I almost have it worked out in Avalanche, today is going much better than yesterday :racehorse:

Would be interested to know what @amosr thinks about doing all this in core instead.

amosr: I am not entirely sure about doing it in Core, because the translation from Core to Avalanche can introduce new variables. right now we don't introduce any optional values, but I'm not certain that we want to rule it out.
I see that it would be easier in Core, but I don't think it's significantly so. however, if you run into unforeseen issues, it's a good backup plan.

jystic: @amosr cool, I have it worked out in Avalanche I think so it's all good

tranma: I really like the name of this transform pass

PR #91, opened 2015-09-15
Update libraries

jystic:
Need to do this so I can add `jetski` as a dependency


tranma:
:full_moon_with_face:

PR #94, opened 2015-09-15
Add jetski dependency

jystic:
Just wanted to get all the cabal stuff out the way so it's not tangled up in any rebasing


tranma:
:+1:

Issue #93, opened 2015-09-15
README does not contain explanation of the pun behind 'icicle'


glasnt:
I mean, it's a cool name, but I'd love to know why.

blever:
Icicle == I(vory) SQL.

jystic:
:cake:

phickey:
It's also related to https://github.com/ambiata/ice (cut down ivory afaik), but I don't know if that pun is explained either....

PR #97, opened 2015-09-15
Add explaination to why icicle

glasnt:
Resolves issue #93


jystic:
:dromedary_camel: :mega:

PR #98, opened 2015-09-15
WIP: Status update

jystic:
This is not really a PR, just giving some visibility on what is going on

When I created Trample I intended it to be a merge of Melt/Constructors, but I think I'll keep them separate and put the code from Trample in to Melt when I'm done.

### DIFF
--- a/src/Icicle/Avalanche/Segfault.hs
+++ b/src/Icicle/Avalanche/Segfault.hs
@@ -1,31 +1,37 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# OPTIONS_GHC -w #-}
 module Icicle.Avalanche.Segfault (
-    segfault
+    SegfaultError (..)
+  , segfault
   ) where
 
 import           Control.Monad.IO.Class (liftIO)
-import           Control.Monad.Trans.Either (EitherT)
+import           Control.Monad.Trans.Either (EitherT(..), hoistEither)
 
 import           Data.Int (Int64)
 import           Data.Text (Text)
 import qualified Data.Text as T
 import           Data.Word (Word64)
+import qualified Data.Map as Map
 
 import           Foreign.C (CChar)
-import           Foreign.Marshal (allocaBytes)
+import           Foreign.Marshal (allocaBytes, withArray, withArrayLen)
 import           Foreign.Ptr (Ptr, nullPtr)
 import           Foreign.Storable (Storable(..))
 
 import           Icicle.Avalanche.Prim.Flat (Prim)
 import           Icicle.Avalanche.Program (Program)
-import           Icicle.Avalanche.ToSea (seaOfProgram)
+import           Icicle.Avalanche.ToSea (seaOfProgram, outputsOfProgram)
 import           Icicle.Common.Annot (Annot)
 import           Icicle.Common.Base
+import           Icicle.Common.Type (ValType(..))
 import           Icicle.Data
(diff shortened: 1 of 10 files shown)


src/Icicle/Avalanche/Statement/Simp/Trample.hs
```diff
@@ -0,0 +1,185 @@
...
+
+------------------------------------------------------------------------
+
+pattern PrimZip  ta tb = PrimArray   (PrimArrayZip ta tb)
+pattern PrimPair ta tb = PrimMinimal (Min.PrimConst (Min.PrimConstPair ta tb))
+pattern PrimFst  ta tb = PrimMinimal (Min.PrimPair  (Min.PrimPairFst   ta tb))
+pattern PrimSnd  ta tb = PrimMinimal (Min.PrimPair  (Min.PrimPairSnd   ta tb))
+pattern PrimMkOpt   tv = PrimOption  (PrimOptionPack          tv)
+pattern PrimIsSome  tv = PrimProject (PrimProjectOptionIsSome tv)
+pattern PrimGet     tv = PrimUnsafe  (PrimUnsafeOptionGet     tv)
```
tranma: nice! oh wait, pattern synonyms work on 7.8.x?

jystic: Yes! :watermelon:

src/Icicle/Avalanche/Segfault.hs
```diff
@@ -0,0 +1,125 @@
...
+        pokeWordOff pState 2 new_date
+        pokeWordOff pState 3 new_fact
+        pokeWordOff pState 4 out_error
+        pokeWordOff pState 5 out_repl
+
+        _ <- compute [argPtr pState]
+
+        whee <- peekWordOff pState 5
+
+        return [(OutputName "repl", DoubleValue whee)]
```
jystic: This isn't doing anything other than calling the JITed C with some random facts (just checks that the generated code doesn't segfault basically)

src/Icicle/Avalanche/Statement/Simp/Trample.hs
```diff
@@ -0,0 +1,185 @@
...
+
+
+------------------------------------------------------------------------
+
+trample :: (Show n, Ord n)
+        => a
+        -> Statement a n Prim
+        -> Fresh n (Statement a n Prim)
```
tranma: so tldr: trample breaks up pairs and options?

jystic: yep, so melt currently does this for pairs already, trample is the same but works for options too

src/Icicle/Avalanche/Segfault.hs
tranma: hm, so this is like a test?

jystic: at the moment it is, but it will be a full evaluator for the repl in a day or so hopefully

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -78,13 +98,47 @@ constructor a_fresh statements
...
+   = Nothing
+
+
   get env n
    | (_,x'):_ <- filter ((==n).fst) env
    = Just x'
    | otherwise
    = Nothing
+
+  -- ANF? :(
```
tranma: I thought arguments to prim apps would definitely be xvars.... perhaps I misunderstood?

tranma:
this is cool. trample :boot: to :ocean:

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
jystic: Hmm maybe my bad in Trample actually, or I need an ANF pass between Trample and Constructor

Issue #89, opened 2015-09-09
Case constraints are too strict


tranma:
When generating constraints for case expressions, Source requires "the scrutinee and the alternatives must have the same temporality". Could this restriction be relaxed? There's a comment that this may be restrictive than necessary....

Example: `let fold1 a = ((0,0),0) : case a | ((n,m),m2) -> (let d = (double value) - m ~> ((n + 1, m + d / n), m2 + d * ((double value) - m))) end ~> ...`

amosr:
ahh. good catch. yep!
so there is a constraint called ``ReturnOfLet`` or something, which computes the return temporality of a let.
maybe we could require all the alternatives to be the same temporality, and then the return temporality will just be ``ReturnOfLet scrut alts``?

amosr:
Tran fixed this

Issue #78, opened 2015-08-27
Flattening/desugaring for source case expressions


amosr:
Case expressions with nested constructors need to be desugared before conversion to Core.
Or, the Core conversion needs to be improved to deal with nested constructors, but I think desugaring is the simpler option.

For example, this case expression with nested constructors cannot be converted as-is:
```
     case x
      | Some (Some v, None) -> v
      | Some (None, Some v) -> v
      | _                   -> 0
     end
```

We need to pull out the nested constructors to separate cases.
Because we don't have local function definitions, it is easier to just duplicate the alternative code for each new alternative.
We might end up with the same expression duplicated a few times, but it's not a big problem for such small queries (and it's possible we might get some specialisation gains from it)

So here, the x first needs to be changed to a case on the outer part of x: just whether it's Some or None.
In the None case, the default -> 0 applies. In the Some case, we need to check if the inner bit is (Some v, None) or (None, Some v). If it's neither, the default of the outermost case applies (0).

```
     case x
     | Some y
     -> case y
        | (l,r)
        -> case l
           | Some v
           -> case r
              | Some _
              -> 0
              | None
              -> v
              end
           | None
           -> case r
              | Some v
              -> v
              | None
              -> 0
              end
           end
        end
     | None
     -> 0
```

To implement this, you'll probably need a fresh name supply, so the type would be something like
```
desugarQ :: Query a n -> Fresh n (Query a n)
```
and it's probably worth adding as a module Icicle.Source.Transform.Desugar or something.

but it might end up that you need the type of the scrutinee as well (but unlikely the type of the alternatives), so you might need:
```
desugarQ :: Query (Annot a n) n -> Fresh n (Query (Annot a n) n)
```

The Core conversion does not handle nested constructors, nor default patterns or variable patterns at the top level.
We need to remove default and variable patterns and duplicate the alternative for each actual constructors - or in the special (useless) case of just one alternative: 
```
case x 
 | _ -> do stuff with x
end
```
we can probably remove the case altogether:
```
do stuff with x
```

After this is implemented, should be able to add more interesting patterns to the Arbitrary in test/Icicle/Test/Source/Arbitrary.hs, line 47.

It will also be useful to look at src/Icicle/Source/ToCore/Exp.hs (convertCase)

amosr:
Actually, if we can do this *before* type checking, it will make exhaustive checks trivial

tranma:
[...]

PR #99, opened 2015-09-16
more on case expressions

tranma:
In-progress work for:
  * binding names introduced by cases
  * relaxing case constraints (https://github.com/ambiata/icicle/issues/89)
  * flattening cases (https://github.com/ambiata/icicle/issues/78)

maybe works...? I need to massage the case constraints a bit still.

/cc @amosr

### DIFF
--- a/src/Icicle/Source/Checker/Constraint.hs
+++ b/src/Icicle/Source/Checker/Constraint.hs
@@ -395,7 +395,7 @@ generateX x
 
            let tmpx = getTemporalityOrPure $ annResult $ annotOfExp scrut'
            let tmpq = getTemporalityOrPure $ resT
-           require a $ CEquals tmpx tmpq
+           -- require a $ CEquals tmpx tmpq
            let t' = resT -- canonT $ Temporality retTmp resT
 
            annotate (Map.unions (sub : subs)) t' $ \a' -> Case a' scrut' pats'
--- a/src/Icicle/Source/Checker/Constraint.hs
+++ b/src/Icicle/Source/Checker/Constraint.hs
@@ -65,6 +65,8 @@ defaults q
    = []
   defaultOfConstraint (CReturnOfLatest _ _ _)
    = []
+  defaultOfConstraint (CTemporalityJoin _ _ _)
+   = []
 
 
 
@@ -343,69 +345,37 @@ generateX x
              $ hoistEither $ errorNoSuggestions (ErrorFunctionWrongArgs a x fErr [])
            annotate Map.empty resT $ \a' -> Prim a' p
 
-    -- Cases require all their alternatives to have the same type, as well as
-    -- the scrutinee and the alternatives to have the same temporality.
-    -- This may be more restrictive than necessary, since lets allow different
-    -- temporalities, but hopefully will not be an issue.
-    --
-    -- TODO: cases with nested constructors must be converted to multiple nested cases.
-    -- For example:
-    -- > case x
-    -- >  | Some (Some v, None) -> v
-    -- >  | Some (None, Some v) -> v
-    -- >  | _                   -> 0
-    -- > end
-    --
-    -- this should be flattened to
-    -- > case x
-    -- > | Some y
-    -- > -> case y
-    -- >    | (l,r)
-    -- >    -> case l
-    -- >       | Some v
-    -- >       -> case r
-    -- >          | Some _
-    -- >          -> 0
-    -- >          | None
-    -- >          -> v
-    -- >          end
-    -- >       | None
-    -- >       -> case r
-    -- >          | Some v
-    -- >          -> v
-    -- >          | None
-    -- >          -> 0
-    -- >          end
-    -- >       end
-    -- >    end
-    -- > | None
-    -- > -> 0
-    --
+    -- Cases require:
+    --  * Alternatives to have "join-able" types, i.e. no mixing of aggregates and elements.
+    --  * The scrutinee and alternatives to have "join-able" temporalities.
     Case a scrut pats
      -> do (scrut', sub) <- generateX scrut
            (pats', subs) <- unzip <$> mapM (generateP $ annResult $ annotOfExp scrut') pats
 
-           resT <- case pats' of
-                    [] -> hoistEither
-                        $ errorNoSuggestions
-                        $ ErrorEmptyCase a x
-                    ((_,alt):_)
-                       -> return $ annResult $ annotOfExp alt
+           let scrutT  = annResult $ annotOfExp scrut'
+           resT <- TypeVar <$> fresh
 
-           -- all alternatives must have the same temporality
-           -- this is too strict too: let fold s = 0 : case s < value | True -> s | False -> value end
-           mapM_ (\(_,alt) -> require a $ CEquals resT $ annResult $ annotOfExp alt) pats'
+           traceM ("SCRUTTY=" <> show (scrutT))
+           traceM ("REST=" <> show resT)
+           traceM ("PATS=" <> show (fmap (annResult . annotOfExp . snd) pats'))
+
+           -- all alternatives must have compatible temporalities
+           joinAlts a scrutT pats'
 
            -- return temporality
-           let tmpx = getTemporalityOrPure $ annResult $ annotOfExp scrut'
-           let tmpq = getTemporalityOrPure $ resT
-           --require a $ CEquals tmpx tmpq
-           let t' = resT -- canonT $ Temporality retTmp resT
-           --let t' = canonT $ Temporality tmpq resT
+           mapM_ (require a . CTemporalityJoin resT scrutT . annResult . annotOfExp . snd) pats'
 
+           let t' = resT
            annotate (Map.unions (sub : subs)) t' $ \a' -> Case a' scrut' pats'
 
  where
+  joinAlts a t pats
+   = mapM_ (require a . uncurry (CTemporalityJoin t))
+   $ cross $ fmap (annResult . annotOfExp . snd) pats
+
+  cross xs
+   = [ (a,b) | a <- xs, b <- xs, a /= b]
+
   annotate s' t' f
    = do cs <- stateConstraints <$> (lift $ lift State.get)
         let a' = Annot (annotOfExp x) t' cs
--- a/src/Icicle/Source/Checker/Constraint.hs
+++ b/src/Icicle/Source/Checker/Constraint.hs
@@ -67,6 +67,8 @@ defaults q
    = []
   defaultOfConstraint (CTemporalityJoin _ _ _)
    = []
+  defaultOfConstraint (CExtractTemporality _ _ _)
+   = []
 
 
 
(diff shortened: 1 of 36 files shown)


jystic:
Nice! this looks pretty difficult :nut_and_bolt: :minibus: :no_pedestrians:

amosr:
so you asked before about patdefault but I can't remember where.
I think if a case has only a default alternative you can just desugar it to a let:
```
case x y z
 | _ -> xxx
end

==>

let _ = x y z
~> xxx
```
but if a case had one constructor and a default, I think duplicate the default alternative for any others.. I guess
```
case zz
 | (Some x, Some y) -> x + y
 | _ -> 0
end

==>

case zz
 | (sx,sy)
 -> case sx
       | Some x
       -> case sy
            | Some y -> x + y
            | None     -> 0
            end
        | None
        -> 0
        end
  end
```

this duplication isn't ideal - if we had continuations or whatever I guess we could do without it, but I don't think it will be a real problem in practice.

tranma:
@amosr yeah, that's more or less what I did. Introduced a let-binding for defaults as well as variables variables. So `case s | Some _ -> None` desugars into:

```
case s
  | (Some desugar$0) -> (let desugar$1 = desugar$0 ~> None)
```

There are more let's than necessary but that could be cleaned up by beta reduction, I hope.

tranma:
Another problem is the constraint "alternatives must have the same temporality", I think this is too strict as well. You might want cases like these `case s | Some x -> Some (x+1) | _ -> None end`, where `None` is `Pure` but `s` isn't.

tranma:
I think this is baked enough for review. Some places are pretty hacky since I don't really know what I'm doing...

/cc @amosr @jystic

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -343,64 +345,35 @@ generateX x
...
-                        $ ErrorEmptyCase a x
-                    ((_,alt):_)
-                       -> return $ annResult $ annotOfExp alt
+           let scrutT  =  annResult $ annotOfExp scrut'
+           resT        <- case pats' of
+                           [] -> hoistEither
+                               $ errorNoSuggestions
+                               $ ErrorEmptyCase a x
+                           ((_,alt):_)
+                              -> return $ annResult $ annotOfExp alt
```
amosr: hmm - if the first alternative is Pure and the others are Element, the return should be Element.
I think we need a fresh type variable for the return, then go through and say the return = the join of all the scruts and alts together.
actually that's rather annoying, because you need a fresh variable for each scrut and alt..

```
case S
 | 1 -> A1
 | 2 -> A2

==>
Constraints:
 T1 = Join S1 A1
 T2 = Join T1 A2

Return temporality = T2

```

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -343,64 +345,35 @@ generateX x
...
+           let t' = resT
            annotate (Map.unions (sub : subs)) t' $ \a' -> Case a' scrut' pats'
 
  where
+  joinAlts a t pats
+   = mapM_ (require a . uncurry (CTemporalityJoin t))
+   $ cross $ fmap (annResult . annotOfExp . snd) pats
+
+  cross xs
+   = [ (a,b) | a <- xs, b <- xs, a /= b]
```
amosr: I don't think you need to cross product, just pair each with the next one

src/Icicle/Source/Checker/Constraint.hs
tranma: so I tried this (`resT <- TypeVar <$> fresh`) then try to unify `resT` with the join of scrut and the alternatives. But if the type of `scrut` is just type variable, and the alternatives are `Element`, that wouldn't unify... because `scrut` would just be `Pure`

amosr: (where T1 and T2 are fresh variables and S1, A1 and A2 are the temporalities of scrutinee and alternatives)

amosr: yeah actually I think you need a fresh variable for each alternative. and you start by joining the scrut and the first alternative, put that in a fresh var. then join the last fresh and the next alternative, and put it in a new fresh. then at the end the last fresh will be the join of all the scrut and alternatives.

wow! that's weird! maybe we can actually change the Join constraint to be (Join (Type n) [Type n]) so the output is a single var but takes multiple inputs

tranma: yeah, I do think you need some kind of fresh for each alternative. Sounds a bit complicated, hm.

tranma:
I "fixed" the constraint checker such that it would ensure the temporalities of the scrutinee + alternatives do join, and the base types (without the temporality) are equal. I think this works....or at least passed 1000 quickcheck tests with random cases.

/cc @amosr

test/cli/repl/t20-lexer/expected
```diff
@@ -30,7 +30,10 @@ Check error:
       mean                 : (Num b) => Element a b -> Aggregate a Double
       date                 : Element Definitely DateTime
       min                  : Element a b -> Aggregate a b
-      sd                   : (Num b) => Element a b -> Aggregate a Double
+      sd                   : (Element =: ReturnOfLet Pure t$107
+                             ,t$105 =: ReturnOfLet a a
+                             ,t$106 =: ReturnOfLet a t$105
+                             ,t$107 =: ReturnOfLet a t$106) => a b Double -> Aggregate b Double
```
amosr: hmm this is a shame, wonder why it's changed to such a strange type

amosr: oh ok that's not related to your typechecker changes, just the different definition of sd. I'll look at that later.

tranma: Huw said it was numerically unstable, so he pointed me to this algorithm https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -343,75 +347,51 @@ generateX x
...
-           let tmpq = getTemporalityOrPure $ resT
-           require a $ CEquals tmpx tmpq
-           let t' = resT -- canonT $ Temporality retTmp resT
 
+           -- Destruct the scrutinee type into the base type
+           -- and the temporality (defaulting to Pure).
+           let scrutT  = annResult $ annotOfExp scrut'
+           scrutTy    <- TypeVar <$> fresh
+           scrutTm    <- TypeVar <$> fresh
+           require a $ CExtractTemporality scrutTm scrutTy scrutT
```
amosr: is this necessary as a separate constraint? I thought you should be able to do "scrutTm = getTemporalityOrPure scrutT"

test/cli/repl/t20-lexer/expected
amosr: yes that's ok, it's just strange that the type is so ugly. it must be a pre-existing typechecker bug

src/Icicle/Source/Checker/Constraint.hs
tranma: That gives you the temporality of the scrutinee (or Pure). But I wanted to do was adding the unification variable `scrutTm` into the context, so that it will resolve to the right temporality when it discharges the joins.

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -419,11 +399,27 @@ generateP scrutTy (pat, alt)
...
+        altTy      <- TypeVar <$> fresh
+        altTp      <- TypeVar <$> fresh
+
+        -- Require alternative types to have the same temporality if they
+        -- do have temporalities. Otherwise defaults to TemporalityPure.
+        require (annotOfExp alt) $ CExtractTemporality altTp altTy altTy'
 
-        e' <- lift $ lift State.get
+        -- Require the alternative types without temporality to be the same.
+        require (annotOfExp alt) $ CEquals resTy altTy
```
amosr: ah, I think "requireData" would suffice here, which checks just the data type as opposed to the temporality and possibility too:
https://github.com/ambiata/icicle/blob/case/src/Icicle/Source/Checker/Constraint.hs#L287

test/cli/repl/t20-lexer/expected
tranma: right, it kinda looks like you need to do more unification for ReturnOfLet?

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -419,11 +399,27 @@ generateP scrutTy (pat, alt)
...
+
+        -- Require return temporality to be compatible with alternative temporalities.
+        resTp' <- TypeVar <$> fresh
+        require (annotOfExp alt) $ CTemporalityJoin resTm resTp' altTp
+
+        e'    <- lift $ lift State.get
         lift $ lift $ State.put (e' { stateEnvironment = stateEnvironment e })
+        rest' <- generateP ann scrutTy resTy resTp' rest
+
+        return (((pat,alt'), sub):rest')
```
amosr: yep these typechecker changes look correct to me. I'm just not convinced that the CExtractTemporality constraint is necessary.

src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -637,19 +638,21 @@ convertReduce xx
...
  | Case (Annot { annAnnot = ann, annResult = retty }) scrut patalts <- xx
- = do   scrut' <- convertReduce scrut
-        let (pats,alts) = unzip patalts
-        alts'  <- mapM convertReduce alts
+ = do   let (pats,alts) = unzip patalts
+
+        scrut' <- convertReduce scrut
+        pats'  <- convertCaseFreshenPats pats
+        alts'  <- mapM convertExp alts
+        alts'' <- hoist runFreshIdentity $ lift $ mapM (ANormal.anormal ()) alts'
```
amosr: is it necessary or beneficial to a-normalise here - why?

test/cli/repl/t20-lexer/expected
amosr: yeah, maybe we can fix it with a new rule "t =: ReturnOfLet a a ==> t =: a"

src/Icicle/Source/ToCore/ToCore.hs
tranma: I wanted to make sure nested case statements get generated to `conv = let .. in .. let .. in ....`. Perhaps that is the case even without a-normalisation... This line might be premature, yeah.

amosr: my mantra: do the smallest, simplest thing that's correct. if it's generating excessive lets we can a-normalise it in a separate pass afterwards (we probably already do)

src/Icicle/Source/Checker/Constraint.hs
tranma: hm. I would like to unify types `t` and `u`, where `t` has some temporality `tempT` and base type `tyT`.
If  `u` is `Temporality _ _`, then the specific temporalities must join. Otherwise, if `u` does not have temporality (e.g. just a BoolT), then default it to Pure and perform the join. If `u` is a type variable, defer the constraint.
I tried to do this with CEquals and getTemporality but that fixes the constraints too early...

amosr: hmm right, okay. I thought it would work because I thought everything that generates a temporality returned them already split up, eg as ``Temporality t u``.
ok, this is good then - maybe I'll have to fix the rest of the typechecker to perform this extraction too

tranma: also, I forgot to ask earlier, with `decomposeT`, it doesn't perform a fixpoint on the type, right? It just assumes that the type is `Temporality (Possibility _ <base type>)`, never any other ordering?

amosr: ah, so decomposeT actually does some magic and rips out nested temporalities etc, which is just to make it easier with dealing with them.
eg if you had (Temporality Element Int) and wanted to wrap it in an Array, you can just go
``canonT $ Array $ Temporality Element Int``
and get back ``Temporality Element (Array Int)``.
(canonT is just recomposeT.decomposeT)

so, yes, it recurses over the type

src/Icicle/Source/ToCore/ToCore.hs
tranma: yeah, you're right. I don't know what I was thinking... a-normal shouldn't change whether the generated Core typechecks.

tranma:
I'm going to move the "simp" stuff out of desugar

amosr:
that'd be great. Maybe move it to ``Icicle.Source.Transform.Simp``?
if you want - probably after merging the case stuff back in - I think it'd be worth doing more simp too, eg dead code removal, case of constructor, case of case?
I feel like this stuff is worth doing at every stage because they're all different levels of abstraction, so simple dead code removal in Source might become a bit more involved in Core for example.
sound reasonable?
:+1:

tranma:
yeah, I think these dead case/let simplifications in particular should be done in Source rather than Core. 

For now, I'd just move them to something like `Icicle.Source.Transform.Simp.Dumb`, as they remove "dumb" lets (`let x = y in let z = x in...`) and "dumb" cases (`case x | y -> ...`) rather than "dead code". The bindings and cases are not unused, just too roundabout.

As for actual DCE and stuff, it might be easier in Source or Core, I dunno...

amosr:
good plan

tranma:
@amosr you can merge this to look at it in more detail if you want. I'm doing some stress testing as well, I could make another PR later. I think I need a test: doesn't typecheck before desugar/simp ==> doesn't typecheck after desugar/simp

amosr:
that sounds like an interesting property. I'm not completely sure about it - I think in general, simplification  could actually make some things well typed? like eliminating dead code in "let x = error in dont-use-x"

tranma:
@amosr oh right, of course. As soon as I write the property it gave me that as a counterexample >_<

amosr:
teehee!

tranma:
this feature by itself is ok now, I think... (passes 100000 quickcheck tests!)

I stumbled across some existing problems in the type checker: `group fold (beaker, waldorf) = (group animal ~> 2 ^ 2) ~> let fold1 statler = animal : 1 ~> - 1` typechecks where it shouldn't, leading to not well-typed Core. But that's a separate issue

PR #101, opened 2015-09-18
Newest shouldn't have all facts stuck to it

HuwCampbell:
Maybe this feature should be renamed to current_state or something as well?


jystic:
:shoe:

tranma:
sounds good :+1:
confused me a bit at first glance though, `latest 1 ~> let fold... ` =)

PR #102, opened 2015-09-18
Add tombstone parsing, and force facts to be in a namespace (even tho…

HuwCampbell:
…ugh it's not used yet).

Evaluators currently don't know how to deal with tombstones either.

### DIFF
--- a/src/Icicle/Storage/Dictionary/TextV1.hs
+++ b/src/Icicle/Storage/Dictionary/TextV1.hs
@@ -14,8 +14,8 @@ import           P hiding (concat, intercalate)
 import           Data.Attoparsec.Text
 
 import           Data.Either.Combinators
-import           Data.Text hiding (takeWhile, singleton)
-import           Data.Set (singleton)
+import           Data.Text hiding (takeWhile)
+import qualified Data.Set as Set
 
 import           Icicle.Storage.Encoding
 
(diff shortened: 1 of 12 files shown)


tranma:
should evaluators handled tombstones as described in https://github.com/ambiata/icicle/blob/master/doc/design/tombstones.md?

jystic:
@tranma yep

src/Icicle/Storage/Dictionary/TextV1.hs
```diff
@@ -14,7 +14,8 @@ import           P hiding (concat, intercalate)
 import           Data.Attoparsec.Text
 
 import           Data.Either.Combinators
-import           Data.Text hiding (takeWhile)
+import           Data.Text hiding (takeWhile, singleton)
+import           Data.Set (singleton)
```
jystic: Should really use qualified imports here

test/Icicle/Test/Encoding.hs
```diff
@@ -7,11 +7,12 @@ module Icicle.Test.Encoding where
 import           Icicle.Test.Arbitrary
 import           Icicle.Encoding
 
-import           P
+import           P hiding (empty)
 
 import           System.IO
 
 import           Test.QuickCheck
+import           Data.Set (empty)
```
jystic: Qualified imports

jystic:
:tangerine: looks good to me

might be good to get @amosr to have a look though (i think he might be in today), I'm not sure what the implications of adding `VTombstone` to `BaseValue` are

amosr:
ok! :banana:

PR #103, opened 2015-09-18
Update melt/constructor to support splitting options in some cases

jystic:
Adds a new transient primop `Option_pack#` which is inserted by melt, and removed by constructor.

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
+++ b/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
@@ -5,12 +5,14 @@ module Icicle.Avalanche.Statement.Simp.Constructor (
   ) where
 
 import              Icicle.Avalanche.Prim.Flat
-import qualified    Icicle.Common.Exp.Prim.Minimal as Min
-import              Icicle.Avalanche.Statement.Statement
 import              Icicle.Avalanche.Statement.Simp.ExpEnv
+import              Icicle.Avalanche.Statement.Statement
+import qualified    Icicle.Common.Exp.Prim.Minimal as Min
 
+import              Icicle.Common.Base
 import              Icicle.Common.Exp
 import              Icicle.Common.Fresh
+import              Icicle.Common.Type
 
 import              P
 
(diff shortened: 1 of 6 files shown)


src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -78,13 +94,35 @@ constructor a_fresh statements
...
+
   get env n
    | (_,x'):_ <- filter ((==n).fst) env
    = Just x'
    | otherwise
    = Nothing
+
+  -- ANF?
+  get' env (XVar _ n) = get env n
+  get' _    xx        = Just xx
```
tranma: so initially I thought Avalanche flattening is A-normalisation, so this should be ok. But now I'm not sure. If that's not the case, you might need to do A normalisation yourself?

jystic: so [melt](https://github.com/ambiata/icicle/blob/master/src/Icicle/Avalanche/Simp.hs#L35) is called before constructors, perhaps i need to make sure `melt` maintains ANF?

tranma: was Avalanche even in ANF before `melt`? Sorry I don't know Avalanche very well.

jystic: Yeah I believe so, because [simp](https://github.com/ambiata/icicle/blob/master/src/Icicle/Avalanche/Simp.hs#L33) is run over the expressions

I'll try doing it again after melt and see if I can remove `get'`

jystic: Hmm so I seem to get left with prims applied to XValue's in some cases, even after A-normalisation:

```
  init [Mutable] [Bool] acc$conv$6$simp$2 = Option_isSome# [Int] (None : Option Int);
  init [Mutable] [Int] acc$conv$6$simp$3 = unsafe_Option_get# [Int] (None : Option Int);
  init [Mutable] [Bool] acc$conv$12$simp$4 = Option_isSome# [Int] (None : Option Int);
  init [Mutable] [Int] acc$conv$12$simp$5 = unsafe_Option_get# [Int] (None : Option Int);
```

tranma: ah, I missed that. I should have just grep for anormal. :+1:

tranma: This `anormal` seems to allow constant values, that's fine. I don't think it matters for `get'`.

jystic: :(

```haskell
-- | Check whether expression is worth extracting
isNormal :: Exp a n p -> Bool
isNormal xx
 = case xx of
    XVar{}   -> True
    XPrim{}  -> True
    XValue{} -> True
    XLam{}   -> True
    XApp{}   -> False
    XLet{}   -> False
```

tranma: Right...I think `get'` should still be fine, since XPrims applied to Vars would still get simp'ed away (from `XApp`)?

jystic: So I think I need to keep `get'` so that it will either take the value or do a lookup

tranma:
:100:

PR #104, opened 2015-09-20
Preliminary C Evaluation

jystic:
No support yet for:

- "latest" accumulators
- maps
- structs
- tombstones
- probably other stuff

### DIFF
--- a/main/icicle-repl.hs
+++ b/main/icicle-repl.hs
@@ -44,6 +44,8 @@ import           Icicle.Data.DateTime
 import           Icicle.Dictionary
 import           Icicle.Internal.Rename
 import qualified Icicle.Repl                          as SR
+import qualified Icicle.Sea.Eval                      as Sea
+import qualified Icicle.Sea.FromAvalanche             as Sea
 import qualified Icicle.Simulator                     as S
 import qualified Icicle.Source.Parser                 as SP
 import qualified Icicle.Source.PrettyAnnot            as SPretty
@@ -101,10 +103,12 @@ data ReplState
    , hasInlined   :: Bool
    , hasCore      :: Bool
    , hasCoreType  :: Bool
+   , hasCoreEval  :: Bool
    , hasAvalanche :: Bool
    , hasFlatten   :: Bool
    , hasJava      :: Bool
-   , hasEval      :: Bool
+   , hasSea       :: Bool
+   , hasSeaEval   :: Bool
    , doCoreSimp   :: Bool }
 
 -- | Settable REPL states
@@ -114,10 +118,12 @@ data Set
    | ShowInlined        Bool
    | ShowCore           Bool
    | ShowCoreType       Bool
-   | ShowEval           Bool
+   | ShowCoreEval       Bool
    | ShowAvalanche      Bool
    | ShowFlatten        Bool
    | ShowJava           Bool
+   | ShowSea            Bool
+   | ShowSeaEval        Bool
    | CurrentDate        DateTime
    | PerformCoreSimp    Bool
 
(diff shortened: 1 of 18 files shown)


main/icicle-repl.hs
```diff
@@ -347,9 +371,30 @@ handleSetCommand state set
...
+          HL.outputStrLn "         _______-----    ___--          \\    /)\\"
+          HL.outputStrLn "   ------_______      ---____            \\__/  /"
+          HL.outputStrLn "                -----__    \\ --    _          /\\"
+          HL.outputStrLn "                       --__--__     \\_____/   \\_/\\"
+          HL.outputStrLn "                               ----|   /          |"
+          HL.outputStrLn "                                   |  |___________|"
+          HL.outputStrLn "                                   |  | ((_(_)| )_)"
+          HL.outputStrLn "                                   |  \\_((_(_)|/(_)"
+          HL.outputStrLn "                                   \\             ("
+          HL.outputStrLn "                                    \\_____________)"
```
charleso: skull

src/Icicle/Sea/Eval.hs
```diff
@@ -0,0 +1,143 @@
...
+
+------------------------------------------------------------------------
+
+seaEval :: (Show a, Show n, Pretty n, Ord n)
+         => Program (Annot a) n Prim
+         -> DateTime
+         -> [AsAt Value]
+         -> EitherT SeaError IO [(OutputName, Value)]
+
+seaEval program date values = do
```
amosr: hmm I'm guessing we'll eventually want to separate this into "make/compile evaluator" and "call evaluator"

tranma: "compiler" and "interpreter"?

amosr: I don't know

src/Icicle/Sea/Eval.hs
```diff
@@ -0,0 +1,143 @@
...
+      withArray    dates $ \pDates ->
+      withArrayLen facts $ \count pFacts -> do
+
+        forM_ [0..19] (\off -> pokeWordOff pState off (0 :: Word64))
+
+        pokeWordOff pState 0 (wordOfDate date)
+        pokeWordOff pState 1 (fromIntegral count :: Int64)
+        pokeWordOff pState 2 pDates
+        pokeWordOff pState 3 pFacts
+        pokeWordOff pState 4 (nullPtr :: Ptr CChar)
```
amosr: this is really beautiful

src/Icicle/Sea/FromAvalanche.hs
```diff
@@ -0,0 +1,470 @@
...
+  , stateOfProgram program
+  , ""
+  , "static const iunit_t iunit  = 0x1c1c13;"
+  , "static const ibool_t ifalse = 0;"
+  , "static const ibool_t itrue  = 1;"
+  , ""
+  , "#define INLINE __attribute__((always_inline))"
+  , ""
+  , "static idouble_t INLINE iint_extend   (iint_t    x)              { return x; }"
+  , "static iint_t    INLINE iint_add      (iint_t    x, iint_t    y) { return x +  y; }"
```
amosr: I feel like this is more effort than just emitting "x + y" for all types

amosr:
fantastic :bird: can you add a cli test that shows the emitted C code?

src/Icicle/Sea/Eval.hs
jystic: Yeah definitely, I think the whole interface to the repl needs a bit of an overhaul to make it friendlier to consume from ice

src/Icicle/Sea/FromAvalanche.hs
jystic: Hmm, we only have two types though, and this makes the pretty printer simpler because everything is prefix.

Did you see the code for `seaOfXPrim` and friends?

amosr: yeah, fair enough

PR #108, opened 2015-09-22
Bring icicle's map output closer to ivory's.

HuwCampbell:
So this is a bit unfortunate, Ivory outputs what appears to be a json struct for map encodings (group bys and the like).
This isn't really that safe though unfortunately, and we lose quite a bit of type information. It does however make things more consistent down the line if both projects do the same thing.

### DIFF
--- a/src/Icicle/Encoding.hs
+++ b/src/Icicle/Encoding.hs
@@ -156,7 +156,7 @@ renderValue tombstone val
    = T.decodeUtf8
    $ BS.toStrict
    $ A.encode
-   $ jsonOfValue (A.String tombstone) val
+   $ jsonOfValue tombstone val
 
 
 -- | Attempt to decode value with given encoding.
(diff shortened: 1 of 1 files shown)


tranma:
so it will always say "null" for tombstones inside a map?

HuwCampbell:
There's a guard above to ensure tombstone's aren't written out as the key.

amosr:
this looks good to me. :peach:
it doesn't seem to me that we lose that much type information, but I might be misunderstanding. could you elaborate?

HuwCampbell:
Just that for the keys, everything goes to String.

amosr:
oh right, ok. yeah maybe just put a caveat in the comments there. I like comments.
I guess we require this to be true:
```
forall (x y : A).
     show x == show y
 ==> x == y
```
which seems reasonable to me

HuwCampbell:
An alternative could be to have a struct with:
{"k":null, "v":null}
And push the other direction for consistency.

amosr:
I don't mind either way. keeping it as a proper key seems useful if one were to use it as json. I think strange things are fine (and necessary) but we just need to clearly document the rationale and the implications

HuwCampbell:
Ok, let's go with this version.

amosr:
:apple:

Issue #81, opened 2015-08-28
Possibly modality being ignored in Source typechecker


amosr:
The Source typechecker (Icicle.Source.Checker.Constraint) should compute Possibilities; Possibly or Definitely.

Possibility is a bit simpler than Temporality, because Possibility is a full lattice (right?): there is always a meet and a join; Possibly U Definitely = Possibly and so on.
(Whereas Temporality is not; Element U Aggregate = {} or something)

Anyway, we need a new constraint for getting the upper bound of two possibilities. Something like
```
PossibilitiesUpperBound Definitely d = d
PossibilitiesUpperBound Possibly _ = Possibly
PossibilitiesUpperBound p q = PossibilitiesUpperBound q p
```
This can be added to Icicle.Source.Type.Constraints.

Now, the return possibility of most things will be the upper bound of all its arguments. Look at filter, for example
```
filter Definitely ~> Definitely   : Definitely
filter Possibly ~> Definitely   : Possibly
filter Definitely ~> Possibly   : Possibly

filter x ~> c : PossibilitiesUpperBound x c
```
This is quite different from temporality, where filter requires its expression to be Element and the return to be Aggregate.

For function application the idea is the same as for temporality.
If the expected (function type) and the actual (argument type) possibility are the same, function application is as normal.
If the expected and actual are different, either:
1. Expect Definitely, actual Possibly: the function must be modified to return a Possibly. Think of this as unboxing the argument, then reboxing the result of function application.
2. Expect Possibly, actual Definitely: application as normal

For Case expressions, any variables you bind in a pattern have the same possibility and temporality as the scrutinee.
For example, if you have
```
case (xs : Element Possibly (List Int))
 | Cons one rest -> …
```
here, ``one`` and ``rest`` are also ``Element Possibly``.
The return possibility is the upper bound of the scrutinee and all the alternatives.

Lets just use the possibility of the "in" body. It is much simpler than temporality for lets.

Primitives can just be Definitely, because any Definitely can be converted to a Possibly. It is basically pure. Note that in Icicle.Source.Checker.Prim, the primitives are just bare, unmoded types. This is fine because "canonT" and "decomposeT" which are used to check the modes in Checker.Constraint will convert them to Definitely.

tranma:
bit late to ask.... but what exactly is `Possibility` modality? I'm running into this in the source type checker and I'm getting unification variables in the result type for this modality thing.

amosr:
[...]

PR #109, opened 2015-09-24
Add Possibilities to typechecker

amosr:
See issue #81.

### DIFF
--- a/src/Icicle/Source/Type/Compounds.hs
+++ b/src/Icicle/Source/Type/Compounds.hs
@@ -6,6 +6,7 @@
 module Icicle.Source.Type.Compounds (
     function0
   , freeT
+  , freeC
   , canonT
   , decomposeT
   , recomposeT
@@ -57,6 +58,20 @@ freeT t
     PossibilityDefinitely   -> Set.empty
 
     TypeVar n               -> Set.singleton n
+
+
+freeC :: Ord n => Constraint n -> Set.Set (Name n)
+freeC c
+ = case c of
+    CEquals p q             -> Set.union (freeT p) (freeT q)
+    CIsNum  p               -> freeT p
+    CReturnOfLetTemporalities
+                    p q r   -> Set.unions
+                             $ fmap freeT
+                             [ p, q, r ]
+    CReturnOfLatest p q r   -> Set.unions
+                             $ fmap freeT
+                             [ p, q, r ]
 
 
 canonT :: Type n -> Type n
--- a/src/Icicle/Source/Type/Compounds.hs
+++ b/src/Icicle/Source/Type/Compounds.hs
@@ -72,6 +72,9 @@ freeC c
     CReturnOfLatest p q r   -> Set.unions
                              $ fmap freeT
                              [ p, q, r ]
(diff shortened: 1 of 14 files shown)


jystic:
:potable_water: looks good, although I have to admit I don't really understand the meat of it because I'm not too familiar with the type checker

amosr:
@tranma I expect we'll have some conflicts around case expressions and their possibilities. do you think I should wait for you to merge before merging this, or just go ahead and we'll deal with it when it happens?

tranma:
@amosr not sure. We could try and see if the case expression branch is good to merge today?

tranma:
woah, documentation :open_mouth:

test/cli/repl/t20-lexer/expected
```diff
@@ -23,17 +23,14 @@ ok, loaded test/cli/repl/data.psv, 8 rows
...
+      max                  : Element a -> Aggregate Possibly a
+      mean                 : (Num a) => Element a -> Aggregate Double
       date                 : Element Definitely DateTime
-      min                  : Element a b -> Aggregate a b
-      sd                   : (Element =: ReturnOfLet Pure t$137
-                             ,t$135 =: ReturnOfLet a a
-                             ,t$136 =: ReturnOfLet a t$135
-                             ,t$137 =: ReturnOfLet a t$136) => a b Double -> Aggregate b Double
+      min                  : Element a -> Aggregate Possibly a
+      sd                   : (Element =: ReturnOfLet Pure a) => a Double -> Aggregate Double
```
tranma: nice

amosr:
> woah, documentation :open_mouth:
does that count as approval-to-merge?

src/Icicle/Source/Type/Compounds.hs
```diff
@@ -60,6 +62,23 @@ freeT t
...
+                    p q r   -> Set.unions
+                             $ fmap freeT
+                             [ p, q, r ]
+    CReturnOfLatest p q r   -> Set.unions
+                             $ fmap freeT
+                             [ p, q, r ]
+    CPossibilityJoin p q r  -> Set.unions
+                             $ fmap freeT
+                             [ p, q, r ]
+
```
tranma: probably get a "pattern not matched" here on the temporality constraints

tranma:
looks alright to me, I mostly read the doc and glazed over the implementation though :eyeglasses:

src/Icicle/Source/Type/Compounds.hs
amosr: doh…

test/cli/repl/t20-lexer/expected
jystic: That's pretty nice

amosr: not as nice as it should be…

tranma: I guess ideally we don't even want to see `ReturnOfLet`, does `Element =: ReturnOfLet Pure a` just implies `a` is `Element`?

amosr: yeah, it does really. I'm tempted to remove the ReturnOfLet stuff altogether..

PR #111, opened 2015-09-25
Don't crash when an import isn't found.

HuwCampbell:
Preserve the source file name when parsing
import for better error messages.

### DIFF
--- a/src/Icicle/Storage/Dictionary/Toml.hs
+++ b/src/Icicle/Storage/Dictionary/Toml.hs
@@ -83,10 +83,15 @@ loadDictionary' parentFuncs parentConf parentConcrete fp
      $ tomlDict parentConf rawToml
 
   parsedImports
-    <- (\fp' ->
-         EitherT
-         $ ((A.left DictionaryErrorParsecFunc) . SP.parseFunctions)
-        <$> T.readFile (rp </> (T.unpack fp'))
+    <- (\fp' -> do
+         let fp'' = T.unpack fp'
+         importsText
+           <- EitherT
+            $ A.left DictionaryErrorIO
+           <$> E.try (T.readFile (rp </> fp''))
+         hoistEither
+            $ A.left DictionaryErrorParsecFunc
+            $ SP.parseFunctions fp'' importsText
        ) `traverse` (imports conf)
 
   importedFunctions
--- a/main/icicle-repl.hs
+++ b/main/icicle-repl.hs
@@ -248,7 +248,7 @@ handleLine state line = case readCommand line of
 
   Just (CommandImportLibrary fp) -> do
     s  <- liftIO $ T.readFile fp
-    case SR.readIcicleLibrary s of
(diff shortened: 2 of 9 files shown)


amosr:
test!

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -83,10 +83,14 @@ loadDictionary' parentFuncs parentConf parentConcrete fp
...
-         $ ((A.left DictionaryErrorParsecFunc) . SP.parseFunctions)
-        <$> T.readFile (rp </> (T.unpack fp'))
+    <- (\fp' -> do
+         let fp'' = T.unpack fp'
+         importsText
+           <- EitherT
+            $ (A.left DictionaryErrorIO)
+           <$> (E.try (T.readFile (rp </> fp'')))
+         hoistEither
+            $ ((A.left DictionaryErrorParsecFunc) . SP.parseFunctions fp'') importsText
```
tranma: I found this a bit voodooey to read...

amosr: yeah this is crazy

HuwCampbell: Geez guys.

HuwCampbell:
Tests are in

tranma:
:8ball:

amosr:
:moneybag:

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -83,10 +83,15 @@ loadDictionary' parentFuncs parentConf parentConcrete fp
...
+    <- (\fp' -> do
+         let fp'' = T.unpack fp'
+         importsText
+           <- EitherT
+            $ A.left DictionaryErrorIO
+           <$> E.try (T.readFile (rp </> fp''))
+         hoistEither
+            $ A.left DictionaryErrorParsecFunc
+            $ SP.parseFunctions fp'' importsText
        ) `traverse` (imports conf)
```
jystic: tbh still not very happy about this

amosr: hmm this is the nicest I can come up with, without too much effort. not sure.

```
parsedImports <- traverse (readImport rp) (imports conf)

where
 readImport rp fp
  = do  let fp' = T.unpack fp
           importsText <- readFile (rp </> fp')
           hoistEither
                  $ A.left DictionaryErrorParsecFunc
                  $ SP.parseFunctions fp' importsText
 readFile file
  = EitherT $ A.left DictionaryErrorIO <$> E.try (T.readFile file)
```

jystic: ^ much happier with this

The lambda expression just grinds my gears because it kind of implies that you needed to close over something in the environment, but you didn't.

amosr:
I guess the risk here is that if we keep going we can get bogged down in trivial syntax issues. oh well lol

PR #112, opened 2015-09-28
Remove Element names in the "rest" of group fold

tranma:
This example: `group fold (beaker, waldorf) = (group animal ~> 2 ^ 2) ~> let fold1 statler = animal : 1 ~> - 1` is now a type error rather than a convert error.

/cc @amosr

### DIFF
--- a/src/Icicle/Source/Checker/Constraint.hs
+++ b/src/Icicle/Source/Checker/Constraint.hs
@@ -171,10 +171,13 @@ generateQ qq@(Query (c:_) _)
     -- >    : Aggregate a
     GroupFold _ k v x
      -> do  (x',sx)    <- generateX x
+
             retKey     <- Temporality TemporalityElement . TypeVar <$> fresh
             retVal     <- Temporality TemporalityElement . TypeVar <$> fresh
+            removeElementBinds
             (q',sq,t') <- withBind k retKey
                         $ withBind v retVal rest
+
             let tgroup  = annResult $ annotOfExp x'
             let t''     = canonT $ Temporality TemporalityAggregate t'
             requireAgg  t'
@@ -288,6 +291,16 @@ generateQ qq@(Query (c:_) _)
    = let (_,_,d1) = decomposeT t1
          (_,_,d2) = decomposeT t2
      in  require a $ CEquals d1 d2
+
+  removeElementBinds
+   = do st       <- lift (lift State.get)
+        let env   = stateEnvironment st
+            elts  = Map.keys $ Map.filter isElementTemporality env
+            env'  = foldr Map.delete env elts
+        lift . lift . State.put $ st { stateEnvironment = env' }
+
+  isElementTemporality ft
+   = getTemporalityOrPure (functionReturn ft) == TemporalityElement
 
 -- | Generate constraints for expression
 generateX :: Ord n => Exp a n -> Gen a n (Exp'C a n, SubstT n)


src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -174,7 +174,8 @@ generateQ qq@(Query (c:_) _)
             retKey     <- Temporality TemporalityElement . TypeVar <$> fresh
             retVal     <- Temporality TemporalityElement . TypeVar <$> fresh
             (q',sq,t') <- withBind k retKey
-                        $ withBind v retVal rest
+                        $ withBind v retVal
+                        $ removeElementBinds >> rest
```
amosr: hmm this looks to me like it would add "k : Element K" and then remove it straight after?

tranma: oh, right. `withBind` binds the name, then run the action. I need to `removeElementBinds` before all this.

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -171,10 +171,13 @@ generateQ qq@(Query (c:_) _)
     -- >    : Aggregate a
     GroupFold _ k v x
      -> do  (x',sx)    <- generateX x
+
             retKey     <- Temporality TemporalityElement . TypeVar <$> fresh
             retVal     <- Temporality TemporalityElement . TypeVar <$> fresh
+            removeElementBinds
             (q',sq,t') <- withBind k retKey
                         $ withBind v retVal rest
+
```
amosr: hmm damn, I think you'll actually need to reset the context after here.
stupid - this is my fault. do you want the honours of killing the state monad?

tranma: yeah, I was a bit unsure about what to do after this. Do I re-add the things that were removed...?

I can try and kill the state monad, just pass both the environment and constraint sets around... Dubious honour :imp:

tranma:
don't think this is correct. You want to remove more than just "Element"s. Consider:

```
feature animal
~> group fold (waldorf
              ,chef) = (group 0
                        ~>
                        let fozzy = 1
                        ~> 2)
~> 2 / (fozzy - 1 - 1)
```

`fozzy` shouldn't be in scope in the final query either, even though it's Pure. I tried to "fix" this in https://github.com/ambiata/icicle/pull/121, hope I'm not doing something very wrong.

Anyhow, this PR should be subsumed by that one.

amosr:
Icicle/Source/Checker/Constraint.hs:
```
    Let _ n x
     -> do  (x',sx) <- generateX x
            (q',sq,tq) <- withBind n (annResult $ annotOfExp x') rest
```

because the "let fozzy" is using withBind, the fozzy should only be in scope of the 2
so if that fozzy is in scope there, it'd be a problem with the Let rule, not the GroupFold rule.

how did you try to fix this in #121? I can't see anything about removeElementBinds in #121.

(I agree that it's not quite correct, but in regards to polymorphic modes. I have a way to fix this, and a nicer explanation of the whole strange rule, but need to think about writing it down)

tranma:
Ah, I'm not sure if I did this right, then:

```
    Let _ n x	
+     -> do  (x', sx, ctx') <- generateX x ctx
+            let ctx'' = ctx' { ctxEnv = ctx Env }
+            (q',sq,tq,ctx') <- withBind n (annResult $ annotOfExp x') ctx'' rest
```

so to generate `rest` I just use the original environment plus the constraints generated from the let definition.

amosr:
that looks fine. "rest" here is only the stuff inside the let's scope, or more generally the stuff under this context:
```
foo (let x = 2 ~> x) (let y = 3 ~> y)
```
here, ``let x = 2 ~> x``, the ``x`` is the rest, and after typechecking the rest we return, the parent uses the original context, so it's out of scope

amosr:
oh wait. why is generateX returning the ctx'? I think generateX should only have to return the constraints - not the new environment.
I'll move this conversation to #121.

PR #113, opened 2015-09-28
Use the annotated pretty printer

HuwCampbell:
Just keeping this simple for now, subtle additional annotations can be added later.

### DIFF
--- a/icicle.cabal
+++ b/icicle.cabal
@@ -34,7 +34,7 @@ library
                      , unordered-containers            == 0.2.*
                      , validation                      == 0.5.*
                      , vector                          == 0.10.*
-                     , wl-pprint                       == 1.2.*
+                     , annotated-wl-pprint             == 0.7.*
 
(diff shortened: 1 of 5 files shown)


tranma:
what does this do?

amosr:
woohoo! :pray:

amosr:
I think you'll need to update the tests.
```
$ test/cli/repl/run
# eyeball the differences. probably only escape codes
$ test/cli/repl/run UPDATE
```

tranma:
I do `git add -p` afterwards and :eyes: each chunk really hard

amosr:
that's probably a lot easier to read...

jystic:
:dromedary_camel: nice :)

HuwCampbell:
Yep, done.
I removed one of the annotation, as it was too high up, and the current reset function doesn't pop, but rather resets completely, so it was a bit ugly.

PR #114, opened 2015-09-28
Don't do a build in testci

amosr:


amosr:
Merging this second cabal change because it fixes the problem with the first one. lol.

PR #116, opened 2015-09-28
Topic/tombstone feature

HuwCampbell:
WIP.
@jystic Not sure where Jetski is supposed to be.
Also there's some java which I have put a todo on.

### DIFF
--- a/data/libs/prelude.icicle
+++ b/data/libs/prelude.icicle
@@ -15,3 +15,6 @@ sd v = let fold a = ((0, 0.0), 0.0)
 # Time statements
 newest v = latest 1 ~> let fold1 s = v : v ~> s.
 oldest v = let fold1 s = v : s ~> s.
+
+# Boolean statements
+not a = case a | True -> False | False -> True end.
--- a/test/cli/repl/t20-lexer/expected
+++ b/test/cli/repl/t20-lexer/expected
@@ -1,5 +1,5 @@
 welcome to iREPL
-ok, loaded 9 functions from data/libs/prelude.icicle
+ok, loaded 10 functions from data/libs/prelude.icicle
 ok, loaded test/cli/repl/data.psv, 8 rows
 > -- Numbers
 > - Core evaluation:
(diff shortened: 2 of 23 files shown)


data/libs/prelude.icicle
```diff
@@ -15,3 +15,6 @@ sd v = let fold a = ((0, 0.0), 0.0)
 # Time statements
 newest v = latest 1 ~> let fold1 s = v : v ~> s.
 oldest v = let fold1 s = v : s ~> s.
+
+# Boolean statements
+not a = let r = case a | True -> False | False -> True end ~> r.
```
tranma: this can just be `not a = case a | True -> False | False -> True end.`

test/cli/repl/t20-lexer/expected
```diff
@@ -30,7 +30,7 @@ Check error:
       mean                 : (Num a) => Element a -> Aggregate Double
       date                 : Element Definitely DateTime
       min                  : Element a -> Aggregate Possibly a
-      sd                   : (Element =: ReturnOfLet Pure a) => a Double -> Aggregate Double
+      not                  : (t$179 =: TemporalityJoin a Pure) => a Bool -> Pure Bool
```
tranma: doh, leftover constraints =(

data/libs/prelude.icicle
HuwCampbell: Yep, I actually already did that one.

jystic:
@HuwCampbell it's fine if you upgrade it, the only difference is the merge commit, my bad for not doing this already.

HuwCampbell:
@amosr would you mind having a look at this. I think someone else will have to do the Java generation part, but apart from that I think it is ok.

amosr:
Yep, this looks good. I suspect this will all change in a little bit once I get the possibly shit going, but for now I'm happy to merge this.
Don't worry about the java generation either, that should fall out of the possibly stuff.

amosr:
:children_crossing:

PR #117, opened 2015-09-29
Add 'now' feature so one can access the snapshot/chord date

HuwCampbell:
@amosr This is what I have come up with for now (puns!)
We still need to expose DaysDifference in source so it's actually useful.

### DIFF
--- a/src/Icicle/Source/ToCore/Base.hs
+++ b/src/Icicle/Source/ToCore/Base.hs
@@ -58,24 +58,25 @@ programOfBinds
     :: OutputName
     -> ValType
     -> CoreBinds a n
+    -> Maybe (Name n)
     -> a
     -> Name n
     -> C.Program a n
-programOfBinds outputName inpType binds a_ret ret
+programOfBinds outputName inpType binds postdate a_ret ret
  = C.Program
  { C.input      = inpType
  , C.precomps   = precomps  binds
  , C.streams    = streams   binds
  , C.reduces    = reduces   binds
  , C.postcomps  = postcomps binds
- , C.postdate   = Nothing
+ , C.postdate   = postdate
  , C.returns    = [(outputName, X.XVar a_ret ret)]
  }
 
 instance Monoid (CoreBinds a n) where
  mempty = CoreBinds [] [] [] []
- mappend (CoreBinds a b c d) (CoreBinds e f g h)
-  = CoreBinds (a<>e) (b<>f) (c<>g) (d<>h)
+ mappend (CoreBinds a b c d) (CoreBinds f g h i)
+  = CoreBinds (a<>f) (b<>g) (c<>h) (d<>i)
 
 
 pre :: Name n -> C.Exp a n -> CoreBinds a n
@@ -89,8 +90,6 @@ red n x = mempty { reduces = [(n,x)] }
 
 post :: Name n -> C.Exp a n -> CoreBinds a n
 post n x = mempty { postcomps = [(n,x)] }
-
-
 
 data ConvertError a n
  = ConvertErrorNoSuchFeature (Name n)
--- a/src/Icicle/Source/ToCore/ToCore.hs
+++ b/src/Icicle/Source/ToCore/ToCore.hs
@@ -80,9 +80,13 @@ convertQueryTop feats qt
                   Just t' -> return t'
         let inpTy'dated = T.PairT inpTy T.DateTimeT
 
-        (bs,ret) <- evalStateT (convertQuery $ query qt) (ConvertState inp inpTy'dated fs Map.empty)
+        (now,bs,ret) <- evalStateT (do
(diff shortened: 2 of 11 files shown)


src/Icicle/Source/ToCore/Base.hs
```diff
@@ -90,7 +91,8 @@ red n x = mempty { reduces = [(n,x)] }
 post :: Name n -> C.Exp a n -> CoreBinds a n
 post n x = mempty { postcomps = [(n,x)] }
 
-
+pstdt :: Maybe (Name n) -> CoreBinds a n
+pstdt n = mempty { postdate = n }
```
tranma: haha, unix naming scheme: 1. remove all vowels, 2. remove all useful letters, 3. repeat until un-pronounceable

src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -80,8 +80,12 @@ convertQueryTop feats qt
                   Just t' -> return t'
         let inpTy'dated = T.PairT inpTy T.DateTimeT
 
-        (bs,ret) <- evalStateT (convertQuery $ query qt) (ConvertState inp inpTy'dated fs Map.empty)
-        let bs'   = strm inp C.Source <> bs
+        (now,bs,ret) <- evalStateT (do
```
tranma: blank space

amosr: lol

HuwCampbell: https://www.youtube.com/watch?v=e-ORhEE9VVg

amosr: last night I was talking about taylor swift and laurel said "I've never heard a taylor swift song" and I said "neither have I". no longer can I make that claim, and I hold grudges for a long time

HuwCampbell: lol

domdere: its ok @amosr  misery loves company and youtube vids are just so easy to trick people into watching....

src/Icicle/Source/ToCore/Base.hs
amosr: post straumatic tress disorder

HuwCampbell: it was named after `strm`.

amosr: I sure do love a slice of humble pie

tranma: slhblp

amosr:
this looks good, but I'm not 100% sure about the monoid instance on postdates. what happens if both are "Just"? maybe you'll need to add a new postcomputation binding, setting the second date to the first?

hm can you add now to the arbitrary instance for Source? might be easier to change the Nothing in fets to just some arbitrary variable

HuwCampbell:
Yeah, Monoids on Maybes is a bit terrible. The only real thing we could do it fake a let binding so the next value points to the real one. It's a bit dumb though.
Arbs added.

amosr:
yeah - I think adding lets is good enough

HuwCampbell:
BTW, this PR is on top of the Tombstone one.

HuwCampbell:
Hmm. I don't think it's too simple. The Monoid is used after the convertQuery is called. So it'd be pretty hard to reach back in and add a Let.

amosr:
I was thinking something like
```
mappend (CoreBinds a b c d (Just date1)) (CoreBinds f g h i (Just date2))
 = CoreBinds (a<>f) (b<>g) (c<>h) ((date2, X.XVar date1) : d<>i) (Just date1)	
```
but the other cases same.

ah! but no annotation! damn! maybe you can add an annotation to the CoreBinds (Maybe (a, Date)) field?

HuwCampbell:
Ok, I don't see how it could possibly get called, but I've put it in.

amosr:
ah ok, I misunderstood. I thought each use of "now" was adding a Just to the CoreBinds' date, rather than only being set once at the top level.
yeah - in this case we're better off removing the Maybe PostDate from the CoreBinds, and just adding it as another argument to programOfBinds - since the intermediate steps that also produce CoreBinds don't need to worry about the postdate.

this is good though :microscope:

tranma:
I think postdate shouldn't exist too, but as it is this is still good :hatching_chick: :hatched_chick: :baby_chick: :chicken: :poultry_leg:

HuwCampbell:
Ahh, yes, good point @amosr. Functions with arguments, love it :P

amosr:
tran is currently fixing the mess I created of the typechecker by refusing to use function arguments. oh boy :-(

HuwCampbell:
Just going to close this and create a new PR on top of master.

tranma:
"fix" is a strong word...

PR #120, opened 2015-09-30
Topic/windows

HuwCampbell:
Some of the code is a little ugly still, but this seems to be functioning well.

### DIFF
--- a/src/Icicle/Avalanche/FromCore.hs
+++ b/src/Icicle/Avalanche/FromCore.hs
@@ -186,9 +186,18 @@ insertStream namer inputType strs reds (n, strm)
        CS.SWindow _ newerThan olderThan inp
         -> let
                genDiff :: WindowUnit -> (X.Exp () n, X.Exp () n)
-               genDiff (Days   d) = (constI d     ,xPrim (PrimMinimal $ Min.PrimDateTime Min.PrimDateTimeDaysDifference))
-               genDiff (Weeks  w) = (constI (7*w) ,xPrim (PrimMinimal $ Min.PrimDateTime Min.PrimDateTimeDaysDifference))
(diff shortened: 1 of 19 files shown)


tranma:
TLDR for the silly...? what's the difference between this and the "windowed" context?

HuwCampbell:
It's on the Windowed, but changes months to actually reflect months, and simplifies the SWindow type.

tranma:
ah, before if was just handling days?

HuwCampbell:
Yeah, months was just 30 days.

src/Icicle/Avalanche/FromCore.hs
```diff
@@ -183,20 +184,36 @@ insertStream namer inputType strs reds (n, strm)
...
-               check  | Just o' <- olderThan
+               (>~~) = prim2 (PrimMinimal $ Min.PrimRelation Min.PrimRelationGt DateTimeT)
+               infix 4 >~~
+
+               (>=~~) = prim2 (PrimMinimal $ Min.PrimRelation Min.PrimRelationGe DateTimeT)
+               infix 4 >=~~
+
+               (<=~~) = prim2 (PrimMinimal $ Min.PrimRelation Min.PrimRelationLe DateTimeT)
+               infix 4 <=~~
+
```
amosr: Maybe we can split all this out to a separate function, ``windowPredicate :: Namer n -> WindowUnit -> Maybe WindowUnit -> Exp () n Prim``

amosr:
the Avalanche.FromCore can be cleaned up, but otherwise :+1:

src/Icicle/Avalanche/FromCore.hs
tranma: trying to compete with `lens`!

HuwCampbell: Yeah, it's the ugliest bit. Got it working, need to make it not terrible before I merge.

PR #121, opened 2015-09-30
remove state from the source type checker

tranma:
why we want to do this: hiding both the name environment and the constraint set in a global state makes it sometimes difficult to determine which environment/constraint go where. There are some places where we had to figure out what to *remove* from the current context before doing something.

This change attempts to make it explicit what names/constraints are being brought into scope...

### DIFF
--- a/src/Icicle/Source/Checker/Constraint.hs
+++ b/src/Icicle/Source/Checker/Constraint.hs
@@ -218,8 +218,9 @@ generateQ qq@(Query (c:_) _) env cons
             retk <- Temporality TemporalityElement . TypeVar <$> fresh
             retv <- Temporality TemporalityElement . TypeVar <$> fresh
 
+            let env' = removeElementBinds env
             (q', sq, t', consr)
-                <- withBind k retk env
+                <- withBind k retk env'
                  $ (fmap flip . withBind) v retv (rest consg)
 
             cons'  <-  requireAgg  t' consr
--- a/src/Icicle/Source/Checker/Constraint.hs
+++ b/src/Icicle/Source/Checker/Constraint.hs
@@ -1,6 +1,8 @@
 -- | Generate type constraints
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE DoAndIfThenElse #-}
 module Icicle.Source.Checker.Constraint (
     constraintsQ
   , generateQ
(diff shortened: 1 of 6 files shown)


src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -93,21 +103,25 @@ constraintsQ env q
...
+-- nested subexpressions and so on.
 --
--- Shoehorning these two differently flowing contexts into a state monad was a mistake, and is a constant source of bugs.
---
-generateQ :: Ord n => Query a n -> Gen a n (Query'C a n, SubstT n)
+generateQ
+  :: Ord n
+  => Query a n
+  -> CheckContext a n
+  -> Gen a n (Query'C a n, SubstT n, CheckContext a n)
```
amosr: we want something like this type for all of the places where the environment/context was used before:
```
=> Query a n
-> Map n FunctionType
-> Gen a n (Query'C a n, SubstT n, [Constraint a n])
```
because the Environment only goes down lexically, and the constraints and type substitutions just have to bubble up

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -154,84 +172,109 @@ generateQ qq@(Query (c:_) _)
...
+            let q''   = with cons' q' t' $ \a' -> Latest a' i
+            return (q'', sq, CheckContext env cons')
 
     -- >   group (Element k'p k'd) ~> Aggregate v'p v'd
     -- > : Aggregate (PossibilityJoin k'p v'p) (Group k'd v'd)
     GroupBy _ x
-     -> do  (x',sx) <- generateX x
-            (q',sq,tval) <- rest
+     -> do  (x', sx, ctxKey)                      <- generateX x ctx
+            (q', sq, tval, CheckContext env cons) <- rest (ctxKey { ctxEnv = qenv })
```
amosr: so I guess here you're passing ctxKey in order to use its constraints, so that rest returns the union of all constraints. however, you can just append the constraints together at this level (like we do with the substitutions)

amosr: (this is still correct - just more complicated than necessary)

src/Icicle/Source/Checker/Constraint.hs
tranma: yes, I wasn't quite sure if I would not need the Environment anywhere, so I make it keep both the Environment + Constraint like the state monad was doing. Now I had a better idea so I'll only return the constraint set. Sorry, so clueless =(

amosr: makes sense - it's a refinement in the right direction

tranma:
should be ok now... /cc @amosr

amosr:
:+1:

PR #122, opened 2015-10-01
Add common logical operators

HuwCampbell:
I've got a feeling that this will make describing complex filters a lot easier.

### DIFF
--- a/src/Icicle/Source/Checker/Prim.hs
+++ b/src/Icicle/Source/Checker/Prim.hs
@@ -36,6 +36,11 @@ primLookup' p
     Op (Relation _)
      -> f1 $ \a at -> FunctionType [a] [] [at, at] BoolT
 
+    Op (LogicalUnary _)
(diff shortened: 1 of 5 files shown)


src/Icicle/Source/Checker/Prim.hs
```diff
@@ -81,6 +86,9 @@ primLookup' p
   fNum f
    = f1 (\a at -> uncurry (FunctionType [a] [CIsNum at]) (f at))
 
+  fBool f
+   = f1 (\a at -> uncurry (FunctionType [a] [CEquals BoolT at]) (f at))
```
amosr: this can just call f with BoolT - it doesn't need a constraint for that.
in fact the type can just be
```FunctionType [] [] [BoolT, BoolT] BoolT```

amosr:
I like it when it's easy to add things
:+1:

PR #123, opened 2015-10-01
Add some nicer messages for shunt errors

HuwCampbell:
Sorry about this one, it's... not important.


amosr:
damn - I wish we could use real error message datatypes for Parsec.
:+1:

amosr:
actually - maybe I just want to wrap the Parsec in an EitherT for other error messages

PR #124, opened 2015-10-06
fix all the things

tranma:
* fix the type checker bug introduced by previous branch
* maybe simplify a bit by not passing the constraints to generateX/Q as arguments?

### DIFF
--- a/src/Icicle/Source/Checker/Function.hs
+++ b/src/Icicle/Source/Checker/Function.hs
@@ -22,7 +22,6 @@ import                  Control.Monad.Trans.Either
 
 import qualified        Data.Map                as Map
 import qualified        Data.Set                as Set
-import qualified        Data.List               as List
 
 checkFs :: Ord n
         => Map.Map (Name n) (FunctionType n, Function (Annot a n) n)
@@ -65,10 +64,10 @@ checkF' fun env cons
       -- Look up the argument types after solving all constraints.
       -- Because they started as fresh unification variables,
       -- they will end up being unified to the actual types.
-      (args, cs) <- List.unzip <$> mapM (lookupArg subs env' cons') (arguments fun)
+      args <- mapM (lookupArg subs env' cons') (arguments fun)
 
       -- Find all leftover constraints and nub them
-      let constrs = ordNub $ fmap snd (concat cs)
+      let constrs = ordNub $ fmap snd cons'
 
       -- We want to remove any modes (temporalities or possibilities)
       -- that are bound by foralls with no constraints on them.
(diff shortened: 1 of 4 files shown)


src/Icicle/Source/Checker/Function.hs
```diff
@@ -59,10 +60,15 @@ checkF' fun env cons
...
+      (q', subs, cons')  <- generateQ (body fun) env' cons
+
+      -- Look up the argument types after solving all constraints.
+      -- Because they started as fresh unification variables,
+      -- they will end up being unified to the actual types.
+      (args, cs) <- List.unzip <$> mapM (lookupArg subs env' cons') (arguments fun)
 
       -- Find all leftover constraints and nub them
-      let constrs = ordNub $ fmap snd cons'
+      let constrs = ordNub $ fmap snd (concat cs)
```
amosr: what's the difference between using ``concat cs`` and ``cons'`` here?

tranma: `cs` has all the additional constraints from `lookup` the arguments (and hence introducing the foralls). Hm, maybe it's not necessary.

amosr: hm ok - are there any of those? it includes ``cons'`` too though? that's ok then

amosr: ah got it. I don't think there will be any constraints attached to the arguments, because they're bound as non-functions. this is ok though.

amosr:
:beers:

tranma:
it's a bit strange that the test case `feature salary ~> (filter value > 10 ~> count), (latest 3 ~> value)` now has a less specific type though.

amosr:
oh, is that this weird one?
```Aggregate Possibly ([32mt[0m$[32m10[0m, (Array Int))```

**Comment by tranma** [comment]
ah, it's exactly because I now use `cs` rather than `cons'`

**Comment by amosr** [comment]
hm ok

**Comment by tranma** [comment]
if there should be no constraints introduced by looking up the arguments, I'll revert to using `cons'` as before, the type for that test case is the same then.

**Comment by amosr** [comment]
sounds good

File: test/cli/repl/t20-lexer/expected [hunk 2]
```diff
@@ -30,7 +30,7 @@ Check error:
       [32mmean[0m                 : (Num [32ma[0m) => Element [32ma[0m -> Aggregate Double
       [32mdate[0m                 : Element Definitely DateTime
       [32mmin[0m                  : Element [32ma[0m -> Aggregate Possibly [32ma[0m
-      [32mnot[0m                  : ([32mt[0m$[32m179[0m =: TemporalityJoin [32ma[0m Pure) => [32ma[0m Bool -> Pure Bool
```
tranma: and this has a better type. Good I suppose.

amosr:
perfect. merge!

PR #125, opened 2015-10-06
WIP: Reify Possibility and typechecker fixes

amosr:
In retrospect, it is a shame that I made these typechecker changes in this branch instead of a fresh one, as the current state of Reify is not passing any tests


src/Icicle/Common/Base.hs
```diff
@@ -49,17 +49,22 @@ data BaseValue
...
- | VTombstone
 
- | VException ExceptionInfo
+ | VError ExceptionInfo
  deriving (Show, Ord, Eq)
 
+-- | Called "exceptions"
+-- because they aren't really errors,
+-- but they aren't really good values either..
 data ExceptionInfo
```
tranma: So exceptions in Core aren't errors, but special values? (errors are the `left`s after ReifyPossibilities?)

amosr: yeah, it'll just be an enum attached to a Left or something. so reify will convert potentially failing computations to ``Either Error t``

amosr:
might merge this...

src/Icicle/Source/Checker/Constraint.hs
```diff
@@ -518,39 +537,38 @@ generateP
...
+        let conss      = require (annotOfExp alt) (CEquals datS t)
 
-        (alt', sub, consa) <- generateX alt envp conss
+        (alt', sub, consa) <- generateX alt envp
 
-        let altTy'  = annResult (annotOfExp alt')
-        altTy      <- TypeVar <$> fresh
-        altTp      <- TypeVar <$> fresh
+        let altTy' = annResult (annotOfExp alt')
+        let altTp  = getTemporalityOrPure altTy'
```
tranma: ok, I see. I tried something similar, but I think I tried to join temporalities on the whole type, rather than unwrapping first here. Anyway, this is good.

tranma:
seems fine on a cursory inspection :+1:

src/Icicle/Source/Checker/Constraint.hs
amosr: yeah - and adding the "a := TemporalityJoin b Pure => a := b" rule got rid of a bunch of unnecessary constraints too. yay!

PR #126, opened 2015-10-07
WIP: Reintroduce ReifyPossibilities

amosr:


test/Icicle/Test/Source/Arbitrary.hs
```diff
@@ -134,8 +134,9 @@ instance Arbitrary n => Arbitrary (Fold (Query () n) () n) where
 instance Arbitrary FoldType where
  arbitrary
   = oneof
-        [ return FoldTypeFoldl1
-        , return FoldTypeFoldl ]
+        -- Disable foldl1 for now as it needs to be put through desugar first
+        -- [ return FoldTypeFoldl1
```
tranma: fold1 needs to be desugared first?

src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -646,16 +580,28 @@ convertReduce xx
...
-        let bs' = fst scrut'
+        -- Because the alternatives can contain more folds and stuff, we need to
+        -- use convertReduce on them.
+        -- However, because the pattern variables must be Aggregates, we know that
+        -- any pattern variables will only be mentioned in the postcomputations.
+        -- Therefore we need to pull out the postcomputations into a let,
+        -- and stick them inside the new case alternative.
+        --
+        -- All the foldy bits of each alternative must be computed, because
+        -- we won't know which ones will be needed until after they are run.
```
tranma: I agreed with that the postcomputations should be squashed into the one `let`, but I'm not sure I understand what sticking them inside a new alternative means?

Also, after çase desugar, there should be no more pattern variables. They should all become let's, e.g. `case x | y -> ...` should become `let y = x in ...`

test/Icicle/Test/Source/Arbitrary.hs
amosr: hmm reify actually. should be reified first

src/Icicle/Source/ToCore/ToCore.hs
amosr: I'm not sure how there can be no pattern variables: what does
```
case x
| Left i -> …
| Right j -> …
```
become?

the reasoning here, anyway, is that if you have
```
case Left (sum value)
| Left a -> (mean value) * a
| Right b -> (min value) * b
```
then the postcomputations of the alternatives will actually just be something like "comp$mean$value * a" and "comp$min$value * b", so you need to pull them back inside the actual case statement underneath where "a" and "b" are bound

tranma: oh, blergh, sorry, I meant top-level pattern variables. Ok, I didn't think of the situation where the scrutinee is some expression, and the pattern binds the result.

test/Icicle/Test/Source/Arbitrary.hs
amosr: really I should just put the call to reify in the tests - but I want to think about making all that pipeline stuff nicer first

amosr:
"seems ok", except for latests. but given that @tranma is already working on latests I might as well leave that until afterwards

amosr:
ok. it has been slow moving, but I think this is finally close enough, except for latest-of-elements. I'll do some more testing tomorrow & hopefully merge.

amosr:
close enough to merge :sunglasses: :cruisecontrolforcool:

test/cli/repl/t15-tombstones/expected
```diff
@@ -5,11 +5,7 @@ ok, loaded test/cli/repl/data.psv, 13 rows
 > ok, loaded test/cli/repl/t15-tombstones/data.psv, 5 rows
 > ok, loaded 1 functions from test/cli/repl/t15-tombstones/numflips.icicle
 > > - Core evaluation:
-[gonzo, [("a",False)
-        ,("b",False)
-        ,(tombstone,True)
-        ,(tombstone,True)
-        ,("c",False)]]
+[gonzo, tombstone]
```
amosr: this is probably bad

amosr:
mergin

PR #127, opened 2015-10-08
Topic/more date ops

HuwCampbell:
This kind of abuses the operators, but the functionality is as it should be.

### DIFF
--- a/src/Icicle/Source/Parser/Token.hs
+++ b/src/Icicle/Source/Parser/Token.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PatternGuards #-}
 module Icicle.Source.Parser.Token (
     Parser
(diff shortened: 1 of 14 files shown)


src/Icicle/Source/Parser/Token.hs
```diff
@@ -53,7 +54,11 @@ pKeyword kw
 
 pOperator :: Parser T.Operator
 pOperator
- = pTok get <?> "operator"
+ = pTok get
+   <|> (pKeyword T.Days *> (T.Operator ("days before") <$ pKeyword T.Before <|> T.Operator ("days after") <$ pKeyword T.After))
+   <|> (pKeyword T.Weeks *> (T.Operator ("weeks before") <$ pKeyword T.Before <|> T.Operator ("weeks after") <$ pKeyword T.After))
+   <|> (pKeyword T.Months *> (T.Operator ("months before") <$ pKeyword T.Before <|> T.Operator ("months after") <$ pKeyword T.After))
+   <?> "operator"
```
tranma: seems ok, it's not that bad...

tranma:
:+1: I think using "operators" in this way is ok...

amosr:
:+1: repl test so we have an example of using it, too.
what happens if you put two spaces in - or line break?
``3 months   before now``
etc

HuwCampbell:
Yep, I have actually been writing up window tests using it, but I'll do one in icicle proper's too.
No problem for multiple spaces or newline, the lexer skips just gives back the tokens.

amosr:
oh man
```
3 months       before              now
```
I thought code mode would respect whitespace. weird

amosr:
ok. double doinks do but triple doinks dont

PR #128, opened 2015-10-08
Relax temp join constraints a bit

tranma:
if either argument is a type var, defer.


src/Icicle/Source/Type/Constraints.hs
```diff
@@ -94,15 +94,18 @@ dischargeC c
      | btemp == ctemp
      -> dischargeC $ CEquals atemp btemp
 
-    CTemporalityJoin (TypeVar _) _          (TypeVar _)
+    CTemporalityJoin (TypeVar _) _ (TypeVar _)
      -> return $ DischargeLeftover c
     CTemporalityJoin (TypeVar _) (TypeVar _) _
```
amosr: actually, can probably remove these cases now - they'll be covered by the two new ones

amosr:
:green_apple: groovy

amosr:
yeah, that makes perfect sense

tranma:
mergerised

PR #129, opened 2015-10-08
Fix tombstone check to be (== Tombstone)

amosr:
In ReifyPossibilities, the tombstone check needs (== Left Tombstone) but before that is introduced it should be just (== Tombstone)

### DIFF
--- a/src/Icicle/Dictionary/Data.hs
+++ b/src/Icicle/Dictionary/Data.hs
@@ -131,7 +131,7 @@ featureMapOfDictionary (Dictionary { dictionaryEntries = ds, dictionaryFunctions
    = X.XPrim () (X.PrimMinimal $ X.PrimStruct $ X.PrimStructGet f t fs)
   xtomb t1
    = X.XApp () (X.XPrim () (X.PrimMinimal $ X.PrimRelation X.PrimRelationEq t1))
-               (X.XValue () (SumT ErrorT t1) (VLeft $ VError ExceptTombstone))
+               (X.XValue () t1 (VError ExceptTombstone))
 
   exps str e'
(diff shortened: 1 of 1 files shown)


tranma:
tried with `feature gender ~> (latest 4 ~> value, tombstone)`? :+`

amosr:
hehe no I just made this change in the web interface because I'm in the middle of reify stuff. :lollipop:

tranma:
tried for you (use `git stash`!)

```
> feature gender ~> (latest 4 ~> value, tombstone)
- Type:
Aggregate Possibly (Array (String, Bool))

- Core:
Program (source : Stream String)
Precomputations:

Stream transformers:
conv$0 =
    source
conv$3 =
    smap    [(String, DateTime)] [(String, Bool)] (
    \conv$4 : (String, DateTime).  pair# [String] [Bool]
                                   (fst# [String] [DateTime] conv$4) (eq# [String]
                                                                      (ExceptTombstone : String) (fst# [String] [DateTime] conv$4))) conv$0
Reductions:
conv$2 =
    rlatest [(String, Bool)] (4 : Int) conv$3
Postcomputations with date as now$conv$1:

Returning:
repl =
    conv$2


- Core evaluation:
[bart, [("m",False)]
,homer, [("m",False)]
,lisa, [("f",False),("NA",False),("m",False)]
,maggie, [("f",False)]
,marge, [("f",False)]]
```

amosr:
thanks!
actually - silly thing - you need to ``:dictionary data/example/DictionaryTrial.toml`` before loading the data. because otherwise it doesn't know "NA" is tombstone in that field.
no type error though so I'm happy

PR #132, opened 2015-10-08
Add a repl test for tombstone checking

HuwCampbell:


amosr:
:hibiscus: :hamburger:

Issue #82, opened 2015-08-31
Sliding windows


HuwCampbell:
It would be useful to have a sliding function which one can fold over.
For example, apart:
```feature age ~> sliding 2 ~> ```
would give access to a stream of two values at a time, so the difference in the date could be looked at for example, or how large the differences are.

HuwCampbell:
Had a hack today on trying this with just case expressions, but without higher order functions I don't think it's possible.

amosr:
I think we can still write meaningful expressions without this, so let's leave this for now

amosr:
there is a branch that started work on window frames, but was not finished.
this will be worth coming back to eventually, but for now Source is expressive enough to focus on Ivory integration.
https://github.com/ambiata/icicle/tree/topic/window-frames

PR #133, opened 2015-10-08
Topic/window frames

HuwCampbell:
WIP.
Avalanche hasn't changed yet, but it's worth a look.


tranma:
if I remember correctly, these are windows with access to one previous element..?

HuwCampbell:
That's right

tranma:
oh, now I get it, "frames" around a window.

HuwCampbell:
So there's a few things still needing to be done here and I wouldn't mind chatting about them.

Firstly: avalanche isn't implementing this.

Secondly: I'm not sure what to do with the date of the feature which is seen "at the window frame". There's no need to know when said feature was dropped, and (because "now" is an aggregate feature) it's not possible to easily know the window's edge without exposing bubblegum filtering issues. On the other hand, fudging the date such that it is at the edge of the window could be dangerous.

Thirdly: if a feature is tombstoned out before the start of the frame, we will see that tombstone, but if there was just no fact written, we won't see anything until inside the window. My intuition is that these two scenarios should give the same result, but here they probably won't.

amosr:
(2)

it seems like, depending on the query, you'd actually want to know both the date of the outside fact, as well as the frame date.
for example, if you were doing a gradient or integral might you want to interpolate between the outside value and the first inside value to get the value at start of frame?

you're right, I don't know whether the "start window" date would be useful as an aggregate, it seems a bit unwieldy. the bubblegum filtering could be dealt with by putting everything in the window in the bubblegum. that is, any predicate filters would filter out the values, but leave the bubblegums in.

tranma:
what's happening with this?

amosr:
if you can update the last previous date to be the edge of the window, and add Avalanche evaluation, I'd be happy for this to be merged

amosr:
this is related to #82. this will be good eventually but closing for now

PR #134, opened 2015-10-09
clean up compiler pipeline a bit

tranma:
just simple ripping out of compiler plumbing stuff from REPL and dictionary loader.

should add some top-level functions that does desugar, etc. for both REPL and dict loader to use. Currently dict loader doesn't do any of it.


src/Icicle/Pipeline.hs
```diff
@@ -0,0 +1,181 @@
...
+  funs      = M.map snd
+            $ D.dictionaryFunctions d
+  inline q' = snd
+            $ Fresh.runFresh
+                (STI.inlineQT funs q')
+                (freshNamer "inline")
+
+
+sourceParseConvert :: Text -> Text -> Either CompileError Program'
+sourceParseConvert base t
```
amosr: don't know whether this is used anywhere. guess it should desugar and reify before converting

amosr:
so I wonder whether the pipeline thing could have a config datatype something like
```
data PipelineConfig
 = PipelineConfig
 { dump :: Dump
 , opt :: Optional
 , eval :: Eval
 }

data Dump
 = Dump
 { dumpDesugar :: Bool
 , dumpReify :: Bool
 , dumpInline :: Bool
 , dumpAnnotated :: Bool
 , dumpCore :: Bool
 , dumpSea :: Bool
 … }
data Optional
 = Optional
 { optionalSimpSource :: Bool
 , optionalSimpCore :: Bool
 , optionalSimpAvalanche :: Bool
 }
data Eval
 = EvalCore | EvalAvalanche | EvalSea | EvalJava
```

then I guess a few things like

```
makeEvaluator :: PipelineConfig -> Query -> Eval…
```

hm not sure

tranma:
oh, so all of the REPL?

tranma:
I'm not sure either. Other users of this pipeline like the dictionary loader don't really need it.

amosr:
actually I think it'd be ideal if we could easily turn on dumps for when loading dictionary and evaluating from ice

src/Icicle/Repl.hs
```diff
@@ -19,120 +20,74 @@ module Icicle.Repl (
...
-     ReplErrorCheck ce
-      -> "Check error:" <> line
-      <> indent 2 (pretty ce)
-     ReplErrorConvert ce
-      -> "Convert error:" <> line
-      <> indent 2 (pretty ce)
-     ReplErrorDecode d
-      -> "Decode error:" <> line
+     ReplErrorCompile d
+      -> "Compile error:" <> line
```
HuwCampbell: Probably don't need this extra bit of text and indent. Plus
```
Compile error:
  ParseError:
    blahdiblah
```
is less clear that just `ParseError`

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -108,13 +109,12 @@ loadDictionary' parentFuncs parentConf parentConcrete fp
...
   let d' = Dictionary (concreteDefinitions <> parentConcrete) importedFunctions
-  let d'' = featureMapOfDictionary $ d'
 
   virtualDefinitions
     <- (\(a, q) -> do
-         (checked, _)  <- sourceCheck d'' q
+         (checked, _)  <- sourceCheck d' q
          let inlined    = sourceInline d' checked
-         (checked', _) <- sourceCheck d'' inlined
+         (checked', _) <- sourceCheck d' inlined
```
HuwCampbell: Need a reify and desugar around here somewhere.

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -45,7 +45,8 @@ import qualified Text.Parsec                        as Parsec
 
 
 data DictionaryImportError
-  = DictionaryErrorIO E.SomeException
+  = DictionaryCompileError P.CompileError
+  | DictionaryErrorIO E.SomeException
   | DictionaryErrorParsecTOML Parsec.ParseError
   | DictionaryErrorParsecFunc Parsec.ParseError
   | DictionaryErrorParse [DictionaryValidationError]
```
HuwCampbell: There's probably one or two cases here that can be removed.

tranma:
@HuwCampbell you mentioned we should be able to load functions from a file, then run them on arbitrary data. That should be possible if we have something from the pipeline like: `runIcicleExp :: Dictionary -> QueryTop -> Result`? In the tests, you would just specify a query, then generate some data and put it in Dictionary?

HuwCampbell:
Data doesn't go in the dictionary, maybe just shove data into `Query` (not `QueryTop`)?

tranma:
updated. Not quite ideal (i.e. dictionary loader can't do dumps or anything fancy), but at least it's easier to add phases to both the REPL and the dictionary loader now.

the dictionary loader is now desugaring functions, but not reifying them.

test/cli/repl/t30-sea/script
```diff
@@ -8,4 +8,4 @@ feature salary ~> oldest value
 feature salary ~> filter value > 300 ~> count
 
 -- An interesting expression
-feature salary ~> sd_old value * (filter value < 300 ~> sd_old value)
+feature salary ~> let v_ = double value ~> sd v_ * (filter v_ < 300 ~> sd v_)
```
HuwCampbell: The cast shouldn't be needed anymore

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -136,36 +104,36 @@ loadDictionary' parentFuncs parentConf parentConcrete fp
...
-freshNamer :: T.Text -> Fresh.NameState SP.Variable
-freshNamer prefix = Fresh.counterPrefixNameState (SP.Variable . T.pack . show) (SP.Variable prefix)
+checkDefs d defs
+ = go `traverse` defs
+ where
+  go (a, q)
+   = do  (checked, _)  <- check' d q
+         let inlined    = P.sourceInline d checked
+         (checked', _) <- check' d inlined
+         pure $ DictionaryEntry a (VirtualDefinition (Virtual checked'))
```
HuwCampbell: Looks like the functions are desugared, but do we need to do the Virtuals as well?

HuwCampbell:
So in Ice I will add the ability to dump core and avalanche (it already can do source).

HuwCampbell:
Will need to run UPDATE again on the tests.

HuwCampbell:
Can we avoid turning off warnings please?

tranma:
? I didn't turn off warnings?

src/Icicle/Repl.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternGuards #-}
+{-# OPTIONS_GHC -fno-warn-missing-signatures #-}
```
HuwCampbell: Here

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -1,55 +1,49 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections     #-}
+{-# OPTIONS_GHC -fno-warn-missing-signatures #-}
```
HuwCampbell: And here

amosr:
might have been inadvertently copied from a test case, somehow. -no-warn-missing-signatures can be handy in tests, but for real code signatures are crucial imo

tranma:
dunno. Accidentally pressed a particular emacs combination? will delete.

tranma:
there are 2 desugar passes: desugar when importing functions, and desugar (+inline) an expression in the REPL, the latter runs after the former. Both of theses desugar passes were using the same `freshNamer "desugar"`, which caused problems due to name clashes. I've made them use different fresh name prefixes, not sure if that's the correct thing to do.

amosr:
hmm. that actually sounds like a bug in the inliner or something - the names of variables inside a function should not affect anywhere the function is used

tranma:
ok. holding off on this for now, will scrutiny it more later.

amosr:
well this change doesn't touch the inliner so chances are it was already broken. you can merge this if you'd like.

tranma:
nah, it's not the inliner. It's the desugar. Specifically the desugar introduces fresh `let`s for variable patterns: `, but the fresh name might shadow something.

amosr:
ahh! okay, that makes sense.
well, that is definitely a shame, but I think using different prefixes is a good solution for now

amosr:
just so I understand: the functions are desugared, then the repl expression has functions inlined, then the inlined is desugared?
I wonder (out of curiosity more than anything) whether we could:

1. desugar the repl expression *before* inlining, OR
2. if we could not desugar the functions, then inline into expr, then desugar both

I guess with 2 we wouldn't get desugar errors like exhaustive matches on functions; that's bad.
it seems like 1 should work but I'm not sure

anyway worth thinking about but I wouldn't bother changing it because the different prefixes work

tranma:
That's right. On master, (2) is the current situation: we do not desugar the functions, we inline them into exp, then desugar the whole query.

on this branch, we desugar the functions, inline them into exp, then desugar again. That's what caused the problem.

amosr:
ah, I see. I think desugaring the functions is good

tranma:
should I merge this?

amosr:
yeah :busts_in_silhouette:

PR #135, opened 2015-10-12
Topic/more prelude

HuwCampbell:


amosr:
neato :bicyclist:

data/libs/prelude.icicle
```diff
@@ -6,11 +6,48 @@ mean v = sum (double v) / count.
...
+minby a b = let fold1 s = (a, b)
+   :  case s | (a1, b1) -> (case a < a1 | True -> (a,b) | False -> (a1,b1) end) end
+   ~> case s | (a1, b1) -> b1 end.
+
+maxby a b = let fold1 s = (a, b)
+   :  case s | (a1, b1) -> (case a > a1 | True -> (a,b) | False -> (a1,b1) end) end
+   ~> case s | (a1, b1) -> b1 end.
+
+# Numerically unstable standard deviation. Best to not use.
 sd_old v = let v_ = double v ~> let sv = (sum v_) ~> (((sum (v_*v_) * count) - (sv * sv)) / (count * (count - 1)))^(0.5).
```
tranma: can't remember why I kept this. Maybe a test was using it?

data/libs/prelude.icicle
```diff
@@ -18,3 +60,20 @@ oldest v = let fold1 s = v : s ~> s.
 
 # Boolean statements
 not a = case a | True -> False | False -> True end.
+
+# Version of num flips which should fail upon finding a tombstone value.
+numflips v = let fold1 s = (v,0) : case s | (s0,i) -> (case (s0 == v) | True -> (v,i) | False -> (v,i+1) end) end ~> case s | (s0,i) -> i end.
+
+# Version of num flips which will accommodate tombstone values. Must have the tombstone feature passed to it.
+numflips' v t
```
charleso: Test for `numflips`, 'minby` and co?

HuwCampbell: Yeah, so I was discussing with Amos a better way to test these functions. Because I don't think the current way is very good.
Maybe a test which runs arbitrary data through R and icicle?

charleso: > Maybe a test which runs arbitrary data through R and icicle?

As discussed IRL, the tricky/dangerous part is that the tests now require that `R` is installed. I don't have any great suggestions, but I would suggest looking for an alternative first.

tranma: send a request to wolfram alpha?

data/libs/prelude.icicle
tranma: I think we can just remove this and "update" any test that might use this `sd_old`?

PR #137, opened 2015-10-13
Fix tests

HuwCampbell:


tranma:
:green_apple: should merge so other branches can go

PR #139, opened 2015-10-14
Topic/get optional

HuwCampbell:
Mostly usable, but there's a bug in the constraints checker (see repl tests).

### DIFF
--- a/test/cli/repl/t16-prelude/expected
+++ b/test/cli/repl/t16-prelude/expected
@@ -1,5 +1,5 @@
 welcome to iREPL
-ok, loaded 18 functions from data/libs/prelude.icicle
+ok, loaded 21 functions from data/libs/prelude.icicle
 ok, loaded test/cli/repl/data.psv, 13 rows
 > -- Sum example
 > - Core evaluation:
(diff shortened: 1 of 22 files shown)


test/cli/repl/t16-prelude/expected
```diff
@@ -53,4 +53,20 @@ ok, loaded test/cli/repl/data.psv, 13 rows
...
+> - Core evaluation:
+[homer, [False],marge, [False]]
+
+> > --GetOrElse example
+>                     [34mλλλλ[0m
+REPL Error:
+Check error:
+  Unsolvable constraint at  1:19
+    1:19  [32mt[0m$[32m502[0m =: TemporalityJoin [32mt[0m$[32m0[0m Aggregate
+    1:19  [32mt[0m$[32m3[0m =: TemporalityJoin Aggregate [32mt[0m$[32m502[0m
```
tranma: sounds like it could be fixed by Amos's https://github.com/ambiata/icicle/commit/a95b5659df8897220d3c72fe37ab93759cbb4dea? (you can `git cherry-pick` that hash)

amosr: yeh - I'm not actually sure whether these leftover constraints should be an error, either

amosr:
:meat_on_bone:
yeah just merge regardless of the bad constraint

amosr:
I also wonder whether we should modify the tests to use a special test-prelude, so we don't have to change the "loaded 21 functions" every time... eh or just don't print that

PR #145, opened 2015-10-19
change latest to be a fold on buffers

tranma:
...with an abstract buf_t type to be done in C and #included some time later.

### DIFF
--- a/src/Icicle/Avalanche/Statement/Flatten.hs
+++ b/src/Icicle/Avalanche/Statement/Flatten.hs
@@ -287,16 +287,12 @@ flatX a_fresh xx stm
        | [i] <- xs
        -> flatX' i
        $  \i'
-       -> do   accN      <- fresh
-               let bufT   = BufT t
-               let accT   = Mutable
+       -> do   tmpN      <- fresh
                let fpBuf  = xPrim (Flat.PrimBuf $ Flat.PrimBufMake t)
-               stm'      <- stm (xVar accN)
+               stm'      <- stm (xVar tmpN)
 
                return
-                 $ InitAccumulator
-                     (Accumulator accN accT bufT (fpBuf `xApp` i'))
-                     (Read accN accN accT bufT stm')
+                 $ Let tmpN (fpBuf `xApp` i') stm'
 
        | otherwise
        -> lift $ Left $ FlattenErrorPrimBadArgs p xs
@@ -308,9 +304,10 @@ flatX a_fresh xx stm
        $  \e'
        -> flatX' buf
        $  \buf'
-       -> do   tmpN <- fresh
+       -> do   tmpN      <- fresh
                let fpPush = xPrim (Flat.PrimBuf $ Flat.PrimBufPush t) `xApp` buf' `xApp` e'
-               return $ Let tmpN fpPush emptyStm
+               return
+                 $ Let tmpN fpPush mempty
 
        | otherwise
        -> lift $ Left $ FlattenErrorPrimBadArgs p xs
@@ -320,12 +317,11 @@ flatX a_fresh xx stm
        | [buf] <- xs
        -> flatX' buf
        $  \buf'
-       -> do   accN <- fresh
-               let bufT   = BufT t
-               let accT   = Mutable
-               let fpRead = xPrim (Flat.PrimBuf $ Flat.PrimBufRead t) `xApp` buf'
-               stm'      <- stm (xVar accN)
-               return $ Let accN fpRead stm'
+       -> do   tmpN       <- fresh
+               let fpRead  = xPrim (Flat.PrimBuf $ Flat.PrimBufRead t) `xApp` buf'
+               stm'       <- stm (xVar tmpN)
+               return
+                 $ Let tmpN fpRead stm'
 
        | otherwise
        -> lift $ Left $ FlattenErrorPrimBadArgs p xs
@@ -460,6 +456,3 @@ flatX a_fresh xx stm
             | otherwise
             = Min.PrimPairSnd ta tb
      in (xPrim $ Flat.PrimMinimal $ Min.PrimPair $ pm) `xApp` e
-
-  emptyStm
-   = Block []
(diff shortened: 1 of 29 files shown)


src/Icicle/Avalanche/Statement/Flatten.hs
```diff
@@ -283,6 +283,54 @@ flatX a_fresh xx stm
...
+       -> lift $ Left $ FlattenErrorPrimBadArgs p xs
+
+
+      Core.PrimLatest (Core.PrimLatestRead t)
+       | [buf] <- xs
+       -> flatX' buf
+       $  \buf'
+       -> do   accN <- fresh
+               let bufT   = BufT t
+               let accT   = Mutable
```
amosr: unused binding ?

src/Icicle/Avalanche/Statement/Flatten.hs
```diff
@@ -283,6 +283,54 @@ flatX a_fresh xx stm
...
+       -> do   accN      <- fresh
+               let bufT   = BufT t
+               let accT   = Mutable
+               let fpBuf  = xPrim (Flat.PrimBuf $ Flat.PrimBufMake t)
+               stm'      <- stm (xVar accN)
+
+               return
+                 $ InitAccumulator
+                     (Accumulator accN accT bufT (fpBuf `xApp` i'))
+                     (Read accN accN accT bufT stm')
```
amosr: bit confused by why this is initialising a mutable accumulator if it's never written to. could it just be a Let?
in the future, perhaps InitAccumulator and co should be removed and replaced with reference primitives

src/Icicle/Avalanche/Statement/Flatten.hs
```diff
@@ -413,3 +461,5 @@ flatX a_fresh xx stm
             = Min.PrimPairSnd ta tb
      in (xPrim $ Flat.PrimMinimal $ Min.PrimPair $ pm) `xApp` e
 
+  emptyStm
+   = Block []
```
amosr: mempty, I think

src/Icicle/Avalanche/Statement/Flatten.hs
tranma: sounds good

src/Icicle/Core/Eval/Exp.hs
```diff
@@ -85,6 +86,26 @@ evalPrim p vs
       | otherwise
       -> primError
 
+     PrimLatest (PrimLatestMake _)
+      | [VBase (VInt i)] <- vs
+      -> return . VBase . VBuf
+      $  List.replicate i (VError ExceptScalarVariableNotAvailable)
```
amosr: I'd rather this as an Int size, and the list of things in it

src/Icicle/Core/Eval/Reduce.hs
```diff
@@ -73,25 +73,6 @@ eval reduction_name xh sh r
...
-                 -- Take the latest i 
-              -> let sv' = latest i $ fst sv
-                 -- and split into history and values
-                 in  return ( BubbleGumFacts $ flavoursOfSource sv'
-                            , VArray $ fmap (snd.fact) sv' )
-
-             _
-              -> Left (SV.RuntimeErrorExpNotOfType num' IntT)
-
  where
```
amosr: this is my favourite bit

src/Icicle/Source/Checker/Invariants.hs
```diff
@@ -40,10 +40,12 @@ invariantQ ctx (Query (c:cs) xfinal)
...
      -> goNotAllowed
      | otherwise
-     -> errNotAllowed "Latest inside a group or distinct is not currently supported. Latest inside a latest is silly, and also not supported"
+     -> errNotAllowed "Latest inside a latest is silly, not supported"
     GroupBy _ x
      | allowWindowsOrGroups inv
-     -> goX x >> goNotAllowed
+     -> case cs of
+         (Latest _ _ : _) -> goX x >> go
+         _                -> goX x >> goNotAllowed
```
amosr: what about
``group key ~> filter pred ~> latest 5 ~> …``?
I think instead of this, might need a separate flag for "allowLatest" or something

tranma: Oh, forgot about this hack... "allowLatest" would work. I wonder if there's a nicer way to write this.

amosr: ideally we'll eventually allow group-in-group and latest-in-latest so that we can throw this away!

test/cli/repl/t10-avalanche/expected
```diff
@@ -40,16 +40,25 @@ c$conv$5 =
...
-    rlatest [Int] (3 : Int)
+conv$9 =
+    rfold   [Int] [Buf Int] (
+    \conv$10 : Buf Int.  
+    \conv$11 : Int. 
+    Latest_push# [Int] conv$10
+    conv$11)
+    ([ExceptScalarVariableNotAvailable
+     ,ExceptScalarVariableNotAvailable
+     ,ExceptScalarVariableNotAvailable] : Buf Int)
```
amosr: oh, I guess we need to mark latest primitives as "impure" so that simplifier cannot duplicate them or evaluate them

tranma: yeah, right

amosr: oh.. or can we just keep it as pure in Core, and only Avalanche needs to be impure? yes, sorry, never mind - we don't need to change Core simplifier

tranma: I didn't notice that this looks a bit strange (it shouldn't evaluate latest_make like this), but now I'm not sure if that's really a problem

amosr: I think it's OK, but I think VBuf should change to (VBuf Int [BaseValue]) or whatever, so this would actually be (Buf 3 [])

amosr:
I think this is good to merge & new stuff as separate :8ball:

tranma:
ok, shall do

PR #147, opened 2015-10-19
Introduce date literals

HuwCampbell:
The lexing could probably be improved a little, but it works fine.


amosr:
fair enough :+1:

tranma:
might be worth adding a CLI test to serve as example

tranma:
cool :tropical_drink:

PR #149, opened 2015-10-19
Topic/possibility

amosr:
Fix the broken tests

### DIFF
--- a/test/Icicle/Test/Source/Arbitrary.hs
+++ b/test/Icicle/Test/Source/Arbitrary.hs
@@ -213,7 +213,14 @@ instance Arbitrary QueryWithFeature where
   = do q   <- arbitrary
        now <- arbitrary
        o   <- arbitrary
-       t   <- arbitrary
+       -- Convert to Source type and back, because this filters out Bufs.
+       -- (It actually converts them to Arrays)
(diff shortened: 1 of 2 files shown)


test/Icicle/Test/Source/Arbitrary.hs
```diff
@@ -213,7 +213,14 @@ instance Arbitrary QueryWithFeature where
...
        o   <- arbitrary
-       t   <- arbitrary
+       -- Convert to Source type and back, because this filters out Bufs.
+       -- (It actually converts them to Arrays)
+       -- This is fine, because Buf can't be in feature types.
+       --
+       -- This was causing issues with conversion from Source, because
+       -- part of the resulting Core was using the ValType with Bufs, and
+       -- part was using the Source-converted type with Arrays.
+       Just t <- (valTypeOfType . typeOfValType) <$> arbitrary
```
tranma: ok, unfortunate but unavoidable I think

tranma:
:baby_chick: :poultry_leg:

test/Icicle/Test/Source/Arbitrary.hs
amosr: yeah - it's a bit weird but I think it's fine

PR #150, opened 2015-10-20
Reify dictionary entries

HuwCampbell:
Found this while bumping another project to use a later version. The error when running computations was: ` fold1 cannot be converted; desugar first`.


amosr:
looks good :+1:

HuwCampbell:
Hmm, breaks dictionary persistence.

amosr:
ah - damn.
can we remove the reify from here, leave the desugar and check, but actually return the original, un-inlined "checked"?
then, when actually running or whatever, inline, desugar and reify before converting?

amosr:
(this will give prettier info in :set too)

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -135,11 +138,11 @@ loadImports
...
  = hoistEither . mapLeft DictionaryErrorCheck
- $ foldlM go parentFuncs parsedImports
+ $ foldlM (go parentFuncs) [] parsedImports
  where
-  go env f
-   = do f' <- P.sourceDesugarF f >>= P.sourceCheckF
-        return $ M.union f' env
+  go env acc f
+   = do f' <- P.sourceCheckF (env <> acc) f
+        return $ acc <> f'
```
tranma: we don't desugar functions anymore?

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -150,10 +153,7 @@ checkDefs d defs
  where
   go (a, q)
    = do  (checked, _)  <- check' d q
-         let inlined    = P.sourceInline d checked
-         blanded       <- hoistEither . mapLeft DictionaryErrorTransform $ P.sourceDesugarQT inlined
-         (checked', _) <- check' d blanded
-         pure $ DictionaryEntry a (VirtualDefinition (Virtual checked'))
+         pure $ DictionaryEntry a (VirtualDefinition (Virtual checked))
```
tranma: why no more inlining? (sorry if I'm missing something big)

src/Icicle/Storage/Dictionary/Toml.hs
amosr: silly thing - I think we want to run desugar here too, because it checks pattern match exhaustiveness, but return the result of checkF

amosr: I could be wrong

HuwCampbell: Ok.
And Tran, yeah, we don't need to carry the desugared version along, and this way when we persist the dictionary it looks pretty much identical.

src/Icicle/Storage/Dictionary/Toml.hs
HuwCampbell: Moving it to a different stage in the pipeline.

amosr: it's silly, but the main reason is so we keep more or less the original, "nicer" version before we make it ugly

amosr:
:mag:

src/Icicle/Storage/Dictionary/Toml.hs
amosr: but it's also like - we might as well do the whole "compilation" in the one place. so it's either have this do the conversion to core and avalanche, or do it all later

tranma:
all good :meat_on_bone:

PR #151, opened 2015-10-20
Avalanche: fact loop is now over a list of name/type pairs

jystic:
This is in preparation for a flattening phase which operates on facts

### DIFF
--- a/src/Icicle/Avalanche/Eval.hs
+++ b/src/Icicle/Avalanche/Eval.hs
@@ -13,6 +13,7 @@ import              Icicle.Avalanche.Program
 import              Icicle.BubbleGum
 
 import              Icicle.Common.Base
+import              Icicle.Common.Type
 import              Icicle.Common.Value
 import qualified    Icicle.Common.Exp as XV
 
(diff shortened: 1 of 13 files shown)


jystic:
Hmm, this is broken on top of reify, fixing it up now

src/Icicle/Avalanche/Eval.hs
```diff
@@ -240,15 +244,18 @@ evalStmt evalPrim now xh values bubblegum ah stmt
 
     -- TODO: evaluation ignores history/bubblegum.
     -- All inputs are new, so history loop does nothing.
-    ForeachFacts _ _ _ FactLoopHistory _
+    ForeachFacts _ _ FactLoopHistory _
      -> return (ah, [])
 
     -- TODO: ignoring outputs inside loops.
-    ForeachFacts n n' _ FactLoopNew  stmts
+    ForeachFacts [(n, _), (n', _)] _ FactLoopNew  stmts
```
amosr: I wonder if we'll be able to split up the fact/value based on types later

src/Icicle/Avalanche/FromCore.hs
```diff
@@ -91,7 +91,8 @@ programFromCore namer p
    = factLoop FactLoopNew (C.reduces p)
 
   factLoop loopType reduces
-   = ForeachFacts (namerElemPrefix namer $ namerFact namer) (namerElemPrefix namer $ namerDate namer) (C.input p) loopType
+   = ForeachFacts [ (namerElemPrefix namer $ namerFact namer, C.input p)
+                  , (namerElemPrefix namer $ namerDate namer, DateTimeT) ] (C.input p) loopType
```
amosr: I think this ``C.input p : ValType`` should be ``T.PairT (C.input p) T.DateTimeT``, so that it covers the whole list of variables

jystic: Oh yeah, that makes more sense

amosr:
:zap: :partly_sunny:

PR #153, opened 2015-10-20
Latest: wrap Possiblies as Array (Sum Error t)

amosr:


amosr:
cool

src/Icicle/Source/ToCore/Fold.hs
```diff
@@ -236,81 +236,23 @@ convertFold q
...
+                     ( CE.pushBuf t'e
+                         CE.@~ CE.xVar n'acc
+                         CE.@~ res )
+           let zero  = CE.emptyBuf t'e
+                         CE.@~ CE.constI i
+
+           let x'    = CE.xLam n'buf t'buf
+                     ( CE.readBuf t'e CE.@~ CE.xVar n'buf )
+
+           return $ ConvertFoldResult kons zero x' t'buf t'arr
```
tranma: cool, this was pretty a bit much

HuwCampbell:
Will this fix #157 ?

src/Icicle/Source/ToCore/Fold.hs
amosr: this is true

amosr: yeah, it's also a shame that this stuff can't all be done in source to source right now

amosr:
not sure..

amosr:
I think I'll merge this

PR #154, opened 2015-10-20
SumT can now be melted, even inside ForeachFacts

jystic:


src/Icicle/Avalanche/Statement/Simp/Melt.hs
```diff
@@ -21,35 +23,88 @@ import              P
...
+
+  , primMkOpt  :: ValType -> Name n -> Name n -> Exp a n p
+  , primIsSome :: ValType -> Exp a n p        -> Exp a n p
+  , primGet    :: ValType -> Exp a n p        -> Exp a n p
+
+  , primMkSum   :: ValType -> ValType -> Name n -> Name n -> Name n -> Exp a n p
+  , primIsRight :: ValType -> ValType -> Exp a n p                  -> Exp a n p
+  , primLeft    :: ValType -> ValType -> Exp a n p                  -> Exp a n p
+  , primRight   :: ValType -> ValType -> Exp a n p                  -> Exp a n p
+  }
```
jystic: I needed to use these in two functions, so I pulled it out to a record (designed to be unpacked with `RecordWildCards`, hope this is ok!)

Might be useful to be its own module as things are written longhand / duplicated in the constructors pass at the moment.

tranma:
nice, was hoping to see this

src/Icicle/Avalanche/Statement/Simp/Melt.hs
tranma: in Source/Core, there are `Combinators` modules where you can stick things like these. Might be worth adding one here

tranma:
not sure if flatten_simp_commutes_value should still hold

jystic:
^ yeah i'm fixing this at the moment, it's a good test so i don't want to break it

src/Icicle/Avalanche/Statement/Simp/Melt.hs
jystic: I'm thinking I might move them to `Icicle.Avalanche.Combinators` in a separate PR?

jystic:
Ok @tranma, `flatten_simp_commutes_value` should hold now

tranma:
yup, fair enough :bomb:

PR #158, opened 2015-10-21
Move more functionality from main and Repl into pipeline

HuwCampbell:
Still might try and generalise the types of these. Keeping away from Text and Var for instance.

### DIFF
--- a/src/Icicle/Pipeline.hs
+++ b/src/Icicle/Pipeline.hs
@@ -130,8 +130,8 @@ type QueryTop'T = SQ.QueryTop (AnnotT Parsec.SourcePos) Var
 
 type Funs a b = [((a, Name b), SQ.Function a b)]
 type FunEnvT a b = [ ( Name b
-                 , ( ST.FunctionType b
-                   , SQ.Function (AnnotT a) b ) ) ]
+                   , ( ST.FunctionType b
+                     , SQ.Function (AnnotT a) b )) ]
 
 
 sourceParseQT :: Text -> Text -> Either (CompileError Parsec.SourcePos Var ()) QueryTop'
--- a/src/Icicle/Pipeline.hs
+++ b/src/Icicle/Pipeline.hs
@@ -2,7 +2,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Icicle.Pipeline
   ( CompileError(..)
-  , QueryTop', QueryTop'T, Program', ProgramT
+  , QueryTop', QueryTop'T, Program'
   , annotOfError
   , sourceParseQT
   , sourceParseF
@@ -123,7 +123,6 @@ type Var        = SP.Variable
 type AnnotT a   = ST.Annot a Var
 
 type Program'   = Core.Program () Var
-type ProgramT   = Core.Program () Text
 
 type QueryTop'  = SQ.QueryTop Parsec.SourcePos Var
 type QueryTop'T = SQ.QueryTop (AnnotT Parsec.SourcePos) Var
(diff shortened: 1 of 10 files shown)


src/Icicle/Repl.hs
```diff
@@ -100,8 +109,8 @@ sourceParse = mapLeft ReplErrorCompile . P.sourceParseQT "repl"
 sourceDesugar :: P.QueryTop' -> Either ReplError P.QueryTop'
 sourceDesugar = mapLeft ReplErrorCompile . P.sourceDesugarQT
 
-sourceReify :: P.QueryTop'T -> Either ReplError P.QueryTop'T
-sourceReify = return . P.sourceReifyQT
+sourceReify :: P.QueryTop'T -> P.QueryTop'T
+sourceReify = P.sourceReifyQT
```
tranma: might as just just export P.sourceReifyQT and remove this?

HuwCampbell: I did actually have that, but I liked that none of the reexports for the repl said QT at the end, so left it like that.

HuwCampbell:
It's very annoying that we swap between `Var` and `Text`.

src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -38,10 +38,10 @@ import           P
 data DictionaryImportError
   = DictionaryErrorIO         E.SomeException
   | DictionaryErrorParsecTOML Parsec.ParseError
-  | DictionaryErrorParsecFunc (P.CompileError Parsec.SourcePos)
+  | DictionaryErrorParsecFunc (P.CompileError Parsec.SourcePos SP.Variable ())
   | DictionaryErrorParse      [DictionaryValidationError]
-  | DictionaryErrorCheck      (P.CompileError Parsec.SourcePos)
-  | DictionaryErrorTransform  (P.CompileError Parsec.SourcePos)
+  | DictionaryErrorCheck      (P.CompileError Parsec.SourcePos SP.Variable ())
+  | DictionaryErrorTransform  (P.CompileError Parsec.SourcePos SP.Variable ())
```
tranma: can these be generalised too...?

HuwCampbell: Don't think so.

HuwCampbell: At least not meaningfully.

amosr:
remind me - what transforms need to be Text? maybe we can just modify them to work on Var.

HuwCampbell:
Icicle.Avalanche.FromCore.namerText, and inside Pipeline there's a bunch of T.pack

amosr:
We might be better off changing those to work on Var, than swapping between the two for no real reason

src/Icicle/Pipeline.hs
```diff
@@ -81,7 +93,7 @@ annotOfError e
     CompileErrorProgram _
      -> Nothing
 
-instance Pretty a => Pretty (CompileError a) where
+instance (IsString b, Ord b, Pretty a, Pretty b, Show a, Show b, Show c) => Pretty (CompileError a b c) where
```
amosr: the ``IsString`` and ``Ord`` seem strange to me. are they really necessary? I can't think offhand why pretty printing an error would require making a variable, so if it is necessary, add a comment please

HuwCampbell: It contains CheckError, which has suggestions, which calls prettyFunFromStrings.

amosr: ah yeah, forgot about that

test/cli/repl/t10-avalanche/expected
```diff
@@ -6,91 +6,91 @@ ok, loaded test/cli/repl/data.psv, 13 rows
...
-  for_facts (gen$fact : ((Sum Error Int), DateTime)) in new {
-    read [Mutable] [(Sum Error (Buf Int))] acc$conv$22 = acc$conv$22;
-    write acc$conv$22 = Sum_fold# [Error] [Buf Int] [(Sum Error (Buf Int))]
+  init [Mutable] [Int] [32macc[0m$[32mc[0m$[32mconv[0m$[32m11[0m = 0 : Int;
+  init [Mutable] [(Sum Error (Buf Int))] [32macc[0m$[32mconv[0m$[32m22[0m = Right Buf 3 [] : (Sum Error (Buf Int));
+  load_resumable [Int] [32macc[0m$[32mc[0m$[32mconv[0m$[32m11[0m;
+  load_resumable [(Sum Error (Buf Int))] [32macc[0m$[32mconv[0m$[32m22[0m;
+  for_facts ([32mgen[0m$[32mfact[0m : ((Sum Error Int), DateTime)) in new {
+    read [Mutable] [(Sum Error (Buf Int))] [32macc[0m$[32mconv[0m$[32m22[0m = [32macc[0m$[32mconv[0m$[32m22[0m;
+    write [32macc[0m$[32mconv[0m$[32m22[0m = Sum_fold# [Error] [Buf Int] [(Sum Error (Buf Int))]
```
jystic: Is there any way we can avoid this escape code stuff in the test output? It makes the diffs pretty hard to read

HuwCampbell: Done

HuwCampbell:
I think I'm done here

tranma:
looks good to me, thanks for this. :banana:

PR #163, opened 2015-10-23
very slightly better core gen

tranma:
this doesn't change much but most of the time the generated Core returns something interesting rather than just a constant. e.g. try `cabal repl test`:

on master you'll probably get `Returning: fozzy = False : Bool`

here you might get a primitive application or something cooler.

NB: could probably do more to improve tryExp....

### DIFF
--- a/test/Icicle/Test/Core/Arbitrary.hs
+++ b/test/Icicle/Test/Core/Arbitrary.hs
@@ -227,7 +227,29 @@ tryExpForType ty env
 --
 tryExp :: Env Var Type -> Gen (Exp () Var Prim, Type)
 tryExp env
- = context env
+ = oneof_sized
+    -- Something from the context, or a primitive application.
+    [ context env ]
+    -- A function?
+    [ funs env ]
+ where
+  funs e
+   = do ts    <- listOf $ firstOrderType
+        r     <- arbitrary
+        let t  = FunT ts r
+        x     <- tryExpForType t e
+        return (x, t)
+
+firstOrderType :: Gen Type
+firstOrderType
+ = oneof [ val, args ]
+ where
+  val
+   = FunT [] <$> arbitrary
+  args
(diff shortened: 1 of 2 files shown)


test/Icicle/Test/Core/Arbitrary.hs
```diff
@@ -394,12 +441,12 @@ programForStreamType streamType
...
 
-  -- A reduction is either a fold or a latest
+  -- A reduction is a fold
   gen_reduce sE pE
    = do (i,t) <- oneof $ fmap return $ Map.toList sE
         at <- arbitrary
-        kx <- gen_exp (FunT [FunT [] at, FunT [] t] at) pE
-        zx <- gen_exp (FunT [] at) pE
+        kx <- gen_exp_for_type (FunT [FunT [] at, FunT [] t] at) pE
+        zx <- gen_exp_for_type (FunT [] at) pE
```
amosr: just a suggestion: I think here, instead of generating the type at, you might be able to generate (zx, at) from tryExp?

amosr: same for map transform

tranma: yup, sounds fine

amosr:
this looks good :penguin:
I am sure this is a vast improvement. it is up to you whether you want to continue improving this or work on something else

tranma:
@amosr bike-shedding it a little bit more

tranma:
what do you think, @jystic?

amosr:
have you run this with bumping the number of iterations up to ten thousand or whatever?

jystic:
As long as it uses things from the context more than 0% of the time, then I'm happy :100:

tranma:
@amosr yeh, I did. Doubling maxDiscardRatio is #JustQuickCheckThings

amosr:
sweet

PR #164, opened 2015-10-23
Add seq function

HuwCampbell:


src/Icicle/Source/ToCore/Prim.hs
```diff
@@ -197,6 +197,15 @@ convertPrim p ann resT xts
   gofun DaysEpoch
    = return $ primmin $ Min.PrimDateTime Min.PrimDateTimeDaysEpoch
 
+  gofun Seq
```
amosr: you should be able to convert this without adding a new primitive to Core. in fact having a Core primitive for it doesn't really make sense.
should be able to do something like:

| [ (xa, ta) , (xb, tb) ] <- xts
, getPossibilityOrDefinitely a == PossibilityPossibly
…
= generate
   Case xa
    | Left err -> Left err
    | Right _ ->
        (if tb is Possibly
         then Right xb
         else xb) 
    end

amosr: the get primitive shouldn't be able to be converted to straight Core, too

amosr: should. SHOULD.

amosr: actually!
you can actually just convert
``seq a b`` to ``b``.
this is because the reify pass will have already added the case statement to unwrap "a" if it is a Possibly. and if b is a possibly, it will just unwrap it and rewrap it.
yes - conversion for seq is actually trivial.

amosr: trivial, but not obvious. make sure you add a comment saying that ReifyPossibility has already taken care of the sequencing possiblies, or something

HuwCampbell:
Cool, that works.

amosr:
looks good. can you or have you added Seq to the arbitrary? :page_facing_up:

PR #165, opened 2015-10-24
Topic/type annotations

HuwCampbell:


src/Icicle/Source/PrettyAnnot.hs
```diff
@@ -1,74 +1,133 @@
...
 import                  P
 
-import                  Data.List (intersperse)
-
 newtype PrettyAnnot q
  = PrettyAnnot
  { getPrettyAnnot :: q }
 
-instance (Pretty a, Pretty n) => Pretty (PrettyAnnot (Exp a n)) where
+instance (Pretty a, Pretty n) => Pretty (PrettyAnnot (Exp (Annot a n) n)) where
```
amosr: oh cool, this is good.
do you suppose there's any way to add these annotations to the ``Pretty`` instance in ``Icicle.Source.Query.Query``, and selectively filter them out or not print them afterwards?
I think adding ``Pretty a`` to the instance there is reasonable.

amosr: it doesn't have to be exactly as-is now either; the annotated output is just for us

amosr:
good stuff :grimacing:

src/Icicle/Source/PrettyAnnot.hs
HuwCampbell: They could only really be added if we only had pretty on the annotated queries (which might not actually be that bad) or we had overlapping instances.
It really easy to not print them, just use display instead of displaywithannotation.

amosr: I was thinking if we just modified the pretty instance for Annot, we wouldn't need overlap?
```
instance {-# don't need a #-} Pretty n => Pretty (Annot a n) where
 pretty a = … pretty (annResult a)

instance Pretty a => Pretty (Exp a n) where
 pretty (Var a n) = annotate a (pretty n)
```

tranma:
this makes Avalanche a lot more readable imo. Thanks for this :+1:

HuwCampbell:
We can revisit Amos's suggestions when he gets back, but for now I think we should merge.

PR #167, opened 2015-10-26
C evaluator now supports melting inputs / unmelting outputs

jystic:
This updates the C evaluator so that it can deal with melted inputs from reified possibilities and other more complex data types.


tranma:
I had a very quick look, seems ok :hammer: waiting for :green_apple:

PR #168, opened 2015-10-26
Only generate non-negative buffer sizes

jystic:
:)


tranma:
oops, my bad :oncoming_police_car:

PR #170, opened 2015-10-29
WIP - melt more stuff

tranma:
incomplete, status review only...

This will eventually melt types inside Array/Buf/Map while leaving Array/Buf/Map abstract themselves. So `ArrayT (PairT IntT IntT)` should become `PairT (ArrayT IntT) (Array IntT)`, to be further melted into separate values. Sums are more complicated, `ArrayT (SumT ErrorT IntT)` can become `PairT (ArrayT BoolT) (PairT (ArrayT ErrorT) (ArrayT IntT))`


tranma:
NB so this is still unsound (i.e. flatten_preserve_value fails sometimes), gonna fix

src/Icicle/Avalanche/Prim/Eval.hs
```diff
@@ -169,5 +201,35 @@ evalPrim p vs
...
+   = return (f     : fs, s     : ss)
+  uz (fs, ss) VUnit -- unitialised array
+   = return (VUnit : fs, VUnit : ss)
+  uz _ _
+   = primError
+
+  us (bs, ls, rs) (VLeft l)
+   = return ( VBool  False  : bs
+            , l             : ls
+            , VUnit         : rs )
```
amosr: should probably use defaultOfType instead of VUnit. I don't think we have partial evaluation on Avalanche yet, but if we did, VUnit would change the type

src/Icicle/Avalanche/Prim/Flat.hs
```diff
@@ -83,12 +83,15 @@ data PrimUnsafe
 
 
 data PrimUpdate
- -- | Should this be unsafe too? It's really both.
- = PrimUpdateArrayPut        ValType
+ = PrimUpdateArrayPut  ValType
+ | PrimUpdateArrayPut2 ValType ValType -- ^ update 2 arrays, should become C statement exps
```
amosr: this shouldn't be necessary, but if it's easier we can do it for now

amosr: actually never mind, this is fine

src/Icicle/Avalanche/Prim/Eval.hs
tranma: yeh, you're right

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -458,3 +465,59 @@ killAccumulator acc xx statements
 
    | otherwise
    = return ((), s)
+
+-- | Simplify expressions with flattened primitives. Performs:
+--    * constant folding for common expressions and flat prims.
+--    * beta reduction
+--
+simpStatementExps :: Ord n => a -> Statement a n F.Prim -> Statement a n F.Prim
+simpStatementExps a_fresh statements
```
amosr: might be able to use transformX here instead of transformUDStmt, to get rid of boilerplate. it doesn't really matter though

src/Icicle/Sea/FromAvalanche/Prim.hs
```diff
@@ -142,6 +142,8 @@ seaOfPrimUpdate p
  = case p of
      PrimUpdateArrayPut t
       -> prefixOfValType (ArrayT t) <> "put"
+     PrimUpdateArrayPut2 t1 t2
+      -> prefixOfValType (PairT (ArrayT t1) (ArrayT t2)) <> "put2"
```
jystic: Would you expect this prim to get all the way to C?

tranma: actually, no. This should become a C statement expression with "put", I was just chucking this here for now.

jystic: excellent, sounds good to me :)

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -96,28 +127,51 @@ constructor a_fresh statements
...
+   | Just (PrimUpdate (PrimUpdateArrayPut _), [arr,ix,v]) <- takePrimApps x
+   , Just arrx              <- resolve env arr
+   , Just (ta, tb, i, a, b) <- fromSummedArray arrx
+   , Just vx                <- resolve env v
+   , Just (_,_,boool,l,r)   <- fromSum vx
+   , tai                    <- ArrayT BoolT
+   , taa                    <- ArrayT ta
+   , tab                    <- ArrayT tb
+   = case boool of
+      XValue _ _ (VBool False)
```
amosr: umm. this is weird, you're doing a case on the runtime tag
maybe it just needs to be something like
```
( put i ix boool , (put a ix l , put b ix r ) )
```

tranma: oh right, that was stupid, what was I thinking =( eh, maybe this simplification isn't necessary, but then it won't get rid of unsum/unzip

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -96,28 +127,51 @@ constructor a_fresh statements
...
+      XValue _ _ (VBool True)
+       -> xPrim (PrimMinimal (Min.PrimPair (Min.PrimPairSnd tai tab)))
+          `xApp` (xPrim (PrimUpdate (PrimUpdateArrayPut2 (ArrayT BoolT) (ArrayT tb)))
+                  `xApp` i `xApp` b `xApp` ix `xApp` xValue BoolT (VBool True) `xApp` r)
+      _ -> x
+
+   | Just (PrimUpdate (PrimUpdateArrayPut _), [arr,ix,v]) <- takePrimApps x
+   , Just arrx           <- resolve env arr
+   , Just (ta, tb, a, b) <- fromZippedArray arrx
+   , Just vx             <- resolve env v
```
amosr: instead of looking in the environment for v, I think you just want to put ``fst v`` and ``snd v``. if ``resolve`` or ``fromPair`` we still want to do this

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -96,28 +127,51 @@ constructor a_fresh statements
...
    = b
 
-   | Just (PrimUnsafe (PrimUnsafeArrayIndex _), [XVar _ n, ix]) <- takePrimApps x
-   , Just x' <- get env n
-   , Just (ta, tb, a, b) <- fromZippedArray x'
+   -- update
+   | Just (PrimUpdate (PrimUpdateArrayPut _), [arr,ix,v]) <- takePrimApps x
+   , Just arrx              <- resolve env arr
+   , Just (ta, tb, i, a, b) <- fromSummedArray arrx
+   , Just vx                <- resolve env v
```
amosr: here too - should use ``isRight v``, ``unsafeLeft`` and ``unsafeRight`` instead of ``resolve`` and ``fromSum``

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
amosr: make sure you run constructor over ``fst v`` too, just in case

src/Icicle/Avalanche/Statement/Simp/Melt.hs
```diff
@@ -529,6 +719,10 @@ meltValue v t
       | otherwise
       -> Nothing
 
+freshes :: Int -> Name n -> Fresh n [Name n]
+freshes i n
+ = replicateM i $ freshPrefix' n
+
 ------------------------------------------------------------------------
 
 unmeltValue :: [BaseValue] -> ValType -> Maybe BaseValue
```
amosr: this whole Melt transform looks right to me

amosr:
looking good. getting there. I think it's just the Constructor transform that needs attention

tranma:
this works for arrays, except when the array is the result of a `Buf_read`, because I haven't done Bufs yet.

amosr:
great :abcd: :cd:

tranma:
gonna merge this then do bufs on another branch. /cc @amosr

PR #171, opened 2015-10-30
Support for melting structs

jystic:
Sorry for the massive PR, I had to fix a lot of things to get the tests working. There was a bunch of lurking issues with melting that supporting structs exposed.

### DIFF
--- a/src/Icicle/Avalanche/Eval.hs
+++ b/src/Icicle/Avalanche/Eval.hs
@@ -261,12 +261,12 @@ evalStmt evalPrim now xh values bubblegum ah stmt
     -- (i.e. where ty == ty' and we only have a singleton list of bindings)
     ForeachFacts [(n, ty)] ty' FactLoopNew stmts
      | ty == ty'
-     -> do  let evalInput ah' input = do
-                  let v0     = snd (fact input)
-                      v1     = VDateTime (time input)
+     -> do  let evalInput ah' inp = do
+                  let v0     = snd (fact inp)
+                      v1     = VDateTime (time inp)
                       vv     = VPair v0 v1
                       input' = Map.insert n (VBase vv) xh
-                      bgf    = Just $ fst $ fact input
+                      bgf    = Just $ fst $ fact inp
 
                   fst <$> evalStmt evalPrim now input' [] bgf ah' stmts
 
@@ -274,9 +274,9 @@ evalStmt evalPrim now xh values bubblegum ah stmt
             return (ahs, [])
 
     ForeachFacts ns ty FactLoopNew stmts
-     -> do  let evalInput ah' input = do
-                  let v0  = snd (fact input)
-                      v1  = VDateTime (time input)
+     -> do  let evalInput ah' inp = do
+                  let v0  = snd (fact inp)
+                      v1  = VDateTime (time inp)
                       vv  = VPair v0 v1
                       mvs = meltValue vv ty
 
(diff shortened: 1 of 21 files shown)


src/Icicle/Avalanche/Eval.hs
```diff
@@ -338,9 +338,9 @@ evalStmt evalPrim now xh values bubblegum ah stmt
     Output n t xts
      -> do  vs  <- traverse ((baseValue =<<) . eval . fst) xts
             case (vs, unmeltValue vs t) of
-              (v:[], _)    -> return (ah, [(n, v)])
-              (_, Just v)  -> return (ah, [(n, v)])
-              (_, Nothing) -> Left (RuntimeErrorOutputTypeMismatch n t vs)
+              (_,    Just v)  -> return (ah, [(n, v)])
+              (v:[], Nothing) -> return (ah, [(n, v)])
```
amosr: would like a short comment.. why would this happen? "the output doesn't match the type, but there's only one output so we might as well return it"

src/Icicle/Avalanche/Statement/Simp/Melt.hs
```diff
@@ -185,6 +201,12 @@ meltAccumulators a_fresh statements
                     . Read nb' nb avt tb
                     $ ss'
 
+           | Just (Mutable, StructT ts, nas)            <- Map.lookup acc env'
+           , fts                                        <- Map.elems (getStructType ts)
+           -> do ns' <- replicateM (length fts) (freshPrefix' n)
+                 ss' <- substXinS a_fresh n (primMkStruct ts ns') ss
+                 go $ foldr (\(n',na,t) -> Read n' na avt t) ss' (List.zip3 ns' nas fts)
```
amosr: this falls out pretty nicely

amosr:
I didn't see where Outputs are split up?
this is good. I am particularly happy that melt and constructor are able to stay separate.
:fries:

src/Icicle/Avalanche/Eval.hs
jystic: This is probably the most horrible part of the PR, basically this is capable of evaluating Avalanche both before and after melting. If it's been melted then unmeltValue will return Just BaseValue otherwise it will return Nothing. Nothing could mean lots of things I guess, but if we only have a single value in the list, it probably means that it was a value that didn't need unmelting because it Avalanche had not been melted yet.

src/Icicle/Avalanche/Statement/Simp/Melt.hs
jystic: Yeah I was surprised! I thought this would be much harder, but it's almost the nicest of all.

src/Icicle/Avalanche/Statement/Simp/Melt.hs
```diff
@@ -402,13 +452,18 @@ meltExp a_fresh x t
...
      OptionT tx
-      -> meltExp a_fresh (primIsSome tx x) BoolT
-      <> meltExp a_fresh (primGet    tx x) tx
+      -> meltExp a_fresh (primIsSome  tx x) BoolT
+      <> meltExp a_fresh (primGetSome tx x) tx
+
+     StructT ts
+      | fts <- Map.toList (getStructType ts)
+      -> let go (nf,tf) = meltExp a_fresh (primGetField ts nf tf x) tf
+         in concat (fmap go fts)
```
jystic: @amosr this is the function used by meltOutputs

PR #174, opened 2015-11-02
Topic/v1 notes

amosr:
Notes about what needs to be done and things and so on


tranma:
also document what type becomes custom C types? these should be `iarray_t, ibuf_t, imap_t` afaict, these ADTs can go in `data/sea/` too

PR #177, opened 2015-11-02
Parser: use "fold" instead of "let fold".

amosr:


jystic:
Looks good to me :8ball:

PR #178, opened 2015-11-03
Adds a memory allocator

jystic:
Adds a memory allocator for use by arrays, buffers, maps and probably strings.


amosr:
Looks good :chocolate_bar:

PR #187, opened 2015-11-04
Implementation of date primitives

jystic:
This finishes off the ivory-dates branch


amosr:
looks good.
I think might be the point where we can split t30-sea up into a few different cases, and I don't think all the C tests need to emit the C, maybe just C-eval

amosr:
:man:

PR #189, opened 2015-11-04
Stupid implementation of Maps in flattening

amosr:
It's awful but it seems to work


jystic:
This is great :100: love it!

PR #192, opened 2015-11-05
Topic/v1 notes

amosr:


tranma:
can haz merge this https://github.com/ambiata/icicle/pull/194 into this branch?

amosr:
please do :+1:

amosr:
can I merge this?

tranma:
:spaghetti:

src/Icicle/Sea/FromAvalanche/Program.hs
```diff
@@ -245,8 +245,39 @@ seaOfXValue v
...
+     VArray vs
+      | ArrayT t' <- t
+      -> let len = length vs
+             writes arr (v',i)
+              = prim (PrimUpdate $ PrimUpdateArrayPut t')
+                     [arr, int i, seaOfXValue v' t']
+             init
+              = prim (PrimUnsafe $ PrimUnsafeArrayCreate t')
+                     [int len]
+        in  foldl writes init (vs `List.zip` [0..])
```
amosr: have a look at this if you don't mind. it should end up as something like
```
 [ "thing" ]
 iarray_put( iarray_create(s->mempool, 1), 0, "thing" )
```

tranma: I haven't seen this in action in the REPL yet, what's an example?

amosr: actually I think it might only be required for maps, after we split maps into two arrays

tranma: also when you read a buffer using `buf_read` (I'll do buffers and we'll see)

amosr: I don't think it will be required for that actually - this is only for compiletime constants

PR #195, opened 2015-11-06
remove Latest

tranma:
Now that RLatest is gone from Core, AccumulatorType and Push statements are basically dead code in Avalanche, so let's just remove them.

I'm not too sure about the bubblegum semantics of circular buffers as accumulators here, presumably it needs to fill the role of what Latest accumulator was doing...

/cc @amosr


amosr:
yeah, the bubblegum for Latest should change but this isn't a big deal
quick eyeball but :+1:

tranma:
mergerizing :boom:

PR #197, opened 2015-11-09
Throwing it all away: remove array melting and map implementation

amosr:
Before implementing C++ templates for arrays and buffers etc


tranma:
looks good so far

amosr:
@jystic saved us from the peril of C++

PR #199, opened 2015-11-10
test seaworthy

tranma:
check all Core programs can be converted to C(++)

@amosr should eventually pass


amosr:
looks good :+1:

tranma:
gonna merge to try out on the c++ branch

markhibberd:
@tranma @amosr Not sure if it is 100% appropriate for this situation but this can be useful - https://github.com/ambiata/disorder.hs/blob/master/disorder-core/src/Disorder/Core/Property.hs#L61-L63 - basically can do xor True, i.e. `.^.`. So that the test/build passes until the test is made to work and then it will fail once the test passes so you can have some sort of party and un-xor'ing ceremony. That way you can include it in the test-suite without breaking the build, and you can dev against it on branches by trying to make the test "fail" instead of pass. Again, not always the right thing, but can sometimes be useful for some of these types of things.

PR #204, opened 2015-11-11
Topic/eq buf

amosr:
buffer equality


jystic:
:cake:

jystic:
greedy for :cake:

PR #206, opened 2015-11-11
Topic/eq buf

amosr:


tranma:
looks fine, hope this makes coresimp better :chart:

PR #207, opened 2015-11-12
make simps a little less slow

tranma:


src/Icicle/Core/Exp/Simp.hs
```diff
@@ -29,40 +29,39 @@ import qualified Data.Set                       as Set
...
 simpX a_fresh isValue = go . beta
   where
     beta  = B.beta isValue
     go xx = case xx of
       -- * constant folding for some primitives
       XApp a p q
-        | p' <- go p
-        , q' <- go q
+        | Just p' <- go p
+        , Just q' <- go q
```
amosr: won't this only happen if both sub-expressions succeed? but they won't succeed if they're base cases?

src/Icicle/Core/Exp/Simp.hs
```diff
@@ -29,40 +29,39 @@ import qualified Data.Set                       as Set
...
-        | p' <- go p
-        , q' <- go q
+        | p' <- fromMaybe p $ go p
+        , q' <- fromMaybe q $ go q
         , Just (prim, as) <- takePrimApps (XApp a p' q')
-        , Just args       <- mapM (takeValue . go) as
-        -> fromMaybe (XApp a p' q') (simpP a_fresh prim args)
+        , Just args       <- join . fmap sequence
+                           $ mapM (fmap takeValue . go) as
+        -> return $ fromMaybe (XApp a p' q') (simpP a_fresh prim args)
```
amosr: will this terminate if simpP returns Nothing? I'm not sure about this fromMaybe?

tranma: I guess, if simpP returns Nothing, but either "go p" or "go q" simp'ed, we should return a new XApp

amosr: yeah .. but if go p and go q are both Nothing, and simpP returns Nothing, I think this will still return ``Just (XApp p q)``

tranma:
coresimp is almost instant but the simplified core is less good, probably needs a bit of fiddles with beta

```
-                                                  ((Left ExceptTombstone, (Right 0, ()))@{((Sum Error Int), ((Sum Error Int), Unit))})
+                                                  (pair#@{(Sum Error Int) ((Sum Error Int), Unit)} (Left ExceptTombstone@{(Sum Error Int)})
+                                                   (
+                                                   let s$conv$22 = right#@{Error Int} (0@{Int})
+                                                    in pair#@{(Sum Error Int) Unit} s$conv$22 (()@{Unit})))
```

amosr:
that's odd...

src/Icicle/Core/Exp/Simp.hs
```diff
@@ -29,40 +29,50 @@ import qualified Data.Set                       as Set
...
       XLet a n x1 x2
-        -- | not $ n `Set.member` freevars (go x2)
-        -- -> go x2
-        -- | otherwise
-        -> XLet a n (go x1) (go x2)
-
-      b@(XVar{})   -> b
-      b@(XPrim{})  -> b
-      b@(XValue{}) -> b
+        -> XLet a n (just go x1) <$> go x2
```
amosr: what if you have ``let x = (5  + 2) in blah``,
the ``go (5 + 2)`` will return ``Just 7`` but ``go blah`` will return Nothing, so this will return Nothing?

amosr:
:+1:

PR #209, opened 2015-11-12
ToCore.Fold: Remove superfluous units on lets

amosr:


amosr:
less clutter means less meaningless stuff for Melt to deal with

jystic:
:cherries:

PR #210, opened 2015-11-12
improve simps

tranma:
only need to crunch twice (crunching once yields avalanche type error from leftover melt stuff)

sadly the performance doesn't improve much for `feature injury ~> group 1 ~> correlation 0 severity`

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
+++ b/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
@@ -80,7 +80,7 @@ constructor a_fresh statements
 
   goS env s
    = let env' = updateExpEnv s env
-         go   = goX env
+         go   = goX env'
          ret s' = return (updateExpEnv s' env', s')
      in  case s of
           If x t e
--- a/src/Icicle/Avalanche/Simp.hs
+++ b/src/Icicle/Avalanche/Simp.hs
@@ -45,22 +45,22 @@ simpFlattened a_fresh p
       let Program i bd s = p'
 
       s' <-  melt a_fresh s
-         >>= forwardStmts a_fresh . pullLets
-         >>= thresher a_fresh
-         >>= forwardStmts a_fresh
-         >>= crunch i bd
-         >>= crunch i bd
+         >>= crunchy i bd . pullLets
          >>= return . simpStatementExps a_fresh
 
       return $ p { statements = s' }
-
  where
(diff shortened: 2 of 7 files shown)


amosr:
ah! fantastico :100:

tranma:
d'oh, quickcheck found cases where you can't just crunch twice. I made it crunch to a fixpoint, but all of the REPL tests only need one round of crunch, it seems.

Also, we were calling simpFlattened twice in the REPL before (this was the required magic sauce to make the simplified flattened avalanche type check then), but this is no longer necessary.

tranma:
`feature injury ~> group 1 ~> correlation 0 severity` takes ~66s rather than ~70. Small steps.

amosr:
whoa!!!
this coupled with some changes I made earlier brought it down to 38s for me

:crystal_ball:

amosr:
our two optimisations combined must have had a multiplicative effect because mine alone only removed 5s, and yours alone only 5s, but together they take off 30s

tranma:
....teamwork? team team team

src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -80,7 +80,7 @@ constructor a_fresh statements
 
   goS env s
    = let env' = updateExpEnv s env
-         go   = goX env
+         go   = goX env'
```
tranma: @amosr I think there was a missing prime here. It'll make constructor less effective (i.e. needs to run more constructor passes that otherwise needed)

amosr: not 100% sure about this, need to make sure the updated environment from a non-recursive Let doesn't affect the let definition

amosr: yeah, this shouldn't have the prime. if you have something like
```
Let x = 5
Let x = (x + 2)
```
here we need to be very careful that the x mentioned in "x + 2" is the "x=5", not the "x=x+2".
(I'm not certain how adding the prime would make it more effective anyway)

tranma: resolved IRL, prime ok

tranma:
should I merge this? /cc @amosr @jystic

amosr:
yep - sorry for the confusion :+1:

PR #211, opened 2015-11-12
Very early version of PSV reading/writing

jystic:
Still a bunch to do before this is production ready, it doesn't handle any kind of escaping yet (C or JSON) and leaks memory (malloc instead of mempool).


amosr:
:camel:

PR #212, opened 2015-11-12
The complicated expression environment is actually unnecessary since …

amosr:
…shadowing is not allowed in Avalanche


tranma:
cool :ice_cream: merge this and I'll make ExpEnv store the freevars as well?

jystic:
:palm_tree:

PR #220, opened 2015-11-16
consolidate compiler/REPL pipeline and tests 

tranma:
make tests use Pipeline so we don't have to jump through tons of hoops just to write a test on Avalanche/Flatten/Sea


amosr:
groovy :+1:

jystic:
Cool :banana:

tranma:
gonna merge this, then write more specific tests for melt/constructor (e.g. no more pack/unpack)

PR #221, opened 2015-11-16
The beginnings of a PSV test

jystic:
This also fixes a bunch of bugs caught by the test.

This borrowed a bunch of stuff from the seaworthy test so it needs to be updated to use the new pipeline.

The next step is actually feeding it the generated input data as a PSV file.

### DIFF
--- a/data/sea/20-simple.h
+++ b/data/sea/20-simple.h
@@ -57,8 +57,9 @@ static ibool_t   INLINE istring_ne    (istring_t x, istring_t y) { return strcmp
 
 static istring_t INLINE istring_copy(imempool_t *into, istring_t val)
 {
-    char* alloc = imempool_alloc(into, strlen(val));
-    strcpy(alloc, val);
+    size_t val0_size = strlen (val) + 1;
+    char  *alloc     = imempool_alloc (into, val0_size);
+    memcpy (alloc, val, val0_size);
     return alloc;
 }
 
--- a/src/Icicle/Sea/FromAvalanche/Analysis.hs
+++ b/src/Icicle/Sea/FromAvalanche/Analysis.hs
@@ -8,6 +8,7 @@ module Icicle.Sea.FromAvalanche.Analysis (
   , accumsOfProgram
   , outputsOfProgram
   , readsOfProgram
+  , typesOfProgram
   ) where
 
 import           Icicle.Avalanche.Prim.Flat
@@ -16,12 +17,16 @@ import           Icicle.Avalanche.Statement.Statement
 
 import           Icicle.Common.Annot
 import           Icicle.Common.Base
+import           Icicle.Common.Exp
 import           Icicle.Common.Type
 
 import           P
 
 import           Data.Map (Map)
 import qualified Data.Map as Map
+
+import           Data.Set (Set)
+import qualified Data.Set as Set
 
 
 ------------------------------------------------------------------------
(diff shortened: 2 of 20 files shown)


src/Icicle/Sea/FromAvalanche/Analysis.hs
```diff
@@ -168,3 +173,61 @@ outputsOfStatement stmt
...
+     Write _ x             -> typesOfExp x
+     LoadResumable _ vt    -> Set.singleton vt
+     SaveResumable _ vt    -> Set.singleton vt
+     Output _ vt xs        -> Set.singleton vt `Set.union`
+                              Set.unions (fmap goOut xs)
+     KeepFactInHistory     -> Set.empty
+
+     Read _ _ vt ss
+      -> Set.singleton vt `Set.union`
+         typesOfStatement ss
```
tranma: I wonder if building up a list of types then just do Set.fromList at the end is faster. It's not a big deal though, perf is not bad atm.

tranma:
yep, WellTyped sounds good, just need to use pipeline stuff for it :+1:

src/Icicle/Sea/FromAvalanche/Analysis.hs
jystic: I'm not sure, I plucked this from the C++ branch.

I suspect you'd end up doing the same number of comparisons either way right? I'm not sure.

amosr: I'd be curious to see - I have no real intuition about why fromList would be faster

data/sea/20-simple.h
```diff
@@ -57,8 +57,9 @@ static ibool_t   INLINE istring_ne    (istring_t x, istring_t y) { return strcmp
 
 static istring_t INLINE istring_copy(imempool_t *into, istring_t val)
 {
-    char* alloc = imempool_alloc(into, strlen(val));
-    strcpy(alloc, val);
+    size_t val0_size = strlen (val) + 1;
+    char  *alloc     = imempool_alloc (into, val0_size);
+    memcpy (alloc, val, val0_size);
```
amosr: oh crap. good catch

src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -242,26 +239,68 @@ checkInputType state
...
+        [ ""
+        , "if (value_size == sizeof (\"true\") - 1 &&"
+        , "    memcmp (\"true\", value_ptr, value_size)) {"
+        , "    program->" <> pretty nx <> "[program->new_count] = itrue;"
+        , "} else if (value_size == sizeof (\"false\") - 1 &&"
+        , "           memcmp (\"false\", value_ptr, value_size)) {"
+        , "    program->" <> pretty nx <> "[program->new_count] = ifalse;"
+        , "} else {"
+        , "    return psv_alloc_error (\"not a boolean\", value_ptr, value_size);"
+        , "}"
```
amosr: would these be nicer in a function in the preamble?

amosr:
:+1:

src/Icicle/Sea/FromAvalanche/Analysis.hs
jystic: I hope this is just as fast, I use this pattern all the time, containers yields such beautiful code in this arrangement :)

tranma: yeah. I was wondering if there would be fewer comparisons if you build up the set afterwards, but it's probably best to leave the performance fiddling to containers :)

PR #222, opened 2015-11-16
WIP: Kill melt bindings

amosr:


tranma:
hm, ok, it's a bit of a surprise to me that `constructor` can deal with Lets where the body needs melting, e.g. `let x = [(0,1)]:{Array (Sum Int Int)}`

amosr:
yeah.. it kind of makes sense though, because Lets on their own are useless, so they'll either be used by a Write, an Output or an InitAccumulator, which will do the unboxing itself

tranma:
ok. So any Lets that binds a "meltable" expression would either be: a) unused (and killed by thresher), or b) be melted by meltAccumulators. Is that right?

amosr:
yup, that's right. we could even substitute in *all* the lets, do constructor, then a-normalise and thresh, and we'd probably get very similar results. in a way we are pretending to substitute all the lets, because constructor keeps the environment and looks through bindings

tranma:
ok, that's good then. :cn:

jystic:
:dromedary_camel:

PR #223, opened 2015-11-16
test that melt/constructor leaves no "meltable" types behind

tranma:
checking that after simpFlattened, we don't get left with pack/unpack and other "meltable" expressions


jystic:
This is quite awesome :100:

tranma:
after updating to master, passed 10000 tests!

jystic:
:bomb: :boom:

PR #225, opened 2015-11-17
prop_seaworthy (no psv)

tranma:
not there yet, we still get errors like `compute function #0 - eval:3:5: error: unknown type name 'ibuf_t__ibuf_t__idate_t'; did you mean 'iarray_t__ibuf_t__idate_t'?`


amosr:
ah, yeah. so the problem is that you can describe programs with nested buffers and unboundedly nested arrays in Core, but not in Source. @jystic's psv test filters out the bad types

tranma:
cool, maybe I can just modify Core Arbitrary?

amosr:
I think maybe he has a "WellTypedAvalanche" or something with a nice arbitrary instance

tranma:
just like prop_psv without the psv now

amosr:
:+1:

PR #226, opened 2015-11-17
Partial evaluator for Avalanche

amosr:


src/Icicle/Avalanche/Statement/Simp/Eval.hs
```diff
@@ -0,0 +1,99 @@
...
+            => EvalPrim a n p
+            -> (p -> Type)
+            -> a
+            -> p
+            -> [Value a n p]
+            -> Maybe (Exp a n p)
+simpEvalP ev ty a_fresh p vs
+ -- TODO: evaluation function should not return value for underapplied primitives
+ -- uncomment if you dare
+ | length (functionArguments $ ty p) == length vs
```
tranma: is this why my previous attempt at constant folding for avalanche was botched?

amosr: yeah :-(

tranma: damnation, I thought primitives are always fully applied by this point

amosr: ah.. so yeah, they are, but if you have
``pack value var``
it will look at "pack value var" and can't simp, because one of the arguments isn't a value. however, then it recurses down and sees "pack value", which is only applied to values…

tranma: damn recursion!

amosr:
ok to merge?

tranma:
looks great :100:

Issue #214, opened 2015-11-13
Relation primops not being melted


amosr:
If you have something like "(a,b) < (b,c)" this isn't being melted out properly. Constructors only melts out ==, but should handle (<), (<=) (>) (>=) (/=) as well.

amosr:
I think this is pretty low priority. who needs comparisons on things

amosr:
turns out this caused some pain, but Tran has now fixed this

PR #228, opened 2015-11-17
melt Ne

tranma:
just did this https://github.com/ambiata/icicle/issues/214

also updated CLI tests, probably was made out of date from the constant folding branch

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
+++ b/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
@@ -35,7 +35,10 @@ constructor a_fresh statements
   xDefault t = xValue t (defaultOfType t)
 
   primAnd = primLogical  Min.PrimLogicalAnd
+  primOr  = primLogical  Min.PrimLogicalOr
+
   primEq  = primRelation Min.PrimRelationEq
+  primNe  = primRelation Min.PrimRelationNe
 
   primFold1 append xs
    = case xs of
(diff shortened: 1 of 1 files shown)


src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -211,17 +217,24 @@ constructor a_fresh statements
...
 
    | otherwise
    = x
 
+  relation Min.PrimRelationGt = primGt
+  relation Min.PrimRelationGe = primGe
+  relation Min.PrimRelationLt = primLt
+  relation Min.PrimRelationLe = primLe
+  relation Min.PrimRelationEq = primEq
+  relation Min.PrimRelationNe = primNe
```
amosr: Is this right?
```
(a,b) /= (c,d)
```
should really be
```
a /= c || b /= d 
```

and the less than is more complicated, I think

tranma: oh yeah, should have actually thought about it for a second

amosr: it's more annoying than it looks! less than is something like..

```
(a,b) < (c,d)
==>
(a < c) || (a == c && b < d)
```
I think.
I don't think ord is particularly important but if /= is easy to fix might as well do that

tranma:
I'll just merge this? Ord can come later after psv stuff is done

amosr:
yup, sounds good :100:

PR #231, opened 2015-11-17
Collect statistics for number of facts + number of entities processed

jystic:
This is used to show fact/entity throughput and also to put line numbers on error messages.

This also fixes the "PSV missing output for last item" problem.


amosr:
I'm a real boy
![http://www.disneyclips.com/imagesnewb/imageslwrakr01/mar235.gif]

PR #232, opened 2015-11-17
Topic/growing arrays

amosr:


data/sea/10-mempool.h
```diff
@@ -105,3 +105,11 @@ void imempool_free (imempool_t *pool)
...
     free(pool);
 }
+
+
+static iint_t INLINE next_power_of_two(iint_t i)
+{
+    iint_t bits = 64 - __builtin_clzll (i);
+    iint_t next = 1L << bits;
+    return next;
+}
```
tranma: :o

jystic: Lol, this was one of the alternatives:
```c
unsigned int const v; // Round this 32-bit value to the next highest power of 2
unsigned int r;       // Put the result here. (So v=3 -> r=4; v=8 -> r=8)

if (v > 1) {
    float f = (float)v;
    unsigned int const t = 1U << ((*(unsigned int *)&f >> 23) - 0x7f);
    r = t << (t < v);
} else {
    r = 1;
}
```

jystic:
:mailbox_with_mail:

PR #237, opened 2015-11-18
Dead code removal

amosr:
Especially for extra buffers produced from latest

### DIFF
--- a/test/Icicle/Test/Avalanche/Flatten.hs
+++ b/test/Icicle/Test/Avalanche/Flatten.hs
@@ -38,7 +38,7 @@ namer = AC.namerText (flip Var 0)
 
 
 -- Flattening - removing all folds keeps value same
-zprop_flatten_commutes_value t =
+prop_flatten_commutes_value t =
  forAll (programForStreamType t)
  $ \p ->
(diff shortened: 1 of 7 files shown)


test/Icicle/Test/Avalanche/Flatten.hs
```diff
@@ -38,7 +38,7 @@ namer = AC.namerText (flip Var 0)
 
 
 -- Flattening - removing all folds keeps value same
-prop_flatten_commutes_value t =
+zprop_flatten_commutes_value t =
```
tranma: why is this test disabled again?

tranma:
nice. Would be great to have a test with some dead code being removed? :dancers:

test/Icicle/Test/Avalanche/Flatten.hs
amosr: haha sorry. that was while I was running the tests from ghci

amosr:
yeah.. maybe I should add another test. but one of the buffers does disappear from an avalanche test!

jystic:
:vibration_mode:

PR #238, opened 2015-11-19
more psv output

tranma:
needs some cleanup but outputting Pairs and Options seems to work

### DIFF
--- a/data/sea/50-psv.h
+++ b/data/sea/50-psv.h
@@ -2,8 +2,6 @@
 
 #if !ICICLE_NO_PSV
 
-#define OUTPUT_BUF_SIZE 128
-
 /* forward declarations for types, implemented by generated code */
 typedef struct ifleet ifleet_t;
 
@@ -60,8 +58,10 @@ static psv_error_t INLINE psv_read_fact
   , idate_t       date );
 
 /* psv driver */
-static const size_t psv_max_row_count = 128;
-static const size_t psv_buffer_size   = 16*1024;
+static const size_t psv_max_row_count   = 128;
+static const size_t psv_buffer_size     = 16*1024;
+static const size_t psv_output_buf_size = 128;
+
 
 static psv_error_t psv_error_add_line (iint_t line, psv_error_t original)
 {
--- a/data/sea/50-psv.h
+++ b/data/sea/50-psv.h
@@ -2,7 +2,7 @@
 
 #if !ICICLE_NO_PSV
 
-#define OUTPUT_BUF_SIZE 1024
+#define OUTPUT_BUF_SIZE 128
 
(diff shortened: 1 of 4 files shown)


data/sea/50-psv.h
```diff
@@ -2,6 +2,8 @@
 
 #if !ICICLE_NO_PSV
 
+#define OUTPUT_BUF_SIZE 128
```
jystic: Maybe put this with the other constants a bit further down:
```c
/* psv driver */
static const size_t psv_max_row_count = 128;
static const size_t psv_buffer_size   = 16*1024;
```

src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -462,77 +463,159 @@ seaOfWriteProgramOutput state = do
...
         attrib      = seaOfEscaped name
+        free s      = indent 4 $ "free (" <> s <> ");"
+        calloc  n s = "char *" <> n <> " = calloc(1," <> s <> " );"
+        decli   n   = "int   " <> n <> " = 0;"
+    in  do (str, bufs) <- strOfOutput ps oname otype0 ts0 ixStart
+           case bufs of
+             ((b,_,_):_) -> pure
+                    $ vsep
+                    $    ["{" ]
+                      <> fmap (\(n,s,_) -> indent 4 $ calloc n s) (List.reverse bufs)
```
jystic: If you wanted you could allocate this out of the mempool which will avoid the function call overhead, also you don't need to worry about freeing it then, the gc will do it.

jystic:
Looking good :+1:

src/Icicle/Sea/FromAvalanche/Psv.hs
tranma: oh cool! It's easy to free this though, maybe less work on the gc is better? dunno, see how outputting arrays goes

tranma:
can output pairs of any nesting as a flat array

amosr:
I like this. it's kind of silly, but you can say
```
feature asx ~> mean open, mean high, mean low
```
and operator precedence just nests it into a (Double,(Double,Double)) or maybe the other way around.
then if we just print that as ``[double, double, double]`` that's neat.

:+1:

jystic:
> I like this. it's kind of silly, but you can say
> ```
> feature asx ~> mean open, mean high, mean low
> ```
> and operator precedence just nests it into a (Double,(Double,Double)) or maybe the other way around.
> then if we just print that as ``[double, double, double]`` that's neat.

yeah that seems like it would be quite useful!

tranma:
this should be ready for final :eyes: 
then I'll try outputting arrays

jystic:
:spaghetti:

PR #241, opened 2015-11-20
psv output arrays

tranma:
```
[feature.salary_test_array]
  expression="""
    feature salary ~> latest 5 ~> value
  """
homer|salary_test_array|[65000,66000,70000,81000,90100]
```


jystic:
This looks good :rotating_light: once you change the payload_n thing

I would really consider pretty printing a call to  `PrimUnsafe (PrimUnsafeArrayIndex a)` instead though as those macros are "implementation details" that aren't very type safe. I can easily see us changing the macros in the future and having generated code which is a hard to debug.

tranma:
you're right, calling the Prim is much nicer. I misunderstood what you said IRL

amosr:
:+1:

Issue #243, opened 2015-11-23
Latest inside group fold doesn't work


amosr:
Maybe latest should be disallowed inside a group fold. (See typechecking invariants, Icicle/Source/Check/Invariants I think)
Or, maybe the conversion inside group fold is missing a "withInputName" or "convertWithInput" or something, that tells the rest of the convert which binding to use as the current element.

```
feature animal
~> group fold (animal
              ,kermit) = (group 1.2181434427721116
                          ~> 3)
~> 
latest 1
~> (latest 1 ~> `2005-06-08`)
Program (source : Stream Unit)
Precomputations:

Stream transformers:
core$0 =
    source
Reductions:
core$4 =
    rfold (
    \core$6 
    \core$7 Map_insertOrUpdate# (
            \core$9 ()) (())
            (1.2181434427721116) core$6)
    (Map []) core$0
Postcomputations with date as waldorf$core$1:
core$5 =
    Map_fold# (
    \core$6 
    \core$10 
    \core$7 Map_insertOrUpdate# (
            \core$13 core$13) (3) core$10
            core$6) (Map []) core$4
core$2 =
    
    let core$20 = Map_fold# (
                  \core$3 
                  \animal$core$16 
                  \kermit$core$17 Latest_push#
                                  core$3 core$0) (Buf []) core$5
     in 
    let core$23 = 
                  let core$18 = Latest_read#
                                core$20
                   in Array_fold# (
                      \core$22 
                      \core$21 Latest_push# core$22
                               (2005-06-08)) (Buf []) core$18
     in Latest_read# core$23
Returning:
gonzo =
    core$2

Left (ProgramErrorPost (ExpErrorVarNotInEnv (NameMod (Variable "core") (Name (Variable "0")))))
```

tranma:
does `feature salary ~> group fold (k,v) = (group date ~> count value) ~> latest 2 ~> count v` make sense? I'm not sure it does.

amosr:
yeah, probably not actually

tranma:
on the other hand, I suppose `> feature salary ~> group fold (k,v) = (group date ~> count value) ~> latest 2 ~> max k` is a legitimately different intention to `> feature salary ~> group fold (k,v) = (group date ~> count value) ~> max k`

what do you think? should we disallow latest inside group folds? or massage it to work?

amosr:
[...]

PR #244, opened 2015-11-23
ban latest inside group folds

tranma:
the semantics of latest inside a group fold is unclear. We should just ban it from Source.

https://github.com/ambiata/icicle/issues/243


amosr:
:bamboo:

PR #245, opened 2015-11-23
Chord file output

amosr:


src/Icicle/Sea/Chords/File.hs
```diff
@@ -0,0 +1,69 @@
...
+       version = 1
+       dates   = maximumOfDates chordmap
+   in  mconcat
+     [ BS.string8 magic
+     , BS.int64LE version
+     , BS.int64LE $ fromIntegral dates ]
+
+record :: (D.Entity, [D.DateTime]) -> BS.Builder
+record (ent, dates)
+ = let entity       = BSC.pack $ T.unpack $ D.getEntity ent
```
jystic: You can use `T.encodeUtf8` to avoid the linked list

jystic:
:cherries:

tranma:
it's beautiful :cactus:

PR #246, opened 2015-11-23
Support for reading the Ambiata special recipe chord file

jystic:
This consumes the format produced by @amosr's work earlier today.

I'll join them together in a later PR.

### DIFF
--- a/data/sea/51-chord.h
+++ b/data/sea/51-chord.h
@@ -1,0 +1,133 @@
+#include "50-psv-error.h"
+
+#if !ICICLE_NO_PSV
+
+typedef struct {
+    const iint_t  entity_size;
+    const iint_t  count;
+    const char    entity[0];
+/*  const char    entity[entity_size + 1]; */
+/*  const idate_t dates[count];            */
+} ichord_t;
+
+typedef struct {
+    const iint_t   magic;   /* CHORDATA */
+    const iint_t   version; /* 1 */
+    const iint_t   max_chord_count;
+    const ichord_t chords[0];
+} ichord_header_t;
+
+typedef struct {
+    void           *mapped_ptr;
+    size_t          mapped_size;
+    iint_t          max_chord_count;
+    const ichord_t *chords;
+} ichord_file_t;
+
+static psv_error_t ichord_file_mmap (int fd, ichord_file_t *file)
+{
+    static ichord_file_t empty;
+    *file = empty;
+
+    if (fd == 0) {
+        file->max_chord_count = 1;
+        return 0;
+    }
+
+    struct stat stat;
+    int stat_error = fstat (fd, &stat);
+
+    if (stat_error != 0)
+        return psv_alloc_error ("failed to stat chord file", 0, 0);
+
+    size_t mapped_size = stat.st_size;
+    void  *mapped_ptr  = mmap (0, mapped_size, PROT_READ, MAP_PRIVATE, fd, 0);
+
+    if (mapped_ptr == MAP_FAILED)
+        return psv_alloc_error ("failed to memory map chord file", 0, 0);
+
+    file->mapped_ptr  = mapped_ptr;
+    file->mapped_size = mapped_size;
+
+    ichord_header_t *hdr = mapped_ptr;
+
+#if ICICLE_DEBUG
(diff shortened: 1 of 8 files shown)


data/sea/51-chord.h
```diff
@@ -0,0 +1,133 @@
...
+
+    return (const idate_t *)entity_end;
+}
+
+static const ichord_t * INLINE ichord_next (const ichord_t *chord)
+{
+    const idate_t  *date_end = ichord_dates (chord) + chord->count;
+    const ichord_t *next     = (const ichord_t *)date_end;
+
+    return next->entity_size == 0 ? 0 : next;
```
amosr: I should probably explicitly filter out empty entity in the chord file output

data/sea/51-chord.h
```diff
@@ -0,0 +1,133 @@
...
+
+static const idate_t * INLINE ichord_dates (const ichord_t *chord)
+{
+    size_t      entity0_size = (size_t) chord->entity_size + 1;
+    const char* entity_end   = chord->entity + entity0_size;
+
+    return (const idate_t *)entity_end;
+}
+
+static const ichord_t * INLINE ichord_next (const ichord_t *chord)
```
amosr: as we start writing more and more complicated C, I'd be tempted to start encoding invariants and preconditions as a Hoare logic sort of thing. doesn't have to be exact, but something like
```
{ chord /= 0 /\ VALID_CHORD_FILE }
chord' = ichord_next(chord)
{ chord' == 0 \/ chord'->entity > chord->entity }
```

just an idea for the future, I wouldn't worry about going back and adding them to this

src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -138,9 +133,9 @@ seaOfAllocProgram mode state
                 <> calloc "psv_max_row_count" (seaOfValType t)
 
    in vsep [ "/* " <> seaOfAttributeDesc (stateAttribute state) <> " */"
-           , programs <> " = " <> calloc maxChords stype
+           , programs <> " = " <> calloc "max_chord_count" stype
            , ""
-           , "for (iint_t ix = 0; ix < " <> maxChords <> "; ix++) {"
+           , "for (iint_t ix = 0; ix < max_chord_count; ix++) {"
```
amosr: oh wow, so the maximum chord count is totally dynamic? that's great

src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -148,21 +143,22 @@ seaOfAllocProgram mode state
...
  [ "#line 1 \"collect fleet state\""
  , "static void psv_collect_fleet (ifleet_t *fleet)"
  , "{"
- , "    imempool_t *into_pool   = imempool_create ();"
- , "    imempool_t *last_pool   = fleet->mempool;"
- , "    iint_t      chord_count = fleet->chord_count;"
+ , "    imempool_t *into_pool       = imempool_create ();"
+ , "    imempool_t *last_pool       = fleet->mempool;"
+ , "    iint_t      max_chord_count = fleet->max_chord_count;"
+ , "    iint_t      chord_count     = fleet->chord_count;"
```
amosr: ah. so this is the number of chords on the current entity - so we only need to garbage collect the first chord_count states?

amosr:
superlative :sunflower:

data/sea/51-chord.h
jystic: Yeah that's a great idea, I wonder if there's anything that can verify those kind of things?

src/Icicle/Sea/FromAvalanche/Psv.hs
jystic: Yeah! It turned out to be easier to do it this way :)

src/Icicle/Sea/FromAvalanche/Psv.hs
jystic: Yep that's right :)

data/sea/51-chord.h
amosr: actually, there was a talk at SAPLING about "verifying C using Whiley". I just asked the guy, and he said "Hey, at the moment it's only really good for smallish array programs like the ones I did at the talk."
I think we'd have to hand-convert from Whiley to C too, which is a shame

PR #247, opened 2015-11-23
WIP: killing dummy annotations

amosr:
This would be nice to do eventually, but it is a lot of work for little/no immediate gain


amosr:
I'm killing this

PR #249, opened 2015-11-24
Only read the input value once

jystic:
This pushes the chord loop down in to read_fact_XX so that we only need to parse a value once, not once per chord date.


amosr:
brill :bride_with_veil:

PR #251, opened 2015-11-24
revamp psv output

tranma:
this uses a fixed buffer for PSV outputs and flushes to file whenever the buffer is full.

### DIFF
--- a/src/Icicle/Sea/FromAvalanche/Psv.hs
+++ b/src/Icicle/Sea/FromAvalanche/Psv.hs
@@ -9,7 +9,7 @@ module Icicle.Sea.FromAvalanche.Psv (
   ) where
 
 import qualified Data.ByteString as B
-import           Data.Char
+import           Data.Foldable (maximum)
 import qualified Data.List as List
 import           Data.Map (Map)
 import qualified Data.Map as Map
@@ -716,9 +716,6 @@ mappingOfField (StructField fname, ftype) vars0 = do
 
 ------------------------------------------------------------------------
 
-type BufSize = Doc
-type BufLen  = Doc
-
 cond :: Doc -> Doc -> Doc
 cond n body
  = vsep ["if (" <> n <> ")"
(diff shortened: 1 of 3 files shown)


src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -726,21 +724,57 @@ cond n body
...
+ , outputDie
+ ]
+
+forStmt :: Doc -> Doc -> Doc -> Doc
+forStmt i n m
+ = "for(iint_t" <+> i <+> "= 0," <+> n <+> "=" <+> m <> ";" <+> i <+> "<" <+> n <> "; ++" <> i <> ")"
+
+outputChar :: Doc -> Doc
+outputChar x
+ = iprintf "%c" ("'" <> x <> "'")
```
jystic: from our conversation IRL, does this work?

jystic:
This is great, no allocation! :100:

src/Icicle/Sea/FromAvalanche/Psv.hs
tranma: yeah, just needs to use single quote there

jystic: all good then :palm_tree:

src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -828,163 +851,132 @@ seaOfWriteOutput ps oname@(OutputName name) otype0 ts0 ixStart
...
     hasDateTime _           = False
 
-seaOfOutput
-  :: Doc -> OutputName -> ValType -> [ValType] -> Int
-  -> Doc                                        -- buffer to print to
-  -> Either SeaError (Doc, [(BufSize, BufLen)]) -- name, alloc size and len
-
-seaOfOutput ps oname@(OutputName name) otype0 ts0 ixStart dstbuf
+seaOfOutput :: Doc -> OutputName -> ValType -> [ValType] -> Int -> Either SeaError Doc
+seaOfOutput ps oname@(OutputName name) otype0 ts0 ixStart
```
jystic: In the end this should really return the `[ValType]` that was not consumed so it can be called recursively like [unmeltValue'](https://github.com/ambiata/icicle/blob/e68df32f1cc536aac783c3bbbcf2e031c10f98ae/src/Icicle/Avalanche/Prim/Eval.hs#L287)

jystic: Probably it needs to return the index which it is up to also

tranma: yeah, ok, I'll need to do something like that for better outputting.

Gonna merge this and do that in another PR

jystic: Yeah go for it, this is awesome

PR #252, opened 2015-11-24
Disable non-blocking mode for all the file descriptors

jystic:
This was causing an error when icicle was trying to read from sort's stdout as sort can't keep up


amosr:
I'm a little confused - I thought blocking was the default?

jystic:
It is, but Haskell understandably must set the fd's to non blocking so that it works nicely with green threads.

jystic:
When you did the fd creation code for file you used posix functions directly, so the default was blocking, but when we convert a Haskell `Handle` I guess they are set to non-blocking.

amosr:
ahh okay. that makes sense. maybe make a note about that in there?
:+1:

PR #254, opened 2015-11-24
Format update

jystic:
I needed these changes to make the ice tests pass


src/Icicle/Sea/FromAvalanche/Psv.hs
```diff
@@ -1008,14 +1005,14 @@ seaOfOutputBase quoteDate err t val
      IntT
       -> pure $ vsep [iprintf "%lld" val]
      DoubleT
-      -> pure $ vsep [iprintf "%f" val]
+      -> pure $ vsep [iprintf "%.17g" val]
```
jystic: This uses either `%f` or `%e` whatever is shorter, but always has 17 digits of precision, trailing zeros are dropped

amosr:
:+1:

PR #255, opened 2015-11-25
More format modifications

jystic:


amosr:
all aboard the gravy train :steam_locomotive:

Issue #263, opened 2015-11-27
Attribute should be case insensitive / lowercase


jystic:
We should lowercase all attributes so that they are valid variable names.

amosr:
I think this is done

PR #264, opened 2015-11-30
tolower attribute names so they can be used as var names

tranma:
a hack for https://github.com/ambiata/icicle/issues/263

### DIFF
--- a/main/icicle-bench.hs
+++ b/main/icicle-bench.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE NoImplicitPrelude   #-}
+{-# LANGUAGE ViewPatterns        #-}
+{-# LANGUAGE PatternGuards       #-}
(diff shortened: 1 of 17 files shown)


tranma:
hungarian notation is an option, is it better? 😅

amosr:
Need to modify the Source parser too, so it can parse "feature ALL_CAPS": at the moment this is tokenised to [KeywordFeature, Constructor "ALL_CAPS"]. best to just modify feature parser to accept variables or constructors

amosr:
right - I guess it isn't necessary, but I think it'd be ideal

tranma:
uh, is that a lexer change?

amosr:
no, I think it's best to just change the parser for "feature".

I'd rather not modify the lexer, because I like to keep the "Constructor" and "variable" distinction in case statements, so it's a little more obvious what is a catch-all variable, and what is a misspelt constructor

```
case foo
  | Nonez  -> 0
  | Some x -> x + 1
```
if variable & constructors were lumped together, this would be fine, although maybe the overlap check would notice that one..

PR #266, opened 2015-11-30
Support capitalised concrete feature names

jystic:
I think maybe we don't need https://github.com/ambiata/icicle/pull/264 if we change this.

Sorry for the wild goose chase @tranma :(


tranma:
cool, that's better

jystic:
Cool, thanks anyways @tranma for doing the lowercase thing :)

PR #267, opened 2015-12-01
va_list is a horrible hack

jystic:


tranma:
I tried so hard to get this right....
:100:

jystic:
I'm curious how va_list actually works under the hood now, maybe i don't want to know

tranma:
well, given that you have to give va_start the argument immediately before the "..." ....

PR #272, opened 2015-12-03
Use the Grisu2 algorithm for printing doubles

jystic:
This gives a 25 minute speedup on our 320G test.

http://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf


amosr:
amazing
:flags:

PR #274, opened 2015-12-03
Fix bench test... not sure why it was working earlier. maybe I was no…

amosr:
…t working off latest master


amosr:
gonna merge yolo

PR #281, opened 2015-12-06
Buffers n strings

jystic:
- Buffers are now always unboxed
- Constant strings are output using @amosr strings

### DIFF
--- a/data/sea/00-includes.h
+++ b/data/sea/00-includes.h
@@ -60,3 +60,6 @@ STATIC_ASSERT(sizeof (void *) == sizeof (uint64_t), icicle_only_supports_systems
 #else
 #   define ICICLE_WHEN_DEBUG(code)
 #endif
+
+#define CONCAT0(x, y) x ## y
+#define CONCAT(x, y)  CONCAT0(x, y)
--- a/data/sea/31-buffer.h
+++ b/data/sea/31-buffer.h
@@ -1,27 +1,25 @@
 #include "30-array.h"
 
-typedef struct
-{
-    iint_t max_size;
-    iint_t cur_size;
-    iint_t head;
-} ibuf_struct;
+#define BUF(n,t)         ibuf_ ## n ## _ ## t
+#define BUF_FUN(n,t,fun) CONCAT(BUF(n,t), _ ## fun)
+#define BUF_T(n,t)       CONCAT(BUF(n,t), _t)
+
 /*
  Invariants
-  0 <= cur_size <= max_size
-  0 <= head     <  max_size
+  0 <= size <= n
+  0 <= head <  n
 
-  read = ARRAY[ head..(head+cur_size) % max_size )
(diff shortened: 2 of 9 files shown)


data/sea/31-buffer.h
```diff
@@ -112,101 +93,141 @@ Read(buf)
...
+                                                                                \
+    for (iint_t ix = 0; ix != min; ++ix) {                                      \
+        iint_t x_ix = (ix + x.head) % n;                                        \
+        iint_t y_ix = (ix + y.head) % n;                                        \
+                                                                                \
+        if (!t##_lt(x.vals[x_ix], y.vals[y_ix]))                                \
+            return ifalse;                                                      \
+    }                                                                           \
+                                                                                \
+    return x.size < y.size;                                                     \
```
amosr: CONCAT(iint_, op) (x.size, y.size)
?

data/sea/00-includes.h
```diff
@@ -60,3 +60,6 @@ STATIC_ASSERT(sizeof (void *) == sizeof (uint64_t), icicle_only_supports_systems
 #else
 #   define ICICLE_WHEN_DEBUG(code)
 #endif
+
+#define CONCAT0(x, y) x ## y
```
tranma: this fine without parens?

src/Icicle/Sea/FromAvalanche/Prim.hs
```diff
@@ -20,48 +22,62 @@ import           Icicle.Sea.FromAvalanche.Type
...
 
+data ArgConv
+ = ArgVal
+ | ArgRef
+
 data PrimDoc
- = PDFun   Doc
- | PDAlloc Doc
+ = PDFun   Doc (Maybe [ArgConv])
+ | PDAlloc Doc (Maybe [ArgConv])
```
amosr: neat

data/sea/00-includes.h
jystic: I needs to have no parens or they'll become part of the string

src/Icicle/Sea/FromAvalanche/Program.hs
```diff
@@ -32,7 +32,7 @@ import           Icicle.Sea.FromAvalanche.Prim
 import           Icicle.Sea.FromAvalanche.State
 import           Icicle.Sea.FromAvalanche.Type
 
-import           P
+import           P hiding (head)
```
amosr: hiding head?

jystic: I wanted to use it as a variable name (for buffer)

Fwiw, I don't even know why we have head in P??

amosr: I'm not sure. is it safe head? maybe it's just re-exported with a warning like error. hmm..

amosr:
so beautiful :+1:

PR #282, opened 2015-12-07
Text conversion cleanup

jystic:
I did a clean up of the current text conversion code to make it a bit easier to understand and more consistent. I thought it would be a good idea to do this before I start writing optimised versions of these functions.

### DIFF
--- a/data/sea/42-text-conversion.h
+++ b/data/sea/42-text-conversion.h
@@ -18,56 +18,59 @@ static const size_t text_iint_max_size = 20; /* 64-bit integer = 19 digits + sig
 
 static ierror_msg_t INLINE text_read_iint (char **pp, char *pe, iint_t *output_ptr)
 {
-    char *p            = *pp;
+    char  *p           = *pp;
     size_t buffer_size = pe - p;
 
-    size_t n = 0;
-    while (n < buffer_size) {
-        if (!text_is_digit (p[n]))
-            break;
-        n++;
+    /* handle negative */
+    int sign      = 1;
+    int sign_size = 0;
+    if (buffer_size > 0 && p[0] == '-') {
+        sign      = -1;
+        sign_size = 1;
+        p++;
+        buffer_size--;
     }
 
-    if (n == 0)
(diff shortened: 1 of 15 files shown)


data/sea/42-text-conversion.h
```diff
@@ -0,0 +1,357 @@
...
+        case  2:  value += (p[n- 2] - '0') * 10;
+        case  1:  value += (p[n- 1] - '0');
+                  value *= sign;
+                  *output_ptr = value;
+                  *pp += n;
+                  return 0;
+
+        default:  return ierror_msg_alloc ("integer too big", p, n);
+    }
+}
```
jystic: This new `text_read_iint` speeds up the ASX 500MB dataset from 410MB/s -> 630MB/s on my computer.

data/sea/42-text-conversion.h
```diff
@@ -0,0 +1,357 @@
...
+{
+    char *p            = *pp;
+    size_t buffer_size = pe - p;
+
+    size_t n = 0;
+    while (n < buffer_size) {
+        if (!text_is_digit (p[n]))
+            break;
+        n++;
+    }
```
jystic: I'm hoping to replace this with a call to `_mm_cmpestr*`

olorin:
Nice. (From a cursory glance, looks like a lot of the complexity in the `glibc` implementation is the base handling.)

amosr:
looks good :+1:

amosr:
somewhat unrelated, but while you're in there, perhaps make sure this time reading code won't dereference before the start of the buffer. (I guess you could also pad the start of the buffer if you wanted?)
https://github.com/ambiata/icicle/blob/topic/text-conversion-cleanup/data/sea/52-psv.h#L116

jystic:
I wonder if it's ok because it's on the stack? so it's already padded in a sense

amosr:
hmm... yes, I suppose so. the stack grows downwards right, so it'd probably be padded by variables in later functions? I'm not sure, I think I'd rather do it explicitly

jystic:
So something like adding 16 bytes to the start and the end of the buffer?

PR #283, opened 2015-12-08
build on both 7.10 and 7.8 (WIP)

tranma:


amosr:
groovy

amosr:
:watch:

jystic:
Superseded by https://github.com/ambiata/icicle/pull/320

src/Icicle/Source/Transform/Desugar.hs
```diff
@@ -322,7 +323,10 @@ data Tree a n x
          [(Constructor, Tree a n x)] (Tree a n x)
  deriving (Functor, Foldable, Traversable, Show)
 
-
+instance Applicative (Tree a n) where
+  pure = return
```
jystic: fwiw, this doesn't work according to the docs, not sure what the effect is, but they say you're supposed to move the `return` implementation here and make `return = pure`

PR #284, opened 2015-12-09
Fledgeling benchmark suite

jystic:
This just has the one stupid benchmark, but hopefully gets all the infrastructure in place for us to add more things over time.


src/Icicle/Benchmark.hs
```diff
@@ -0,0 +1,219 @@
...
+  first BenchSourceError $ do
+    let inlined = sourceInline dict virtual
+
+    desugared    <- sourceDesugarQT inlined
+    (checked, _) <- sourceCheckQT dict desugared
+
+    let reified = sourceReifyQT checked
+
+    core <- sourceConvert dict reified
+    let simplified = coreSimp core
```
tranma: this (and above) is a bit like Pipeline stuff? Is it worth putting them in the main pipeline for repl and tests to use?

tranma:
the generators are super cool @jystic, they could even be useful outside of icicle!
:100:

src/Icicle/Benchmark.hs
jystic: yeah I think so, probably could simplify ice as well actually

tranma: yeh I saw some of these exact functions in ice

PR #285, opened 2015-12-09
Fix Haddock errors

jystic:
This is breaking the build on `master` at the moment


tranma:
:+1: thanks

jystic:
@amosr if you can think of a different style for formatting records that you'd prefer, I'd be happy to change it

PR #286, opened 2015-12-10
Add a bunch more benchmarks

jystic:
Would be good to vary the dictionary as well for all of these.


amosr:
looks very useful :+1:

Issue #287, opened 2015-12-10
Segfault when running prop_psv


jystic:
Just writing down that I saw this happen on Travis:

```
=== prop_psv from test/Icicle/Test/Sea/Psv.hs:51 ===
Test suite test: FAIL
```

Which likely means that it segfaulted, we really need that crazy forking machinery or a signal handler to make this better.

amosr:
I wonder if it could be the date parsing thing I ran into for merge, where the data was something like "|\n" and it found the \n and looked back 20 chars for a pipe

jystic:
hmm, the psv test is only generating valid data at the moment though :(

tranma:
saw this again in https://travis-ci.org/ambiata/icicle/builds/102267649

I've never seen this locally, or on jenkins though. Not sure if it's platform-related

tranma:
I'm looking at this, looks like a good old null pointer :)

```
* thread #1: tid = 0xe457b, 0x000000010cc341f4 jetski.dylib`iarray_ibool_index(x=0x0000000000000000, ix=1) + 20 at type definitions:1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
  * frame #0: 0x000000010cc341f4 jetski.dylib`iarray_ibool_index(x=0x0000000000000000, ix=1) + 20 at type definitions:1
    frame #1: 0x000000010cc32f1b jetski.dylib`psv_write_outputs(fd=13, buffer=0x0000000101d83000, buffer_end=0x0000000101dc2fff, buffer_ptr_ptr=0x00007fff5fbfb770, entity=0x0000000102883000, entity_size=4, fleet=0x0000000101f0db00) + 747 at write all outputs:52
``

amosr:
did you fix this, @tranma ?

tranma:
@amosr pretty sure I did. Passed 100000 tests without segfault, might as well be labelled "verified compiler"!

PR #288, opened 2015-12-10
Add segfault handler to PSV test

jystic:
We can now trap segmentation faults and print a stack trace as well as some user data before exiting.

In the case of the PSV test, the user data is the test case which caused the segfault.

### DIFF
--- a/data/sea/06-segv.h
+++ b/data/sea/06-segv.h
@@ -2,8 +2,8 @@
 
 /* this global is pretty nasty, don't know how else we can get
  * information in a signal handler though */
-static const size_t segv_user_data_size = 1024*1024;
-static char segv_user_data[segv_user_data_size];
+#define SEGV_USER_DATA_SIZE (1024*1024)
+static char segv_user_data[SEGV_USER_DATA_SIZE];
 
 static void segv_handler (int sig)
 {
@@ -19,7 +19,7 @@ static void segv_handler (int sig)
 
 void segv_install_handler (const char *user_data, size_t user_data_size)
 {
-    size_t size = MIN (user_data_size, segv_user_data_size-1);
+    size_t size = MIN (user_data_size, SEGV_USER_DATA_SIZE-1);
     memcpy (segv_user_data, user_data, size);
     segv_user_data[size] = '\0';
 
--- a/data/sea/06-segv.h
+++ b/data/sea/06-segv.h
@@ -1,0 +1,44 @@
+#include "05-error.h"
+
+/* this global is pretty nasty, don't know how else we can get
+ * information in a signal handler though */
+static const size_t segv_user_data_size = 1024*1024;
+static char segv_user_data[segv_user_data_size];
+
+static void segv_handler (int sig)
+{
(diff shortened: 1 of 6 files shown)


jystic:
This should solve https://github.com/ambiata/icicle/issues/287 more or less

tranma:
oh wow, this is really impressive @jystic

data/sea/06-segv.h
```diff
@@ -0,0 +1,44 @@
...
+#define SEGV_USER_DATA_SIZE (1024*1024)
+static char segv_user_data[SEGV_USER_DATA_SIZE];
+
+static void segv_handler (int sig)
+{
+    void *frames[50];
+    size_t n_frames = backtrace (frames, sizeof(frames));
+
+    fprintf (stderr, "icicle: segmentation fault:\n");
+    backtrace_symbols_fd (frames, n_frames, STDERR_FILENO);
```
amosr: this is amazing. I had no idea this was possible

data/sea/06-segv.h
```diff
@@ -0,0 +1,44 @@
...
+    fprintf (stderr, "\n%s\n", segv_user_data);
```
tranma: I wonder what happens when the segv_handler segfaults

data/sea/10-mempool.h
```diff
@@ -1,4 +1,4 @@
-#include "05-error.h"
+#include "06-segv.h"
 
 const size_t iblock_size = 1 * 1024 * 1024;
```
amosr: gcc doesn't complain about this one though? maybe it just doesn't like ``static const``?

data/sea/52-psv.h
```diff
@@ -61,11 +61,20 @@ static ierror_msg_t INLINE psv_read_fact
...
 
-/* psv driver */
+
+/*
+Constants
+*/
+
 static const size_t psv_max_row_count   = 128;
 static const size_t psv_buffer_size     = 16*1024;
 static const size_t psv_output_buf_size = 16*1024;
```
amosr: hmm never mind

data/sea/06-segv.h
jystic: o_O probably an infinite loop

jystic: I think i'll change this to put the default one back as the first thing it does, then hope that doing that doesn't segfault.

amosr:
great :point_right:

data/sea/06-segv.h
jystic: segfaults per second
^ yep, this happens

data/sea/10-mempool.h
jystic: i think it's because of where it's used rather than the declaration

data/sea/06-segv.h
tranma: well if this handle segfaults too, probably stack overflow, it's all good :)

tranma: ah ok, that's worse!

PR #289, opened 2015-12-10
Out of order input data is now rejected

jystic:


data/sea/42-text-conversion.h
```diff
@@ -245,16 +243,16 @@ static ierror_msg_t INLINE fixed_read_itime (const char *p, const size_t size, i
...
 
-        const iint_t month = p[5] * 10
-                           + p[6];
+        const iint_t month = (p[5] - '0') * 10
+                           + (p[6] - '0');
 
-        const iint_t day   = p[8] * 10
-                           + p[9];
+        const iint_t day   = (p[8] - '0') * 10
+                           + (p[9] - '0');
```
tranma: is this why the test was segfaulting?

markhibberd:
:100: :fire_engine:

data/sea/42-text-conversion.h
jystic: I don't think so, it would have just had very weird dates, this would have been `49 * 10` instead of `1 * 10` for example

tranma:
looks ok from eyeballing, just needs to change the test to sort the input first, I think. :eyes:

data/sea/52-psv.h
```diff
@@ -142,10 +148,9 @@ static ierror_msg_t psv_read_buffer (psv_state_t *s)
         const char  *value_end  = time_ptr - 1;
         const size_t value_size = value_end - value_ptr;
 
-        const bool new_entity = entity_cur_size != entity_size
-                             || memcmp (entity_cur, entity_ptr, entity_size) != 0;
+        const int new_entity = memcmp (entity_cur, entity_ptr, entity_size);
```
amosr: I'm not sure about this. I think you need to find the minimum size, then if it's 0 subtract the two sizes.

suppose the old entity is ``ABC`` and the new one is ``AB``. I think this would compare with the pipe, so ``memcmp("ABC", "AB|", 3)`` which would say ``AB|`` is larger than ``ABC``, pipe being ascii 124 apparently

does that make sense?

jystic: Yeah you're right this is broken, although `entity_size` refers to `entity_ptr` so it would compare `AB` to `AB`.

amosr:
yep - looks great :european_castle:

PR #293, opened 2015-12-14
Duplicate functions are allowed as long as they have the same source

jystic:


jystic:
We need this unbreak the build /cc @tranma

src/Icicle/Source/Checker/Function.hs
```diff
@@ -37,13 +37,24 @@ checkFs :: Ord n
...
+    case List.lookup (snd name) env' of
+     Nothing
+      -> pure (env' <> [(snd name , (funtype, annotfun))])
+
+     Just (t, f)
+      | f'        <- reannotF (const ()) f
+      , annotfun' <- reannotF (const ()) annotfun
+      , t  == funtype
+      , f' == annotfun'
+      -> pure env'
```
tranma: will this consider alpha equivalence?

jystic: no I think it's just if the source is the same, it's just a `#pragma once`

amosr:
:+1:

PR #296, opened 2015-12-18
Pre-allocate all buffers using calloc so we can try large buffers (>= 1MB)

jystic:


amosr:
what's all this, then? the buffers would be too large for the stack otherwise?

jystic:
yeah that's right, I wanted to try 1MB buffers and it was causing segfaults

we may not need 1MB in the end, just trying to debug a performance issue with what looks like entities with a small number of facts (i.e. 1).

my thinking now is that we shouldn't flush the output buffer for each entity, but wait until it is full, that might solve the problem

amosr:
sounds good :+1:

amosr:
yes - I agree, I see no point to flush the output buffer for every entity, probably just when it's full

PR #297, opened 2015-12-20
Only flush output as often as required

jystic:


data/sea/52-psv.h
```diff
@@ -76,8 +80,8 @@ Constants
 */
 
 static const size_t psv_max_row_count      = 128;
-static const size_t psv_input_buffer_size  = 1024*1024;
-static const size_t psv_output_buffer_size = 1024*1024;
+static const size_t psv_input_buffer_size  = 256*1024;
+static const size_t psv_output_buffer_size = 256*1024;
```
tranma: magic! this is the best size?

tranma:
looks ok :fish:

data/sea/52-psv.h
jystic: haha i don't know, needs more experimentation!

it was the same as the previous 1MB buffer size for the big test I was doing, so I figure smaller is better where possible

PR #299, opened 2015-12-21
Off-by-one in mean fold

olorin:
Could be misreading (not sure how to actually run the thing), but

```
(v_, 1) : case s | (m, n) -> (m + (v_ - m) / (n + 1), n + 1)
```

Input: [5]
Initial case: (5, 1)
Output: 0 + (5 - 0) / (1 + 1), 1 + 1 = 2.5

/cc @jystic @amosr

### DIFF
--- a/data/libs/prelude.icicle
+++ b/data/libs/prelude.icicle
@@ -5,7 +5,7 @@ count v = fold c = 0 : seq v (c + 1) ~> c.
 
 # Numerically stable mean
 mean v = let v_ = double v
-      ~> fold1 s = (v_, 1) : case s | (m, n) -> (m + (v_ - m) / (n + 1), n + 1) end
+      ~> fold1 s = (v_, 1) : case s | (m, n) -> (m + (v_ - m) / n, n + 1) end
       ~> case s | (m, _) -> m end.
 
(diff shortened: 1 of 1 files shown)


tranma:
I think the current code is right, just need to start at 0 rather than 1.

tranma:
(see the online variance algorithm below this)

amosr:
ah. I think this is right, but I agree it's confusing. the ":" in the fold means "followed by" or "then", so it will only be executed for the second input (or first input if it were fold instead of fold1)

So in this case the value of the fold ``s`` is (5,1), then it just extracts the first part afterwards.
For a two-element input

```
[5, 2]

fold1 s = (_v, 1) : case s | (m, n) -> (m + (v_ - m) / (n + 1), n + 1)
==>
fold1 s = (5, 1) : case s | (m, n) -> (m + (v_ - m) / (n + 1), n + 1)
==>
fold1 s = case (5,1) | (m, n) -> (m + (v_ - m) / (n + 1), n + 1)
==>
fold1 s = (2 + (5 - 2) / (1 + 1), 1 + 1)
==>
fold1 s = (3.5, 2)
```

olorin:
Yep, figured out how to run stuff, current code is indeed correct. :+1:

PR #301, opened 2015-12-21
WIP: Updating Core: easier to implement Scan primitive

amosr:
WORST


amosr:
TODO:

1. isStreamWindowed in Stream/Stream
2. condenseStreams in Program/Condense
3. tracking bubblegum in Core evaluator
4. Distinct could be simplified to store (Map key ()) of whether given key has been seen and update accumulator in-place
5. Core->Avalanche conversion ignores KeepFactInHistory/bubblegum and windows
6. Core->Avalanche does not allow fold initialisers to refer to previous fold initialisers but should

amosr:
this is a hell of a PR. it seems to be passing most tests now, a few cil ones are broken but majority are fine. it is not condensing after fusion yet either, but I think that can be done separately afterwards

amosr:
tests are passing. works the same except for bubblegum, but will have some performance regressions. going to merge and fix separately

jystic:
I'm happy if you're happy :+1:

amosr:
I won't merge this yet after all, and will keep this branch separate for a little while longer

amosr:
OK. I think this is good enough to merge. just building & testing now

tranma:
@amosr if you TLDR what changed, we could skim the code a bit? Or just give proof by intimidation 😄

amosr:
yep, sounds good.
so the main change was in the core representation. instead of

```
Program
{ precomps :: [(Name, Exp)]
, streams    :: [(Name, Stream)]
, folds         :: [(Name, Fold)]
, postcomps :: [(Name, Exp)]
```

we have merged streams and folds together into a single "Stream" type:

```
Stream
 = SFold Name (Exp -- initialiser) (Exp -- accumulator expression)
 | SFilter (Exp -- predicate) [Stream] -- folds to perform only when filter is true
```

on top of that, the folds no longer have an explicit "input", but can just refer to the current values of all previous folds, as well as the input values.

as a result a program to compute the mean would look like..

```
Program
Stream (value : Int)
  count = Fold 0 (count + 1)
  sum   = Fold 0 (sum + value)
Returns
  repl = count / sum
```

since the folds and filter predicates can refer to values of previous folds, this is going to make implementing some sort of scan primitive a lot easier: something like "filter where running sum > 5" could be
```
Program
Stream (value : Int)
  sum   = Fold 0 (sum + value)
  filter (sum > 5)
    countFiltered = Fold 0 (countFiltered + 1)
Returns
  repl = countFiltered
```

so this change affects the conversion from Source a bit, but not too much really.
the conversion to Avalanche becomes a bit different.

the next difference is the Core evaluator, as well as that the Core evaluator no longer handles bubblegum correctly. the bubblegum can be fixed later, I think.

the biggest change was probably the Condense pass, which removes duplicate computations from a Core program.

tranma:
` | SFilter (Exp -- predicate) [Stream]`

the `[Stream]` can only be `SFold`s, right? No more `SFilter`s?

jystic:
Is the order of the folds important?

amosr:
no, there can be nested filters

amosr:
hmm. yes, the order of the folds are somewhat important because they can refer to previous folds. however, you cannot refer to later folds. so I suppose any order that typechecks (ie having folds defined before they are mentioned) will be equivalent

tranma:
oh! so permute the order of the Folds until it typechecks?

amosr:
why not! or find the topological ordering, if there is one

PR #302, opened 2016-01-04
convert buf values to explicit buf primitives

tranma:
...so that a chain of `buf_read (buf_push (buf_create, elem))` etc. could be a-normalised

### DIFF
--- a/src/Icicle/Avalanche/Prim/Eval.hs
+++ b/src/Icicle/Avalanche/Prim/Eval.hs
@@ -48,6 +48,13 @@ evalPrim p vs
       | otherwise
       -> primError
 
+
+     PrimBuf (PrimBufMake _ _)
+      | [VBase VUnit] <- vs
+      -> return . VBase . VBuf $ []
+      | otherwise
+      -> primError
+
      PrimBuf (PrimBufPush i _)
       | [VBase (VBuf as), VBase e] <- vs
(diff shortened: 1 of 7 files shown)


jystic:
Nice :8ball:

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -33,6 +36,56 @@ import qualified    Data.List as List
 import              Control.Monad.Trans.Class
 
 
+convertValues :: a -> Statement a n Prim -> Statement a n Prim
+convertValues a_fresh statements
+ = runIdentity
+ $ transformUDStmt trans () statements
+ where
+  trans _ s
```
amosr: I think there is a "transformX" combinator which would be ever-so-slightly easier

amosr:
:mans_shoe:

src/Icicle/Avalanche/Prim/Eval.hs
```diff
@@ -48,6 +48,13 @@ evalPrim p vs
       | otherwise
       -> primError
 
+
+     PrimBuf (PrimBufMake _ _)
+      | [] <- vs
```
amosr: ah - this should take a unit here

tranma:
should be ok now, tested locally

PR #304, opened 2016-01-05
Fix #294: Bug in substitution of function types; was not shadowing foralls

### LINKED ISSUE #294: Type checker bug
```
[fact.age]
  encoding = "int"
```

```
feature age
  ~> let top = (filter value >= 18 ~> filter value < 25 ~> count value)
  ~> top / count value
```
gives this type error:
```
Check error:
  Cannot discharge constraints at 54:108:test/cli/sortings/input/dictionary.toml
  Constraints:
    54:108:test/cli/sortings/input/dictionary.toml  Cannot unify: Possibly
                                                    With type:    Int
                                                    These types were required to be equal, but are not.
```

amosr:


jystic:
:nail_care:

PR #305, opened 2016-01-05
Topic/show inbuilt functions

HuwCampbell:


src/Icicle/Source/Query/Exp.hs
```diff
@@ -93,7 +93,7 @@ data Fun
  | DaysBetween
  | DaysEpoch
  | Seq
- deriving (Show, Eq, Ord)
+ deriving (Show, Eq, Ord, Enum, Bounded)
```
amosr: can you make a new function here, ``listOfAllPrims :: [Prim]``? I don't mind the enum/bounded trick but I'd rather it be very localised. that way if we decide to add more interesting, non-enumerable prims, we only need to modify this file

main/icicle-repl.hs
```diff
@@ -530,8 +531,8 @@ showState state
     , flag "c-eval:       " hasSeaEval
     ,      "now:          " <> T.unpack (renderTime $ currentTime state)
     ,      "data:         " <> show (length $ facts state) <> " rows"
-    ,      "dictionary:   " <> show (prettyDictionarySummary (dictionary state))
```
tranma: hm, why is this removed?

main/icicle-repl.hs
```diff
@@ -530,8 +531,8 @@ showState state
...
     ]
+ prettyHL $ prettyDictionarySummary (dictionary state)
```
tranma: oh, you just print it separately

tranma:
I like the fresh namer hack :+1:

src/Icicle/Source/Query/Exp.hs
HuwCampbell: Sure

amosr:
:+1: if you want to merge this in that'd be cool, if you don't, I'll get around to it after looking at the core rejig

PR #306, opened 2016-01-07
Upgrade mafia

jystic:


jystic:
Damn travis, why aren't the profiling libraries for base installed, how is that even possible??

jystic:
ugh, just about done with travis:
```
Disallowing packages: ghc-7.8.4-prof
If you require these packages, please review the package approval process at: https://github.com/travis-ci/apt-package-whitelist#package-approval-process
```

domdere:
nobody profiles

olorin:
I always used stack GHC with travis builds, seemed like the least-painful option.

jystic:
@olorin cool, got an example?

heh cabal+mafia+stack is *all* you need!

olorin:
@jystic https://github.com/anchor/oauth2-server/blob/master/.travis.yml

olorin:
(That's using stack for the build, but can just run `stack setup` and then munge the path before running `mafia`.)

jystic:
Yay, it works! /cc @tranma @amosr

tranma:
cool! I can just use mafia in this https://github.com/ambiata/icicle/pull/283, yeah?
:100: :100:

jystic:
@tranma yeah!

PR #307, opened 2016-01-07
Fix segfault in repl!

amosr:
Remove hardcoded stateWordsOfProgram function and replace with function to properly compute size of struct


jystic:
looks good :grapes:

PR #315, opened 2016-01-13
fix trailing comma in case of tombstone

tranma:
If you do a `latest 2 ~> value`, but one of the last 2 values is a tombstone, PSV output will not print it, but will print a comma. e.g. `"ID00000002|expression_latest_two|[c,]"`

this fixes the PSV output code so that the comma-outputting code is included inside the conditional.

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp.hs
+++ b/src/Icicle/Avalanche/Statement/Simp.hs
@@ -59,7 +59,9 @@ convertValues a_fresh statements
   goX xx
    = case xx of
        XValue _ (BufT n t) v
-        -> goP xx n t v
+        -> goB xx n t v
+       XValue _ (ArrayT t) v
+        -> goA xx t v
        XApp a x1 x2
         -> XApp a (goX x1) (goX x2)
        XLam a n t x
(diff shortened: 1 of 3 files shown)


tranma:
also de-savage-ify the quote-strings flag

tranma:
now looks like this `ID00000002|expression_latest_two|["c"]`

jystic:
looks legit :dragon_face:

tranma:
also did the conversion for A-normalisation that we did for Bufs, but this time for Arrays. Hope this is right...
/cc @amosr

amosr:
yup, this is good. less necessary than for bufs, but still a good idea.
you could simplify seaOfXValue now, as it doesn't have to deal with arrays or bufs.
:+1:

jystic:
@amosr we were getting some C that wasn't valid without this I think (for the same reason as with bufs)

amosr:
hmm ok.. my understanding is:
I thought the problem with bufs was that BUF(t) is a struct and the primitive functions require a pointer.
but ARRAY(t) is a pointer, and I believe all the primitives require a pointer too.

I'm not 100% convinced it will fix any C errors but I think it's a good thing regardless.

tranma:
@amosr this is the kind of invalid C we may get sometimes in quickcheck tests:

```
compute function #0 - windy:15:113: error: cannot take the address of an rvalue of type 'ibuf_1_ibool_t'
        iarray_ibuf_1_ibool_t simp43          = iarray_ibuf_1_ibool_put (mempool, simp42, 1, ibuf_1_ibool_push (&ibuf_1_ibool_make (), itrue)); /* let */
                                                                                                                ^~~~~~~~~~~~~~~~~~~~~
```

here ideally A-normalisation should have pulled out the applications into lets. I'm not exactly sure why that is happening, will investigate.

amosr:
hmm okay - this doesn't look like an array value though

tranma:
@amosr true, it's the buf value `ibuf_1_ibool_push (&ibuf_1_ibool_make (), itrue)` that wasn't being pulled out

PR #317, opened 2016-01-14
fix psv output bug

tranma:
fixes the current CI failure (was applying the struct transform function to the wrong thing in the OptionT case)


jystic:
:+1: on :green_heart:

tranma:
travis failure seems related to https://github.com/ambiata/icicle/issues/287. Still not sure what happened, but it's unrelated to this PR so merging.

Issue #260, opened 2015-11-27
no support for escaped strings in PSV input/output


tranma:
We currently don't deal with escapes at all....example:

```
marge|fresh|💀|1858-11-14
marge|fresh|{"green":"\""}|1858-11-16
```

```
line 2: terminator ',' or '}' not found: }
```

On the card for v1.

amosr:
using aeson to parse now, so it should be ok

PR #319, opened 2016-01-15
no shadowed C names

tranma:
Some generated C names were being shadowed, e.g. to output an array of array of bools

```
for (iint_t kermit_i = 0, kermit_n = p->kermit->count; kermit_i < kermit_n; ++kermit_i) {
  for (iint_t kermit_i = 0, kermit_n = iarray_iarray_ibool_index(p->kermit)->count; kermit_i < kermit_n; ++kermit_i) {
     (BOOM)
  }
}
```

This is likely what was causing the intermittent segfaults in https://github.com/ambiata/icicle/issues/287.

Fixed the issue by adding a "used names environment" and passing it around, pretty savage.

### DIFF
--- a/src/Icicle/Sea/FromAvalanche/Psv.hs
+++ b/src/Icicle/Sea/FromAvalanche/Psv.hs
@@ -40,6 +40,7 @@ import           Icicle.Sea.FromAvalanche.State
 import           Icicle.Sea.FromAvalanche.Type
 
 import           P
+import           Prelude (String)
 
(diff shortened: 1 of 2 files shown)


tranma:
we now run into https://github.com/ambiata/icicle/issues/260 every time, hopefully because all the bugs have been fixed :)

I'll just change WellTyped to not generate escaped strings until we support them.

jystic:
Looks legit :snowboarder:

PR #320, opened 2016-01-18
Support for GHC 7.10

jystic:
/cc @tranma

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -11,10 +11,7 @@ import           Distribution.Simple
 import           Distribution.Simple.Setup (BuildFlags(..), ReplFlags(..), TestFlags(..), fromFlag)
 import           Distribution.Simple.LocalBuildInfo
 import           Distribution.Simple.BuildPaths (autogenModulesDir)
-import           Distribution.Simple.Utils (createDirectoryIfMissingVerbose, rewriteFile)
-
-import           System.FilePath ((</>), (<.>))
-import           System.Process (readProcess)
+import           Distribution.Simple.Utils (createDirectoryIfMissingVerbose, rewriteFile, rawSystemStdout)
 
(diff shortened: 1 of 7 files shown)


tranma:
some nasty dependency errors! changes looks ok though :pig: when :green_apple:

Setup.hs
```diff
@@ -50,12 +60,19 @@ genBuildInfo verbosity pkg info = do
...
+timestamp :: IO String
+timestamp =
+  readProcess "date" ["+%Y%m%d%H%M%S"] "" >>= \s ->
+    case splitAt 14 s of
+      (d, n : []) ->
+        if (length d == 14 && filter isDigit d == d)
+          then return d
+          else fail $ "date has failed to produce the correct format [" <> s <> "]."
+      _ ->
+        fail $ "date has failed to produce a date long enough [" <> s <> "]."
```
tranma: what's wrong with using `formatTime`...? Is it slow?

jystic:
yeah it's a bug where cabal/ghc will try to link  `Setup.hs` against all the versions of `unix` available

See https://www.reddit.com/r/haskell/comments/3634x2/cabal_is_giving_a_weird_error_when_attempting_to/

The fix is to not using anything other than base & Cabal in Setup.hs files, or wait for Cabal 1.24

Setup.hs
jystic: Similar problem to multiple `unix` packages, if you configure with 7.8.4 then try to configure with 7.10.2 it will break

tranma: wow, so we can't even use `time` in `Setup.hs`. Savages

jystic: nope, too risky for the moment

cabal-1.24 adds support for Setup.hs dependency lists

PR #322, opened 2016-01-18
Add environment variable ICICLE_CC_OPTIONS

jystic:
This will allow us to have warnings as errors on our dev boxes (read: with clang), even though we want icicle to be more forgiving when used with gcc.

For example:

```shell
export ICICLE_CC_OPTIONS="-Werror -Weverything"
```

### DIFF
--- a/src/Icicle/Sea/Eval.hs
+++ b/src/Icicle/Sea/Eval.hs
@@ -28,7 +28,7 @@ module Icicle.Sea.Eval (
 
   , codeOfPrograms
   , assemblyOfPrograms
-  , compilerOptions
+  , getCompilerOptions
   ) where
 
 import           Control.Monad.Catch (MonadMask(..))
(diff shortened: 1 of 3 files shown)


tranma:
Should we set default options to `-Werror`, and just change the env variable when building on boxes? I'd like Werror to be on whenever possible...

Alternative detect if CC is GCC and print "lol" (savages).

jystic:
Hmm I'm really not sure, probably hard to build without warnings across all random C compilers, not sure though, I could be convinced

tranma:
Perhaps just `Werror` for clang? It's useful when running tests that generate C

jystic:
Yeah definitely, that's really why I added this in the first place, although I'm hoping we can use it for stuff like turning on ASAN or extra debug info which we've #defined out too

jystic:
For reference
```
ASAN_OPTIONS="symbolize=1:color=always" ICICLE_CC_OPTIONS="-fsanitize=address -fno-omit-frame-pointer" lldb ./dist/build/test/test
(lldb) run
... busy ...
... stops ...
(lldb) continue
```

tranma:
when ASAN option is used, actually run ice in lldb and send it commands?

amosr:
fantastic :+1:

Issue #327, opened 2016-01-21
Core conversion error


tranma:
```
=== prop_convert_is_well_typed from test/Icicle/Test/Source/Convert.hs:34 ===
*** Failed! Falsifiable (after 84 tests): 
QueryWithFeature {qwfQuery = Query {contexts = [Let () (Name (Variable "waldorf")) (Var () (Name (Variable "fozzy"))),Windowed () (Months 1) (Just (Weeks 1))], final = Prim () (Lit (LitString "kyle"))}, qwfNow = Just (Name (Variable "gonzo")), qwfOutput = OutputName "waldorf", qwfFeatureT = ArrayT BoolT, qwfFeatureN = Name (Variable "fozzy"), qwfTimeName = Name (Variable "waldorf")}
feature fozzy
~> let waldorf = fozzy
~> 
windowed between 1 weeks and 1 months
~> "kyle"
Program (core$0 : Stream Array Bool, core$1 : SNAPSHOT_TIME)
Precomputations:
  core$3               = "kyle"

Streams:
core$2               = sfold [(Array Bool, (Array Bool, Time))] (pair#
                                                                 (fst# core$0)
                                                                 core$0) then (pair#
                                                                               (fst#
                                                                                core$0)
                                                                               core$0)
sfilter window# 1 months1 weeks
        core$1 (fst# core$2)


Postcomputations


Returning:
  waldorf              = core$3

Left (ProgramErrorStream (StreamErrorExp (ExpErrorApp (XApp () (XPrim () (PrimWindow (Months 1) (Just (Weeks 1)))) (XVar () (NameMod (Variable "core") (Name (Variable "1"))))) (XApp () (XPrim () (PrimMinimal (PrimPair (PrimPairFst (ArrayT BoolT) (PairT (ArrayT BoolT) TimeT))))) (XVar () (NameMod (Variable "core") (Name (Variable "2"))))) (FunT [FunT [] TimeT] BoolT) (FunT [] (ArrayT BoolT)))))
```

putting up an issue so we don't forget

PR #329, opened 2016-01-25
WIP: fixing up the bubblegum

amosr:
I have added a new "FactIdentifierT" type that is really just a time. Latest push in Core takes a fact identifier and stores it in the buffer. The Core program also binds the current fact identifier, the current fact time, and the actual value (which is still the pair of date and time)

TODO still:
1. Core evaluation to read out the fact identifiers
2. Conversion to Avalanche should preserve the new fact identifier bindings

### DIFF
--- a/src/Icicle/Source/ToCore/ToCore.hs
+++ b/src/Icicle/Source/ToCore/ToCore.hs
@@ -385,7 +385,7 @@ convertQuery q
             let r = sfold n' pairt (pair (CE.emptyMap tkey T.UnitT) (foldZero res)) (beta kons)
 
             -- Perform a fold over that map
-            let p = post n'' $ xsnd $ CE.xVar n'
+            let p = post n'' $ beta (mapExtract res CE.@~ xsnd (CE.xVar n'))
 
             return (r <> p, n'')
 
--- a/src/Icicle/Source/ToCore/ToCore.hs
+++ b/src/Icicle/Source/ToCore/ToCore.hs
@@ -139,9 +139,10 @@ convertQuery q
      -> do  (bs, b) <- convertQuery q'
             now  <- convertDateName
             time <- convertFactTimeName
+            fact <- convertFactIdName
 
             let e'  = CE.makeApps () (CE.xPrim $ C.PrimWindow newerThan olderThan)
-                    [ CE.xVar now, CE.xVar time ]
+                    [ CE.xVar now, CE.xVar time, CE.xVar fact ]
             let bs' = filt e' (streams bs) <> bs { streams = [] }
             return (bs', b)
 
--- a/src/Icicle/Source/ToCore/ToCore.hs
+++ b/src/Icicle/Source/ToCore/ToCore.hs
@@ -406,12 +406,16 @@ convertQuery q
 
                 convertModifyFeaturesMap (Map.insert b (FeatureVariable (annResult $ annotOfExp def) xfst False) . Map.map (\fv -> fv { featureVariableExp = featureVariableExp fv . xsnd }))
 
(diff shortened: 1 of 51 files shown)


src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -132,19 +134,13 @@ convertQuery q
...
-            now <- convertDateName
-            (inp,_) <- convertInput
-
-            fs <- convertFeatures
-            let f = featureContextFactTime fs
-            fact <- case Map.lookup f (featureContextVariables fs) of
-                        Just f' -> return $ featureVariableExp f' $ CE.xVar inp
-                        Nothing -> convertError $ ConvertErrorExpNoSuchVariable (annAnnot ann) f
+            now  <- convertDateName
+            time <- convertFactTimeName
```
amosr: this fixes #327: it was trying to get the second of the "input stream", but the local input stream in that case was a let-binding returning a pair ((fact,time),True) or whatever instead of the global, original input.
now it will just use the fact time.

tranma: :+1:

tranma:
are fact identifiers just fact times? (and therefore all times must be unique?)

amosr:
yep, all times need to be unique (well, entity * attribute * time), so we can use the time as the fact identifier locally. I'd like to keep it relatively opaque though, if we can

markhibberd:
> yep, all times need to be unique (well, entity * attribute * time)

This isn't true, and I don't think it can be without another dimension.

tranma:
> yep, all times need to be unique (well, entity * attribute * time)
> This isn't true, and I don't think it can be without another dimension.

yeah this is what I fear, sorry @amosr

amosr:
hm ok - we are enforcing this restriction in Ice, I'm not sure where we got that idea from @jystic?

amosr:
in that case, I'll just represent it as an int index in Core and Avalanche, and then let the C output do what it wants (maybe just timestamp paired with an int index or something)

tranma:
@amosr the way we are enforcing input checks is not wrong, there are just other cases (input as events, etc.). We don't have one coherent story to cover all the cases yet, but anyway int index sounds good :+1:

test/Icicle/Test/Core/Arbitrary.hs
```diff
@@ -437,6 +446,8 @@ baseValueForType t
      -> VBool <$> arbitrary
     TimeT
      -> VTime <$> arbitrary
+    FactIdentifierT
+     -> VFactIdentifier . FactIdentifier <$> arbitrary
```
tranma: arbitrary fact identifiers work? would you run into duplicates often?

amosr: damn, good point. I guess this should fail, and perhaps FactIdentifierT shouldn't be generated in the type arbitrary either

amosr:
re fact identifiers: jacob and I decided last friday that we're just going to use indices for fact identifiers, and assume the entire entity's data will fit in memory (this is only necessary for latest type queries)

jystic:
> assume the entire entity's data will fit in memory

@markhibberd this should be ok right?

amosr:
this has a lot of changes, mainly to Core, but some changes to conversion from Source and conversion to Avalanche.
The Core evaluator now keeps track of bubblegum, but only the "historical" kind: latests and windows.
I have had to ban "latest inside latest" for now. I think this is a relatively minor problem.
I would like to merge this before starting on the next part - Avalanche bubblegum.

jystic:
@amosr want to walk us through it when you get a minute?

jystic:
:smoking: :yen: :dollar: :pound: :euro:

PR #330, opened 2016-01-27
output dense psv

tranma:
assuming output is sorted by entities (pretty sure it is?)

```
marge|f|0.0
maggie|f|NA
lisa|m|NA
bart|m|NA
homer|m|75466.66666666667
```

maybe we need a header row for attribute names too


jystic:
This looks great, thanks for splitting the modules up, can't believe how simple the dense output function ended up being, it's beautiful :8ball:

tranma:
merging this, output dictionary will be in a separate PR for easier reviewing

Issue #331, opened 2016-02-01
weird control characters following chord output


tranma:
just putting this up as an issue so I don't forget. Chord output has these control characters at the end of each line:

```
ID00000005|sauerkraut_grapefruits|-1757910297|2014-06-03T00:00:00^@^@
```

PR #333, opened 2016-02-02
fix chord output bug

tranma:
@jystic  fix https://github.com/ambiata/icicle/issues/331

also piggybacking a change to use custom missing value in dense output


jystic:
Looks good :+1:

PR #334, opened 2016-02-03
really fix chord_time

tranma:
damn strings


jystic:
nobody knows strings like the icicle team :snake:

PR #335, opened 2016-02-12
dense psv input

tranma:
Parse dense PSV (snapshots, feeds) as input. e.g. this sample data feed:

```
homer|0|CA|1989-12-17
lisa|5|CA|1989-12-18
lisa|9|UT|1991-12-19
marge|8|CA|1989-12-17
marge|9|UT|1991-12-19
```

This input will be parsed as a single struct `(misery:int,state:string)`. The format is specified in the TOML dictionary:

```
[feed.person]
  columns = ["misery", "state"]

[fact.person]
  encoding = "(misery:int,state:string)"

[feature.mean_misery_by_state]
  expression = "feature person ~> group state ~> mean misery"
```

Sample output:

```
marge|[["CA",8.0],["UT",9.0]]
homer|[["CA",0.0]]
lisa|[["CA",5.0],["UT",9.0]]
```

The current caveat is that we can't have a struct as one of the dense fields, e.g. this is an error:

```
lisa|5|{"location":"head", "serverity":2}|1992-01-01
```

This is because we require structs to have primitive encodings. It should be changed eventually, at least when we are processing dense input.

@markhibberd @HuwCampbell @jystic

### DIFF
--- a/data/sea/52-psv.h
+++ b/data/sea/52-psv.h
@@ -65,7 +65,8 @@ static ierror_msg_t INLINE psv_write_outputs
   , size_t entity_size
   , ifleet_t *fleet );
 
-static ierror_loc_t INLINE psv_read_fact_sparse
+#if ICICLE_PSV_INPUT_SPARSE
+static ierror_loc_t INLINE psv_read_fact
   ( const char   *attrib_ptr
   , const size_t  attrib_size
   , const char   *value_ptr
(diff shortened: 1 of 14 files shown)


data/sea/52-psv.h
```diff
@@ -152,6 +161,7 @@ static ierror_loc_t psv_read_buffer (psv_state_t *s)
             error = ierror_loc_format (attrib_ptr, n_ptr, "missing '|' after attribute");
             goto on_error;
         }
+#endif
```
jystic: monster :)

tranma: keeping with fine tradition

jystic:
This is nice :+1:

PR #339, opened 2016-02-23
optional dense fields

tranma:
@HuwCampbell should fix https://github.com/ambiata/data-science-support/issues/231

If no missing value is specified, `""` is automatically used.

example:

input:

```
title = "toy"
namespace = "story"
[fact.stuff]
  encoding = "(age:int*,name:string)"
[feed.stuff]
  columns = ["age", "name"]
[feature.severity]
  expression = "feature stuff ~> filter isSome age ~> mean (getOrElse 0 age)"
```

```
woody|1|sheriffwoody|1991-01-01
buzz|NA|buzzlightyear|1991-01-01
```

run:

```
./dist/build/Ice/ice snapshot foo.psv foo.toml 2015-01-01 --input-psv dense --output-psv sparse           
woody|severity|1.0
```

### DIFF
--- a/src/Icicle/Data.hs
+++ b/src/Icicle/Data.hs
@@ -167,4 +167,4 @@ attributeOfStructField (StructField _ attr _)
 data StructFieldType =
     Mandatory
   | Optional
-  deriving (Eq, Show)
+  deriving (Eq, Ord, Show)
--- a/src/Icicle/Sea/Psv/Input.hs
+++ b/src/Icicle/Sea/Psv/Input.hs
@@ -154,6 +154,8 @@ seaOfDenseFieldMapping m (FieldMapping fname ftype vars) = do
   let sea   = wrapInBlock
             $ vsep [ "char *ent_pe = pe;"
                    , "char *pe     = memchr(p, '|', ent_pe - p);"
+                   , "if (pe == NULL)"
+                   , "    pe = ent_pe;"
                    , ""
                    , fieldSea ]
 
@@ -406,11 +408,10 @@ seaOfReadDenseInput missing inType inVars
                                    , val_sea ]
                  body_false = indent 4
                             $ assignVar (pretty nb) BoolT "false"
-                 strncmp s  = "strncmp(p, \"" <> pretty s <> "\", pe - p)"
              case missing of
                Just m
                  -> pure $ vsep
-                     [ "if (" <> strncmp m <> " == 0) {"
+                     [ "if (" <> seaOfStringEq m "p" (Just "pe - p") <> ") {"
                      , "    p = pe;"
                      , body_false
                      , "} else {"
--- a/src/Icicle/Sea/Psv/Input.hs
+++ b/src/Icicle/Sea/Psv/Input.hs
@@ -29,9 +29,9 @@ import           Icicle.Avalanche.Prim.Flat (meltType)
 import           Icicle.Common.Type (ValType(..), StructType(..), StructField(..))
 import           Icicle.Common.Type (defaultOfType)
 
-import           Icicle.Data (Attribute(..))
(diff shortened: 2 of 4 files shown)


HuwCampbell:
shallow review :jeans:
Can you add some CLI tests, or does the REPL not take dense? Also, where above was the NA specified?

tranma:
@HuwCampbell there is no way to switch to dense psv from the REPL, but the quickcheck tests do test dense psv, so I'll make them generate some optional fields too.

For examples, I'll add some to the Ice CLI tests and readme.

tranma:
@HuwCampbell sorry, the above dictionary should have this

```
[feed.stuff]
  missing = "NA"
  columns = ["age", "name"]
```

tranma:
@jystic C code eyeballing? Generated C for the above input looks like:

```C
#line 1 "read fact #0 - stuff"
static ierror_loc_t INLINE psv_read_fact_0 (const char *value_ptr, const size_t value_size, itime_t time, imempool_t *mempool, iint_t chord_count, iprogram_0_t *programs)
{
    ierror_loc_t error;

    char *p  = (char *) value_ptr;
    char *pe = (char *) value_ptr + value_size;

    ierror_t new_severity_conv_0_simp_190;
    ibool_t new_severity_conv_0_simp_191 = ifalse;
    iint_t new_severity_conv_0_simp_192 = 0;
    istring_t new_severity_conv_0_simp_193 = "";

    {
        new_severity_conv_0_simp_190 = ierror_not_an_error;

        for (;;) {
            {
                /* age */
                {
                    char *ent_pe = pe;
                    char *pe     = p;
                    while (pe != ent_pe && *pe != '|')
                        pe++;

                    if (strncmp(p, "NA", pe - p) == 0) {
                        p = pe;
                        new_severity_conv_0_simp_191 = false;
                    } else {
                        new_severity_conv_0_simp_191 = true;
                        iint_t value;
                        error = text_read_iint (&p, pe, &value);
                        if (error) return error;
                        new_severity_conv_0_simp_192 = value;
                    }
                }

                char *term_ptr = p++;
                if (*term_ptr != '|')
                    return ierror_loc_format (p-1, p-1, "expect field separator '|'");

                if (term_ptr == pe)
                    break;

            }
            {
                /* name */
                {
                    char *ent_pe = pe;
                    char *pe     = p;
                    while (pe != ent_pe && *pe != '|')
                        pe++;

                    istring_t value;
                    error = text_read_istring (mempool, &p, pe, &value);
                    if (error) return error;
                    new_severity_conv_0_simp_193 = value;
                }

                char *term_ptr = p++;
                if (*term_ptr != '|')
                    return ierror_loc_format (p-1, p-1, "expect field separator '|'");

                if (term_ptr == pe)
                    break;

            }
            return ierror_loc_format (p-1, p-1, "invalid dense field start");
        }
    }
  ...
}
```

jystic:
```c
                    while (pe != ent_pe && *pe != '|')
                        pe++;
```

Curious why we need this?

tranma:
@jystic `pe` points to the end of the whole dense value (right before the `time`), but for each dense field I want `pe` to be the end of the field.

jystic:
```c
                    if (strncmp(p, "NA", pe - p) == 0) {
```

This should probably be using the `StringWord` for consistency with the rest of our string equality, also avoids a call in to libc. See here for [example](https://github.com/ambiata/icicle/blob/01248c322c06b8d80983cbac87a2323b40d1a8b6/src/Icicle/Sea/Psv/Input.hs#L570)

jystic:
> @jystic pe points to the end of the whole dense value (right before the time), but for each dense field I want pe to be the end of the field.

I guess I'm wondering about the loop? You could use 'memchr' instead

jystic:
Looks good to me :+1:

PR #340, opened 2016-02-24
Add lock files

jystic:
This upgrades mafia and adds lock files for 7.8.4 and 7.10.2

Lets see how this works out! :)

/cc @tranma


tranma:
this includes all transitive dependencies, yeah? didn't expect statistics to be there :+1:

jystic:
yeah that is pretty surprising!

jystic:
@tranma 
```
$ ./mafia depends statistics --tree
criterion-1.1.0.0
 └─╼ statistics-0.13.2.3
```

PR #344, opened 2016-03-01
improve compilation time

tranma:
Just tried the lowest hanging fruit: add a hashcode the names.

On our example dictionary that was taking >30m to compile, this now takes ~15m. There is still improvements to be made but this is a start.

@jystic @markhibberd


amosr:
:money_with_wings: 
what do you think about a convenience-typeclass called Namelike or something?

```
class (Pretty n, Eq n, Hashable n) => Namelike n where
   nameOfString :: String -> n
   stringOfName :: n -> String
```

do you think we gain much from having the name polymorphic? I quite like it, but it does have some annoying repercussions

tranma:
@amosr I actually thought the polymorphic name was a bit annoying :)

amosr:
no problem - we can get rid of it whenever we're brave enough, then

jystic:
> @amosr I actually thought the polymorphic name was a bit annoying :)

I like the polymorphic name, keeps us honest :D

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -331,10 +334,11 @@ substXinS a_fresh name payload statements
 --  * Remove some other useless code:
 --      statements that do not update accumulators or return a value are silly.
 --
-thresher :: (Ord n, Eq p) => a -> Statement a n p -> FixT (Fresh n) (Statement a n p)
+thresher :: (Hashable n, Eq n, Eq p) => a -> Statement a n p -> FixT (Fresh n) (Statement a n p)
 thresher a_fresh statements
- = transformUDStmt trans emptyExpEnv statements
+ = fmap (reannotS fst) $ transformUDStmt trans emptyExpEnv $ freevarsStmt statements
```
tranma: @amosr is this ok? (for thresher, I only compute the freevars once)

tranma:
after a few savage "improvements", we can compile a dictionary with ~160 columns in 12m (previously >30m)

src/Icicle/Avalanche/Statement/Simp.hs
amosr: looks groovy. did it help with speed much?

tranma: yeah, about 30% less time/alloc. It could be better if I could minimise finding freevars in other heavy transforms such as nestBlocks as well.

Curiously now `foldStmt` is quite expensive, surely walking the Avalanche AST can't be that bad....

amosr: actually, I guess freevars could probably be replaced by allvars in nestBlocks, since it's just going to be a bit more pessimistic in what it needs to rename (can't remember if you tried that)

tranma: I tried exactly that and made it infinitely worse :(

jystic:
:rage1: :rage2: :rage3: :rage4: :goberserk: on :green_apple:

tranma:
weird, the verbose times info didn't match real time

```
time ./dist/build/ice/ice snapshot empty.psv dictionary-commercial.toml   2016-01-01 --input-psv dense -v
ice: detected 8 processors
ice: compiling dictionary-commercial.toml
ice: compilation time = 519.48s
.....(no data so everything else took 0s)....
./dist/build/ice/ice snapshot empty.psv dictionary-commercial.toml 2016-01-01  964.78s user 521.43s system 286% cpu 8:39.65 total
```

PR #347, opened 2016-03-07
Fix bug in distinct: make Option_fold explicitly delay the 'none' cas…

amosr:
Fix the bug, however we one day need a better analysis on avalanche/flatten to see which map/array operations can be made mutable updates, and which need to copy beforehand.


tranma:
I think for that we'll need something like linear types, not ad-hoc analysis to see what can be mutable, but that's a discussion for another time...

this looks good :dart:

amosr:
yeah, true. I suspect given the very limited nature we could do a pretty good job with an analysis though

PR #348, opened 2016-03-09
melt relation primops

tranma:
....so we can compare things like `OptionT IntT`

might need a REPL test to demonstrate.
@amosr @jystic

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
+++ b/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
@@ -252,26 +252,14 @@ constructor a_fresh statements
    $ fmap (\(t, ix) -> primArrayPut aix t (primUnpack ix (ArrayT tx) na)
                                           (primUnpack ix         tx  nv)) tis
 
-   -- equality
-   | (PrimMinimal (Min.PrimRelation Min.PrimRelationEq t), [nx, ny]) <- prima
-   , Just tis <- withIndex tryMeltType t
-   = Just
-   $ primFold1 primAnd
-   $ fmap (withPrim primEq t nx ny) tis
-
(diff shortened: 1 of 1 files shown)


src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -234,35 +242,75 @@ constructor a_fresh statements
...
 
    | (PrimMinimal (Min.PrimRelation Min.PrimRelationNe t), [nx, ny]) <- prima
-   , Just ts <- tryMeltType t
-   , tis     <- List.zip ts [0..]
-   = Just $ primFold1 primOr
-   $ fmap (\(tv, i) -> primNe tv (primUnpack i t nx) (primUnpack i t ny)) tis
+   , Just tis <- withIndex tryMeltType t
+   = Just
+   $ primFold1 primOr
+   $ fmap (withPrim primNe t nx ny) tis
```
jystic: Would this be the same as using the generalised version below for Eq/Ne ?

tranma: yea I guess it would be

jystic: I think we should do that then :)

tranma: we can do that for Eq but not Ne

jystic: Not sure if it's better, but I guess we could have a function for Op -> AppendOp

tranma:
you think this is ok, @amosr?

jystic:
Looks great :space_invader: 

Now our comparison ops are superior to C! :cake:

amosr:
woohoo!

PR #350, opened 2016-03-11
Separate built-in funs as primitives

tranma:
* Add a `PrimBuiltinFun` primitive to `Minimal`, which includes built-in maths functions, etc.
* Add `keys`, `vals` primitive to get the keys/vals out of the map
* Fix pretty printer to separate argument types with comma

@amosr @jystic


src/Icicle/Common/Exp/Prim/Minimal.hs
```diff
@@ -47,7 +46,7 @@ data Prim
 data PrimArithUnary
  = PrimArithNegate
  | PrimArithAbsolute
- deriving (Eq, Ord, Show)
+ deriving (Eq, Ord, Show, Enum, Bounded)
```
amosr: what are these for?

tranma: to be used for `arbitraryBounedEnum`

amosr:
:octopus:

src/Icicle/Common/Exp/Prim/Minimal.hs
amosr: oh, nice

PR #351, opened 2016-03-13
Avalanche Bubblegum: split out Flatten, make KeepFactInHistory take fact id 

amosr:


amosr:
Next step is to traverse all the accumulators after ForeachFacts, and find all nested buffers of facthistory, and call KeepFactInHistory on them

amosr:
@tranma mind having a look at this?

I split up Flatten and modified it to convert types "Buf n t" to "(Buf n FactIdentifier, Buf n t)".

KeepFactInHistory also takes an expression for the FactIdentifier.

tranma:
this makes sense :+1: golden finger

PR #355, opened 2016-03-14
Faster!

amosr:
With 80 fields:

```
amos@Amos icicle $ time dist/build/icicle-repl/icicle-repl ":dictionary bigdict.toml" "feature big ~> group s0 ~> newest s1" < /dev/nullwelcome to iREPL
ok, loaded dictionary with 2 features and 24 functions
- Core evaluation:
[]

>
real	0m12.959s
user	0m12.797s
sys	0m0.156s
```

while this was around 50s before. Yay!

### DIFF
--- a/src/Icicle/Avalanche/FromCore.hs
+++ b/src/Icicle/Avalanche/FromCore.hs
@@ -11,6 +11,7 @@ module Icicle.Avalanche.FromCore (
 import              Icicle.Common.Base
 import              Icicle.Common.Type
 import qualified    Icicle.Common.Exp as X
+import              Icicle.Common.Fresh
 
 import              Icicle.Core.Exp.Prim
 
@@ -23,8 +24,8 @@ import qualified    Icicle.Core.Stream          as CS
 import              P
 import              Data.String
 import qualified    Data.Map as Map
+import qualified    Data.Set as Set
 
-import              Data.Functor.Identity
 import              Data.Hashable (Hashable)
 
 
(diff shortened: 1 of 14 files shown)


tranma:
try a pair aggregate, and group-fold, like this: `group fold (x,y) = (group foo ~> newest bar, mean bar) ~> ...`

I'll try on my big dictionary!

src/Icicle/Avalanche/FromCore.hs
```diff
@@ -75,40 +80,19 @@ programFromCore namer p
...
-   = factLoop FactLoopNew
-
-  factLoop loopType
-   = ForeachFacts (FactBinds (C.factTimeName p) (C.factIdName p) [(C.factValName p, inputType')]) inputType' loopType
-   $ Block factStmts
-
-  inputType' = PairT (C.inputType p) TimeT
-
-  (accNames, factStmts)
-   = makeStatements p namer (C.streams p)
```
tranma: what are these changes for, doesn't look like they would make a difference to the simp steps? /lazy

src/Icicle/Avalanche/Simp.hs
```diff
@@ -65,17 +66,28 @@ simpFlattened a_fresh p
...
+   -- Doing this straight away means a smaller program for later passes.
+   >>= return .  dead
    -- Perform let-forwarding on statements, so that constant lets become free
    >>= forwardStmts a_fresh
    -- Try to evaluate any exposed primitives
    >>= transformX return (return . simpEvalX Flat.evalPrim Flat.typeOfPrim)
    -- Pull Let statements out of blocks. This just allows thresher to remove more duplicates
    >>= lift . nestBlocks   a_fresh
-   -- Thresh again. Surprisingly, having both threshers makes simpFlattened twice as fast!
+   -- Thresh to remove duplicate expressions
```
tranma: the order of doing these things makes sense now

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -589,33 +573,3 @@ nestBlocks a_fresh statements
...
-   | Write n _ <- s
-   , n == acc
-   = return (AccumulatorUsage True False)
-
-   | Read _ n _ _ <- s
-   , n == acc
-   = return (ors r (AccumulatorUsage False True))
-
-   | otherwise
-   = return r
```
tranma: this isn't useful anymore?

tranma: oh right, I see. Dead code is already dealing with it, nevermind me :)

tranma:
my dictionary goes from >40m to ~22s
:100: :100: :100:

jystic:
:godmode:

jystic:
Exponential speed ups!

src/Icicle/Avalanche/FromCore.hs
amosr: it was sort of a combined bug in conversion to Avalanche and Dead, to do with shadowing. we decided to keep Dead simple and modify the conversion to have no shadowing.

conversion to Avalanche was producing something like

```
read x = x$acc
write x$acc = 0
read x = x$acc
write x$acc = x + 1
```

where the first "x" is dead, but Dead was confused by the shadowing.

tranma: ah that's great. I :+1: no-shadowing

tranma:
disregard the fails (this was passing before CI rolled), can LGTM :poultry_leg:

PR #357, opened 2016-03-16
Add cli test for windows, just for fun

amosr:


tranma:
:rocket: :rooster: :new_moon_with_face:

PR #359, opened 2016-03-17
Bubblegum: save fact identifiers used by latest buffers

amosr:
This change affects flatten.
Flatten keeps track of all the accumulators in scope, and after the ForeachFacts loop, digs down into all the accumulators and pulls out any FactIdentifiers inside buffers.
We might want to add an option to disable this, since these loops are not necessary for Ice.

### DIFF
--- a/src/Icicle/Avalanche/Prim/Compounds.hs
+++ b/src/Icicle/Avalanche/Prim/Compounds.hs
@@ -41,6 +41,9 @@ data FlatOps a n = FlatOps {
   , sndF     :: ValType -> ValType -> X a n -> X a n
   , pair     :: ValType -> ValType -> X a n -> X a n -> X a n
 
+  , structGet:: StructField -> ValType -> StructType
+             -> X a n -> X a n
+
   , relEq    :: ValType -> X a n -> X a n -> X a n
 
   }
@@ -81,6 +84,9 @@ flatOps a_fresh
   fstF   t u    = p1 (PrimMinimal $ Min.PrimPair $ Min.PrimPairFst t u)
   sndF   t u    = p1 (PrimMinimal $ Min.PrimPair $ Min.PrimPairSnd t u)
   pair   t u    = p2 (PrimMinimal $ Min.PrimConst $ Min.PrimConstPair t u)
+
+  structGet f t st
+                = p1 (PrimMinimal $ Min.PrimStruct $ Min.PrimStructGet f t st)
 
   relEq     t   = p2 (PrimMinimal $ Min.PrimRelation Min.PrimRelationEq t)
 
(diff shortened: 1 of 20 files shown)


amosr:
and just for fun, I made takeApps a bit better:

original:
```
> feature big ~> group s0 ~> newest s1 (struct with 160 fields)
real	0m29.063s
user	0m28.562s
sys	0m0.488s
```

new:
```
real	0m21.339s
user	0m20.867s
sys	0m0.381s
```
(!!!)

src/Icicle/Avalanche/Statement/Flatten/Save.hs
```diff
@@ -0,0 +1,150 @@
...
+
+  fpIsSome    t = xPrim (Flat.PrimProject (Flat.PrimProjectOptionIsSome t))
+  fpOptionGet t = xPrim (Flat.PrimUnsafe (Flat.PrimUnsafeOptionGet t))
+
+  fpIsRight  t u = xPrim (Flat.PrimProject (Flat.PrimProjectSumIsRight t u))
+  fpSumLeft  t u = xPrim (Flat.PrimUnsafe (Flat.PrimUnsafeSumGetLeft  t u))
+  fpSumRight t u = xPrim (Flat.PrimUnsafe (Flat.PrimUnsafeSumGetRight t u))
+
+  fpFst t u     = xPrim (Flat.PrimMinimal $ Min.PrimPair $ Min.PrimPairFst t u)
+  fpSnd t u     = xPrim (Flat.PrimMinimal $ Min.PrimPair $ Min.PrimPairSnd t u)
```
tranma: I've seen these functions elsewhere I think, perhaps it'd be good to make a `Compounds` module for them at some point?

tranma:
makes sense :+1: :sun_with_face:

src/Icicle/Avalanche/Statement/Flatten/Save.hs
amosr: probably should...

src/Icicle/Avalanche/Prim/Compounds.hs
```diff
@@ -0,0 +1,92 @@
...
+  , structGet:: StructField -> ValType -> StructType
+             -> X a n -> X a n
+
+  , relEq    :: ValType -> X a n -> X a n -> X a n
+
+  }
+
+flatOps :: a -> FlatOps a n
+flatOps a_fresh
+ = FlatOps{..}
```
amosr: record wildcard probably not necessary here, and could be written just as cleanly without it I guess

src/Icicle/Avalanche/Statement/Flatten/Exp.hs
```diff
@@ -63,12 +65,14 @@ flatX a_fresh xx stm
...
 
   -- Annotation plumbing
-  flatX'    = flatX    a_fresh
-  makeApps' = makeApps a_fresh
-  xVar      = XVar     a_fresh
-  xPrim     = XPrim    a_fresh
-  xValue    = XValue   a_fresh
-  xApp      = XApp     a_fresh
+  flatX'           = flatX        a_fresh
+  Flat.FlatOps{..} = Flat.flatOps a_fresh
```
amosr: I guess record wildcards only save a single argument here, but I do think it's a bit cleaner

PR #360, opened 2016-03-18
Standardise the prelude

tranma:
* group_values is gone because it's silly (higher order functions!!!)
* snake case
* `return` which just promotes `Pure` to `Aggregate` at the query top level

@amosr

### DIFF
--- a/data/libs/prelude.icicle
+++ b/data/libs/prelude.icicle
@@ -2,23 +2,37 @@
 ### Base ###
 ############
 
+# mod, relying on rounding.
+mod x y
+  = x - (trunc (double x / double y) * y).
+
 # Time statements
 newest v = latest 1 ~> fold1 s = v : v ~> s.
 oldest v = fold1 s = v : s ~> s.
 
+# check if a day is a weekend.
+is_weekend v
+  =  let day = mod (days v) 7
+  ~> day == 3 || day == 4.
+
+# starting at epoch 1600/01/03, identify the week that containing this day.
+week_of_day d
+  = trunc (double (days d) / 7).
+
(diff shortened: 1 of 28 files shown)


data/libs/prelude.icicle
```diff
@@ -96,3 +111,15 @@ numflips' v t
...
+### Backwards Compatibility ###
+###############################
+
+isSome a = is_some a.
+isNone a = is_none a.
+
+getOrElse d a = get_or_else d a.
+
+minby a b = min_by a b.
+maxby a b = max_by a b.
```
amosr: enterprise :necktie:

markhibberd:
Would be good to discuss how prelude should be versioned. In an ideal world I feel something like the following would be nice:
 - Prelude is explicit import (or at least can be disabled).
 - Prelude is always kept versioned, prelude-1.0.icicle, prelude-2.0.icicle, prelude-2.1.icicle.....
 - It is possible to, and by standard internally we always, import an explicitly versioned prelude
 - There is some sort of convenience mode that grabs the latest one, but would be primarily for demonstration/educational purposes.

amosr:
that sounds doable (and sensible).
it would also mean changes to the prelude wouldn't affect all the tests (which are most of the changes here), if we pinned the majority of the test cases to prelude-0.

tranma:
Should we publish the prelude as a separate thing from ice/icicle-repl, that can be installed with payload or appshop?

amosr:
I'm not sure. I'm tempted to keep them in Icicle, in case future versions of the prelude depend on newer version of Icicle?

PR #361, opened 2016-03-21
Optimise a-normal pass (among a few others)

amosr:


src/Icicle/Avalanche/Simp.hs
```diff
@@ -88,7 +88,7 @@ simpFlattened a_fresh p
    -- Pull Let statements out of blocks. This just allows thresher to remove more duplicates
    >>= lift . nestBlocks   a_fresh
    -- Thresh to remove duplicate expressions
-   >>= thresher     a_fresh
+   >>= thresherNoAlpha     a_fresh
```
tranma: you don't need to check duplicate lets with alpha eq after nestBlocks?

tranma:
just by looking at the number of generated variable names, it seems much better :alien:

src/Icicle/Avalanche/Simp.hs
amosr: luckily not: because lets are non-recursive, the only need for alpha checks are before flattening, where lambdas are used

PR #363, opened 2016-03-22
Update tests after optimisations

tranma:
@amosr this fixes le build


amosr:
oh! whoops. I don't know how I missed that. :+1:

PR #365, opened 2016-03-24
Upgrade Disorder

amosr:
```
$ DISORDER_RUN_FEWER=100 DISORDER_RUN_NORMAL=500 DISORDER_RUN_MORE=10000 ./mafia test
```

It also found a bug in Source.ToCore so I fixed that. Lol.

### DIFF
--- a/test/Icicle/Test/Source/Convert.hs
+++ b/test/Icicle/Test/Source/Convert.hs
@@ -5,6 +5,7 @@
 {-# OPTIONS_GHC -fno-warn-missing-signatures #-}
 module Icicle.Test.Source.Convert where
 
+import           Icicle.Test.Arbitrary
 import           Icicle.Internal.Pretty
 import qualified Icicle.Core.Program.Check      as CCheck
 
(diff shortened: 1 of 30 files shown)


tranma:
should have a... `DISORDER_RUN_AS_GOOD_AS_VERIFIED`

tmcgilchrist:
:wine_glass:

test/Icicle/Test/Source/Convert.hs
```diff
@@ -5,6 +5,7 @@
 {-# OPTIONS_GHC -fno-warn-missing-signatures #-}
 module Icicle.Test.Source.Convert where
 
+import           Icicle.Test.Arbitrary
```
amosr: I got a *really* surprising/confusing error message when I mentioned "$checkAllWith" but hadn't imported this. It was something along the lines of
```
GHC staging restriction:
cannot mention checkAllWith in a splice because it is defined in the same module; you should move it out to a separate module
```

tranma: yeah I don't know why it thinks just using an undefined splice is the same as "defining" it. This isn't Javascript....

amosr: error message. the first one is fine, but the second one is odd

```
test/Icicle/Test/Source/Convert.hs:51:9:
    Not in scope: ‘checkAllWith’
    In the splice: $checkAllWith

test/Icicle/Test/Source/Convert.hs:51:9:
    GHC stage restriction:
      ‘checkAllWith’ is used in a top-level splice or annotation,
      and must be imported, not defined locally
    In the expression: checkAllWith
    In the splice: $checkAllWith
```

amosr: oh well. now we know

PR #367, opened 2016-03-29
Big data mode: ingest invariants

amosr:
This adds the big data mode to check that every fold expression is encased in a window or latest.
This was easier than I expected.

### DIFF
--- a/src/Icicle/Source/Checker/Invariants.hs
+++ b/src/Icicle/Source/Checker/Invariants.hs
@@ -106,14 +106,14 @@ invariantX
         -> Result   a n
 invariantX ctx x
  = case x of
-    Var _ n
-     -> goFun n []
+    Var a n
+     -> goFun a n []
     Nested _ q
      -> invariantQ ctx q
     App{}
      | (f,xs) <- takeApps x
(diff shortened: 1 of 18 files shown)


tranma:
looks correct :moneybag:

src/Icicle/Source/Checker/Invariants.hs
```diff
@@ -122,11 +122,11 @@ invariantX ctx x
      -> invariantX ctx s >> mapM_ (invariantX ctx . snd) ps
 
  where
-  goFun n args
+  goFun a n args
    | Just fun <- Map.lookup n $ checkBodies ctx
    = let ctx' = foldl bindArg ctx (arguments fun `zip` args)
-     -- TODO: wrap error in function information
-     in  invariantQ ctx' $ body fun
+     in  errorInFunctionEither a n
```
amosr: The idea is right, but the rest is wrong: it turns out that "checkBodies" is always empty, and the only reason I/we *thought* this was working, was because invariant check is also run after inlining.

amosr:
I have no tests yet, but I think that is the next step

PR #368, opened 2016-03-30
Add primitive box :: OptionT a -> Possibly a

tranma:
now we can do

```
feature injury ~> days between (newest (box admitted)) now
```

rather than `feature injury ~> case admitted | None -> .... | Some a -> ...`

### DIFF
--- a/src/Icicle/Source/ToCore/Prim.hs
+++ b/src/Icicle/Source/ToCore/Prim.hs
@@ -212,6 +212,13 @@ convertPrim p ann resT xts = go p
    | otherwise
    = convertError
    $ ConvertErrorPrimNoArguments ann 2 p
+  -- Similarly this is not a tombstone.
+  godata Box
+   | [(x,_)] <- xts
+   = return x
+   | otherwise
+   = convertError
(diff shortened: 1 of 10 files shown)


src/Icicle/Source/ToCore/Prim.hs
```diff
@@ -212,6 +212,13 @@ convertPrim p ann resT xts = go p
    | otherwise
    = convertError
    $ ConvertErrorPrimNoArguments ann 2 p
+  -- Similarly this is not a tombstone.
+  godata Box
```
jystic: `godata` sounds like some scary golang keyword

tranma: below is `gomath` and `gomap`

jystic:
Look good to me :+1:

src/Icicle/Source/ToCore/Prim.hs
jystic: `gomath` sounds like you're gunna have a bad time

amosr:
fantastic :facepunch:

PR #370, opened 2016-03-30
more interesting example queries

amosr:


jystic:
:rabbit2:

tranma:
the second one ("Check if maximum of date is after now") is a low-hanging fruit we can probably tackle soon. I'll add more as I see them.

amosr:
Yeah - when you do, I'd like to make it an option to the typechecker. Just in case we ever do want a "resumable with no refutations", that requires folds not to mention now.

PR #373, opened 2016-03-31
Multi-line strings in toml can now contain quotes

jystic:
/cc @amosr @tranma


tranma:
great :godmode:

amosr:
nice...
to be honest I try to write parsers so they don't need backtracking, but I think the boat's boated on this one.

:floppy_disk:

PR #376, opened 2016-04-04
Smarter Source generator and tests for big data mode

amosr:
A small improvement in the generator can make a big difference in the number of discards, and time it takes to test them. This will be more necessary for more tests

### DIFF
--- a/test/Icicle/Test/Arbitrary/Source.hs
+++ b/test/Icicle/Test/Arbitrary/Source.hs
@@ -9,7 +9,7 @@
 module Icicle.Test.Arbitrary.Source where
 
 import           Icicle.Internal.Pretty
-import           Icicle.Source.Checker.Base (optionSmallData)
+import           Icicle.Source.Checker.Base (optionSmallData,optionBigData)
 import           Icicle.Source.Checker.Checker
 import           Icicle.Source.Checker.Error
 import           Icicle.Source.Query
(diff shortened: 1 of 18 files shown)


amosr:
I think this ends up being quite an improvement

tranma:
looks like this will still generate "group inside group", right? (that's not a problem, just more stuff to be discarded)

amosr:
I want to add more tests later, but I think this is OK to merge now

amosr:
oh. this ended up adding tests for BIG DATA MODE as well

test/Icicle/Test/Arbitrary/Source.hs
```diff
@@ -276,3 +182,283 @@ qwfDesugar qwf
...
+  = do tt <- arbitrary
+       genExp $ TypedGenInfo Map.empty tt tgiDefaultTableflip
+
+instance Arbitrary Prim where
+ arbitrary
+  = oneof
+        [ Lit . LitInt    <$> pos
+        , Lit . LitDouble <$> pos'
+        , Lit . LitString <$> (elements southpark)
+        , Lit . LitTime   <$> arbitrary ]
```
tranma: could we generate other prims (such as builtin functions) and generate the arguments for them? It would be nice to have the prims covered by quickcheck

amosr: a bunch of other prims are handled under "operator_bin" and "operator_unary" elsewhere, so they can be generated with arguments

amosr: looks like generating constructor applications are missing though

tranma: ah right, just have to remember to add any new prim to those generators.

tranma: yeah it's a bit difficult to keep track of what is and isn't being covered by the tests

test/Icicle/Test/Arbitrary/Source.hs
```diff
@@ -276,3 +182,308 @@ qwfDesugar qwf
...
+        , BuiltinMath Floor
+        , BuiltinMath Ceiling
+        , BuiltinMath Round
+        , BuiltinMath Truncate
+        , BuiltinTime DaysBetween
+        , BuiltinTime DaysEpoch
+        , BuiltinData Seq
+        , BuiltinData Box
+        , BuiltinMap MapKeys
+        , BuiltinMap MapValues ]
```
amosr: better @tranma

tranma: yeah much clearer 👍

jystic:
Looks legit to me 👍 on 🍵

amosr:
ignoring 7.10 build failure for now, this pr is large enough

PR #377, opened 2016-04-04
pure/mutable array update

tranma:
Makes array update immutable (copy-update) by default, and recovers mutability whenever possible.

This probably needs some Avalanche arbitrary tests? I don't think either the Core tests or the repl tests cover enough array update situations.

@amosr


src/Icicle/Avalanche/Prim/Eval.hs
```diff
@@ -111,7 +111,7 @@ evalPrim p vs
       -> primError
 
 
-     PrimUpdate (PrimUpdateArrayPut t)
+     PrimArray (PrimArrayPutMutable t)
```
amosr: this should really be mutating but it would require the evaluator to take a heap. this is good for now

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -116,7 +116,7 @@ convertValues a_fresh statements
            -> XPrim a_fresh (PrimUnsafe (PrimUnsafeArrayCreate t))
                `xApp` n
          (x : xs)
-           -> XPrim a_fresh (PrimUpdate (PrimUpdateArrayPut t))
+           -> XPrim a_fresh (PrimArray  (PrimArrayPutImmutable t))
```
amosr: I think this can be a mutable put, since it is operating on a newly created array that nothing else can reference. also, I think this transform is done after mutate, so won't be recovered by mutate.
(not that we have many array literals)

src/Icicle/Avalanche/Prim/Eval.hs
tranma: yeah, because this isn't really mutating I can't rely on the Avalanche evaluator for tests

amosr: that's a shame. ah! I wonder whether the easiest way might be to actually create a new Statement for mutating put, instead of having mutating put as a Exp Flat Prim.
if it were on the statement level like WriteAccumulator etc currently are, it might be easier to modify the evaluator so you can update the scalar/exp heap there..

tranma: hmm after digging around in the tests, I think making this evaluator do actual mutation is the easiest way to test the `mutate` transform

src/Icicle/Avalanche/Statement/Simp/Mutate.hs
```diff
@@ -0,0 +1,174 @@
...
+              in  ForeachInts n (mutateExp livevars x)
+                                (mutateExp lvs y)
+                                (go s)
+
+         ForeachFacts fs vt ft s
+           -> ForeachFacts fs vt ft (go s)
+
+         Block stmts
+           -> let f (st:sts) = mutateStmt (livevars <> Set.unions (fmap livevarsStmt sts)) st
+                              : fmap go sts
```
amosr: I think the first part here is right, getting the livevars of all the remaining sts, but shouldn't this be `f sts` rather than `fmap go sts`? as it is, each successive substatement will lose the livevars of those after it

amosr:
:dancers: 
except for `fmap go sts` comment

looks good. 
I don't think it deals with multiple reads of different names (eg below).
however, given that accumulators are only introduced by folds, and read then written straight away, we can probably show that this will never happen from Core.
we could make it an invariant that the body of a read cannot contain another read of the same accumulator. (not suggesting you implement that invariant - just that it's possible, and fine justification for not handling this more complicated case that won't happen)

yep, this is good. :+1: 

```
read FIRSTREAD = array

read x = array
let x' = copyUpdate#  x ix val
write array = x'

let oldval = index# FIRSTREAD ix
```

src/Icicle/Avalanche/Statement/Simp/Mutate.hs
tranma: ah yes, this is meant to be `f sts`, thanks for the catch :+1:

tranma:
Added a test that compares the eval results of Core vs Avalanche Flat vs Sea

@amosr @jystic

test/Icicle/Test/Language.hs
```diff
@@ -0,0 +1,154 @@
...
+            -> case seaRes of
+                 Left err
+                   -> return
+                    $ counterexample "Sea eval failed"
+                    $ counterexample (show $ pretty err)
+                    $ failed
+                 Right retSea
+                   -> return
+                    $ property
+                    $ retCore == retFlat && retFlat == retSea
```
amosr: === and .&&. would be better here?

test/Icicle/Test/Sea/Seaworthy.hs
```diff
@@ -34,7 +33,7 @@ namer = A.namerText (flip Var 0)
 --   Like prop_psv, but without psv.
 prop_seaworthy wt
  = testIO
- $ do let seaProgram = Map.singleton (Attribute "eval") (wtAvalanche wt)
+ $ do let seaProgram = Map.singleton (Attribute "eval") (wtAvalancheFlat wt)
```
amosr: huh?! it was trying to evaluate the unflattened program? oh, or did you just rename the field?

tranma: I renamed the field to make it clear the avalanche is flat

src/Icicle/Sea/FromAvalanche/Program.hs
```diff
@@ -205,7 +205,7 @@ seaOfXValue v t
       | ArrayT t' <- t
       -> let len = length vs
              writes arr (v',i)
-              = prim (PrimUpdate $ PrimUpdateArrayPut t')
+              = prim (PrimArray $ PrimArrayPutImmutable t')
```
amosr: hm- this should be dead code now, because this transform is elsewhere. can you change it to error if an array value makes it this far?

src/Icicle/Avalanche/Program.hs
```diff
@@ -27,12 +28,50 @@ instance TransformX Program where
...
         statements' <-      transformX names exps  $ statements p
 
-        return $ Program 
+        return $ Program
                { input      = input p
                , bindtime   = bindtime'
                , statements = statements'
                }
 
+renameProgram :: (Name n -> Name n') -> Program a n p -> Program a n' p
```
amosr: nice...

amosr:
looks good :dog:

src/Icicle/Sea/FromAvalanche/Program.hs
tranma: ok, sounds fair

tranma: @amosr wait, on second thought.... are you sure about this? Buffers should be gone but arrays values can still occur here, I think?

amosr: oh - whoops

amosr: no - perhaps I changed it, I can't remember, but we need to convert arrays in case they contain any buffers

amosr: so they should be gone by now

tranma: the tests failed if I just error here when encountering an array value though

amosr: ok, fair enough. perhaps the arrays are only converted if they actually contain buffers

PR #379, opened 2016-04-05
don't discard substitions

tranma:
@amosr


amosr:
:musical_keyboard:

amosr:
tempted to, in the future, have substitutions as an abstract type rather than a type synonym for Map: just so it is obvious that map operations won't necessarily be correct

PR #381, opened 2016-04-08
Some speed things

amosr:
Substitution takes a Map of substitutions rather than performing a single substitution.

ForwardStmts keeps track of an environment of substitutions to perform, rather than substituting many, many times.

Statement mappend does not perform simplification on the blocks produced, instead leaving them for nestBlocks to simplify these later. This produces a few naming differences.

Dead usage only needs to check if the sizes are the same; since elements are only added to the usage, if the size is the same it means nothing was added.

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp.hs
+++ b/src/Icicle/Avalanche/Statement/Simp.hs
@@ -176,17 +176,55 @@ pullLets statements
 -- | Let-forwarding on statements
 forwardStmts :: (Hashable n, Eq n) => a -> Statement a n p -> FixT (Fresh n) (Statement a n p)
 forwardStmts a_fresh statements
- = transformUDStmt trans () statements
+ = transformUDStmt trans Map.empty statements
  where
-  trans _ s
+  sub e x = lift $ subst a_fresh e x
+  subS e x f 
+   = do x' <- sub e x
+        return (e, f x')
+
+  trans e s
    = case s of
       Let n x ss
        | isSimpleValue x
-       -> do    s' <- lift $ substXinS a_fresh n x ss
-                progress ((), s')
+       -> do  x' <- sub e x
+              progress (Map.insert n x' e, Block [ss])
+       | otherwise
+       -> do  x' <- sub e x
+              return (Map.delete n e, Let n x' ss)
 
-      _ -> return ((), s)
+      If x ss es
+       -> subS e x $ \x' -> If x' ss es
 
+      ForeachInts n from to ss
+       -> do from' <- sub e from
+             to'   <- sub e to
+             let e' = Map.delete n e
+             return (e', ForeachInts n from' to' ss)
+
+      InitAccumulator (Accumulator n vt x) ss
+       -> subS e x $ \x' -> InitAccumulator (Accumulator n vt x') ss
+
+      Write n x
+       -> subS e x $ Write n
+
+      Output n t xs
+       -> do let subF (x,t') = (,t') <$> sub e x
+             xs' <- mapM subF xs
+             return (e, Output n t xs')
+
+      Read n _acc _t _ss
+       -> return (Map.delete n e, s)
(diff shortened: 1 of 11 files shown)


src/Icicle/Avalanche/Statement/Simp/Dead.hs
```diff
@@ -119,10 +118,17 @@ deadS us statements
...
        then (sU, sS)
        -- Make sure to use the original statements
        else deadLoop sU ss
+ where
+  -- We can cheat when checking 'equality' of the usage sets:
+  -- if they are the same size, they must contain the same elements.
+  -- (Because one is obtained by only inserting elements to the other)
+  eqUsage a b
+   =  Set.size (usageAcc a) == Set.size (usageAcc b)
+   && Set.size (usageExp a) == Set.size (usageExp b)
```
tranma: :+1:

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -176,17 +176,55 @@ pullLets statements
...
+      ForeachFacts ns _t  _f _ss
+       -> return (foldl (flip Map.delete) e (fmap fst $ factBindsAll ns), s)
+      Block{}
+       -> return (e,s)
+      KeepFactInHistory{}
+       -> return (e,s)
+      LoadResumable{}
+       -> return (e,s)
+      SaveResumable{}
+       -> return (e,s)
```
tranma: import Data.Map.Strict and see if it makes a difference?

src/Icicle/Common/Exp/Compounds.hs
```diff
@@ -216,21 +217,41 @@ substMaybe name payload into
...
+ | otherwise
+ = subst' a_fresh env (Set.unions $ Map.elems $ Map.map freevars env) into
+
+subst'  :: (Hashable n, Eq n)
+        => a
+        -> Map.Map (Name n) (Exp a n p)
+        -> Set.Set (Name n)
+        -> Exp a n p
+        -> Fresh n (Exp a n p)
+subst' a_fresh env frees into
```
tranma: here too, strict map just in case

test/cli/repl/t10.3-flatten/expected
```diff
@@ -97,18 +97,28 @@ for_facts (conv$2@{Time}, conv$1@{FactIdentifier}, conv$0$simp$60@{Error}, conv$
...
+  
   write acc$conv$26$simp$16 = Buf_push#@{Buf 3 Error} acc$conv$26$simp$16 conv$0$simp$60;
   read acc$conv$26$simp$17 = acc$conv$26$simp$17 [Buf 3 Int];
+  
+  
   write acc$conv$26$simp$17 = Buf_push#@{Buf 3 Int} acc$conv$26$simp$17 conv$0$simp$61;
+  
+  
+  
+
```
tranma: what's with blank lines? strange...

amosr: yeah - I think it is because of weird nesting of blocks in the pretty printed code. maybe something like `Block [Block [], Block [], Block []]`?
it is weird, but it doesn't seem to be stopping any later optimisations

src/Icicle/Common/Exp/Compounds.hs
amosr: good idea.
is it strange that there is no `Data.Set.Strict`?

tranma: I think they just make the keys in `Set` strict by default. For strict `Map` it's still only strict (whnf) keys though, not sure how much help that will be

amosr: ah, okay. I am not sure - I tried just switching to `Data.Map.Strict` and no real differences

amosr:
ok to merge?

tranma:
yeh sounds good :rocket: :moon:

PR #382, opened 2016-04-08
kill multiple accumulators at once

tranma:
this reduces my small example from ~2s to 1s
@amosr

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp/Dead.hs
+++ b/src/Icicle/Avalanche/Statement/Simp/Dead.hs
@@ -1,9 +1,10 @@
 {-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE PatternGuards     #-}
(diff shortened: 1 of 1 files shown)


src/Icicle/Avalanche/Statement/Simp/Dead.hs
```diff
@@ -40,70 +44,71 @@ instance Eq n => Monoid (Usage n) where
   , usageExp = Set.union (usageExp a) (usageExp b)
   }
```
amosr: could perhaps put kill as a field in here?

tranma: I tried that as well, but thought it's nicer on its own (maybe not)

amosr: no - you are right actually. I just saw that it doesn't thread through quite the same way after I commented

amosr:
looks good :+1:

PR #383, opened 2016-04-11
Compilation time improvements

tranma:
* Keep an annotation of variables in an expression, to avoid recomputing it during avalanche simplification. This brings the compilation time down quite a bit, from 5-6hrs to maybe half an hour. It's not great but it's a start:

```
./dist/build/ice/ice snapshot empty.psv dictionary-postpaid.toml 2016-01-01    3112.85s user 3233.29s system 312% cpu 33:52.69 total
```

### DIFF
--- a/src/Icicle/Avalanche/Simp.hs
+++ b/src/Icicle/Avalanche/Simp.hs
@@ -88,6 +88,8 @@ simpFlattened a_fresh p
    -- and now "x" is no longer mentioned, so can be removed.
    -- Doing this straight away means a smaller program for later passes.
    >>= return .  dead
+   -- Kill off statements that have no observable effect (no write to accumulators, etc.)
+   >>= return . killNoEffect
    -- Perform let-forwarding on statements, so that constant lets become free
    >>= forwardStmts a_fresh
    -- Try to evaluate any exposed primitives
--- a/src/Icicle/Avalanche/Simp.hs
+++ b/src/Icicle/Avalanche/Simp.hs
@@ -1,32 +1,34 @@
 -- | Convert Core programs to Avalanche
-{-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE DoAndIfThenElse   #-}
+{-# LANGUAGE DoAndIfThenElse     #-}
+{-# LANGUAGE NoImplicitPrelude   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 module Icicle.Avalanche.Simp (
     simpAvalanche
   , simpFlattened
   , pullLets
   ) where
 
-import              Icicle.Common.Exp
-import              Icicle.Common.Fresh
-import              Icicle.Common.FixT
+import           Icicle.Common.Exp
+import           Icicle.Common.FixT
+import           Icicle.Common.Fresh
 
-import qualified    Icicle.Core.Exp.Prim as CorePrim
-import qualified    Icicle.Core.Eval.Exp as CorePrim
+import qualified Icicle.Core.Eval.Exp                        as CorePrim
+import qualified Icicle.Core.Exp.Prim                        as CorePrim
 
-import qualified    Icicle.Avalanche.Prim.Eval as Flat
-import qualified    Icicle.Avalanche.Prim.Flat as Flat
-import              Icicle.Avalanche.Statement.Simp
-import              Icicle.Avalanche.Statement.Simp.Eval
-import              Icicle.Avalanche.Statement.Simp.Constructor
-import              Icicle.Avalanche.Statement.Simp.Melt
-import              Icicle.Avalanche.Statement.Simp.Mutate
-import              Icicle.Avalanche.Program
+import qualified Icicle.Avalanche.Prim.Eval                  as Flat
+import qualified Icicle.Avalanche.Prim.Flat                  as Flat
+import           Icicle.Avalanche.Program
+import           Icicle.Avalanche.Statement.Simp
+import           Icicle.Avalanche.Statement.Simp.Constructor
+import           Icicle.Avalanche.Statement.Simp.Eval
+import           Icicle.Avalanche.Statement.Simp.Melt
+import           Icicle.Avalanche.Statement.Simp.Mutate
+import           Icicle.Avalanche.Statement.Statement
 
-import              P
+import           P
 
-import              Control.Monad.Trans.Class
-import              Data.Hashable (Hashable)
+import           Control.Monad.Trans.Class
+import           Data.Hashable                               (Hashable)
 
 
 simpAvalanche
@@ -35,7 +37,7 @@ simpAvalanche
   -> Program a n CorePrim.Prim
   -> Fresh n (Program a n CorePrim.Prim)
 simpAvalanche a_fresh p
- = do p' <- transformX return (simp a_fresh) p
+ = do p' <- transformX return (simpAnn a_fresh) p
       s' <- (once $ forwardStmts a_fresh $ pullLets $ statements p')
          >>= once . thresherWithAlpha     a_fresh
          >>= once . forwardStmts a_fresh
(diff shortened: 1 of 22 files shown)


src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -90,44 +92,48 @@ convertValues a_fresh statements
...
+           -> xPrim (PrimArray  (PrimArrayPutMutable t))
                 `xApp` arrPrim (i + 1) n t xs
-                `xApp` (XValue a_fresh IntT (VInt i))
+                `xApp` (xValue IntT (VInt i))
                 `xApp` goB' t x
 
-  xApp
-   = XApp a_fresh
+  xApp x1 x2
+   = XApp (a_fresh, snd (annotOfExp x1) <> snd (annotOfExp x2)) x1 x2
```
amosr: :+1:

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -391,10 +430,16 @@ substXinS a_fresh name payload statements
...
+ -> Statement (Ann a n) n Prim
+ -> FixT (Fresh n) (Statement (Ann a n) n Prim)
 thresherNoAlpha a_fresh statements
  = transformUDStmt trans Map.empty statements
  where
+  xVar n
+   = XVar (a_fresh, Set.singleton n) n
   trans env s
    -- Check if it actually does anything:
    -- updates accumulators, returns a value, etc.
```
amosr: I think this "hasEffect" might end up being the bottleneck soon. it might be faster in a completely separate transform to thresher, since thresher is top-down and hasEffect seems to be bottom-up.
you could perhaps put it in dead (it does make sense), but dead stuffs around with the fixpoint of loops, so hasEffect might be better as a separate pass since it doesn't need loop fixpoints.

src/Icicle/Common/Exp/Compounds.hs
```diff
@@ -226,68 +228,75 @@ subst1 :: (Hashable n, Eq n)
...
+   subst' env frees xx
+    = let go = subst' env frees
+      in case xx of
+       XVar _ n
+        | Just payload <- Map.lookup n env
+        -> return payload
+        | otherwise
+        -> return xx
+       XApp a p q
+        -> XApp a <$> go p <*> go q
```
amosr: do you need to rebuild a's allvars from `go p` and `go q`, if substitution was applied in one of them?
if not, maybe explain as a comment why it works for anormal

src/Icicle/Common/Exp/Simp/ANormal.hs
```diff
@@ -53,18 +57,19 @@ anormal a_fresh xx
...
         -- Tag the result with the union of the original bindings (bs)
         -- as well as the new names of the bindings.
-        let allNames' = allNames <> Set.fromList (fmap fst bs')
-        return $ makeLets (a_fresh, allNames') bs' x'
+        let !allNames' = {-# SCC anormal_allnames #-} allNames <> Set.fromList (fmap fst bs')
+        let !ret = {-# SCC anormal_mklets#-}makeLets (a_fresh, allNames') bs' x'
+        return ret
 
  where
   -- All the variables inside the bindings
```
amosr: there are a few more opportunities here: renames does a `makeLets` then `takeLets` dance, with a single substitution at each step.
you could probably give renames a substitution environment to apply to each binding, just adding to the environment and substituting the whole environment, each time. and you could build up `seen` in reverse order or something, I guess.

amosr:
fantastic :monkey: :man:
am worried that substitution might need to do more to keep track of allvars correctly, but there might be an obvious reason that it is not necessary for anormal to be correct

src/Icicle/Avalanche/Statement/Simp.hs
tranma: yeah, that sounds fair. I'll put in some SCC to see how much `hasEffect` has an effect (heh) on the performance

src/Icicle/Common/Exp/Compounds.hs
tranma: hm, so I thought anormal only substitutes with a `XVar fresh_variable` that it just generated anyway, but it's probably more principled to just rebuild the set here. I hope that doesn't undo all the performance gains just achieved.

amosr: ahh, right. but I think there are other parts that perform substitution after you introduce allvars into the annotation: `constructor` and `forwardStmts` inside `crunch`, might perform substitution with non-fresh names, so the annotation won't be valid afterwards

tranma: yeah you're right, this subst is actually used in substXinS as well. I think I'll carry around a flag: if substitution happened  in the subtrees, rebuild the annotation. Hopefully that isn't too expensive.

amosr: sounds reasonable

jystic:
Wow, this is awesome! 6 hrs -> 30mins 💯

src/Icicle/Avalanche/Simp.hs
```diff
@@ -1,75 +1,81 @@
...
-         >>= once . forwardStmts a_fresh
-         >>= nestBlocks   a_fresh
-         >>= once . thresherWithAlpha     a_fresh
-         >>= transformX return (return . simpEvalX CorePrim.evalPrim CorePrim.typeOfPrim)
-         >>= return .  dead
-
-      return $ p { statements = s' }
+ = do !p' <-    transformX return (simpAnn a_fresh) p
+      !s' <-   (once $ forwardStmts a_fresh $ pullLets $ statements p')
+          >>=!! once . thresherWithAlpha     a_fresh
```
amosr: this is the best operator ever - I had no idea it existed

tranma: that's because I just made it up 😄

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -318,10 +316,109 @@ substXinS
  :: (Hashable n, Eq n)
  => a
  -> Name n
+ -> Exp a n p
+ -> Statement a n p
+ -> Fresh n (Statement a n p)
+substXinS a_fresh name payload statements
```
amosr: you may get some benefit from making this take a substitution map rather than a single substitution, I'm not sure

amosr:
this is awesome - what is it down to now?

tranma:
at commit 389117, it was ~14m with shitty NFData (I just printed the AST to the console). But the last 3 commits have huge memory blowup problems, so I think I'm going to tweak 389117 and merge that. The rest can come in another PR.

src/Icicle/Pipeline.hs
```diff
@@ -296,10 +297,10 @@ flattenAvalanche av
  = join
  . second snd
  . first CompileErrorFlatten
- $ Fresh.runFreshT go (freshNamer "flat")
+ $!! Fresh.runFreshT go (freshNamer "flat")
  where
   go = do s' <- AS.flatten () (AP.statements av)
-          return $ checkAvalanche (av { AP.statements = s' })
+          return $ checkAvalanche (av { AP.statements = force s' })
```
tranma: @amosr so this exact amount of strictness proved to be the sweet spot (time ~14m, space < 8GB). If I make things more strict, ice eats up so much memory it gets killed by the OS. I'll play around with more strictness in another PR.

tranma:
Probably can't make it much better than 14m because clang took up over half of that time.

amosr:
> clang took up over half of that time

that sounds like a pretty good "litmus test" for later benchmarks and issues, I suppose. I'm curious - can you turn off clang's `-O2` or whatever, and if so is it faster? maybe that would be an OK option for the ice users, when they're just testing with smaller datasets.

tranma:
@amosr I thought the ice users would be working with small datasets too, but it's actually somewhat large (still fit on one machine). But maybe clang `O2` isn't necessary because the generated C code is already pretty efficient (I think).

amosr:
Oh yeah, I misread or misremembered the runtime in Wilson's example: I thought it was 3m but it's 30m.
It'd be interesting anyway, if it takes the 15m to 5m and only makes the runtime a few minutes faster, worth a try.

PR #384, opened 2016-04-14
Fix chord output dense

tranma:
It's currently like this for `chord --output-psv dense`:

```
ID00|foo|bar|2016-01-02|foo|bar|2016-01-03
```

fixes so that it becomes:

```
ID00|foo|bar|2016-01-02
ID00|foo|bar|2016-01-03
```

@jystic


jystic:
👍

PR #387, opened 2016-04-20
sort arrays

tranma:
this adds a baked-in shitty sort to Avalanche, exposed in Source as a builtin `sort` function. We need this for computing quantiles among other things.

### DIFF
--- a/main/icicle-repl.hs
+++ b/main/icicle-repl.hs
@@ -104,6 +104,7 @@ data ReplState
    , hasSeaPreamble   :: Bool
    , hasSea           :: Bool
    , hasSeaAssembly   :: Bool
+   , hasSeaLLVMIR     :: Bool
    , hasSeaEval       :: Bool
    , doCoreSimp       :: Bool }
 
@@ -124,6 +125,7 @@ data Set
    | ShowSeaPreamble    Bool
    | ShowSea            Bool
    | ShowSeaAssembly    Bool
+   | ShowSeaLLVMIR      Bool
    | ShowSeaEval        Bool
    | CurrentTime        Time
    | PerformCoreSimp    Bool
@@ -144,7 +146,7 @@ data Command
 
 defaultState :: ReplState
 defaultState
-  = (ReplState [] demographics (unsafeTimeOfYMD 1970 1 1) False False False False False False False False False False False False False False False False False)
+  = (ReplState [] demographics (unsafeTimeOfYMD 1970 1 1) False False False False False False False False False False False False False False False False False False)
     { hasCoreEval = True
     , doCoreSimp  = True }
 
@@ -213,6 +215,9 @@ readSetCommands ss
 
     ("+c-assembly":rest)   -> (:) (ShowSeaAssembly True)  <$> readSetCommands rest
     ("-c-assembly":rest)   -> (:) (ShowSeaAssembly False) <$> readSetCommands rest
+
+    ("+c-llvm-ir":rest)   -> (:) (ShowSeaLLVMIR True)  <$> readSetCommands rest
+    ("-c-llvm-ir":rest)   -> (:) (ShowSeaLLVMIR False) <$> readSetCommands rest
 
     ("+c-eval":rest)       -> (:) (ShowSeaEval     True)  <$> readSetCommands rest
     ("-c-eval":rest)       -> (:) (ShowSeaEval     False) <$> readSetCommands rest
@@ -356,6 +361,12 @@ handleLine state line = case readCommand line of
                Left  e -> prettyOut (const True) "- C assembly error:" e
                Right r -> prettyOut (const True) "- C assembly:" r
 
+           when (hasSeaLLVMIR state) $ do
+             result <- liftIO . runEitherT $ Sea.irOfPrograms Sea.NoPsv [(Attribute "repl", f')]
+             case result of
+               Left  e -> prettyOut (const True) "- C LLVM IR error:" e
+               Right r -> prettyOut (const True) "- C LLVM IR:" r
+
            when (hasSeaEval state) $ do
              result <- liftIO . runEitherT $ P.seaEval (currentTime state) (facts state) finalSource f'
              case result of
@@ -442,6 +453,10 @@ handleSetCommand state set
     ShowSeaAssembly b -> do
         HL.outputStrLn $ "ok, c assembly is now " <> showFlag b
         return $ state { hasSeaAssembly = b }
+
+    ShowSeaLLVMIR b -> do
+        HL.outputStrLn $ "ok, c llvm ir is now " <> showFlag b
+        return $ state { hasSeaLLVMIR = b }
 
     ShowSeaEval b -> do
         HL.outputStrLn $ "ok, c evaluation now " <> showFlag b
@@ -538,6 +553,7 @@ showState state
     , flag "c-preamble:   " hasSeaPreamble
     , flag "c:            " hasSea
     , flag "c-assembly:   " hasSeaAssembly
+    , flag "c-llvm-ir:    " hasSeaLLVMIR
     , flag "c-eval:       " hasSeaEval
     ,      "now:          " <> T.unpack (renderTime $ currentTime state)
     ,      "data:         " <> show (length $ facts state) <> " rows"
--- a/src/Icicle/Avalanche/Check.hs
+++ b/src/Icicle/Avalanche/Check.hs
@@ -85,7 +85,13 @@ checkStatement frag ctx stmt
                    $ checkExp frag (ctxExp ctx) x
                Let n x' <$> go stmts
 
-        ForeachInts n from to stmts
+        While t n to stmts
+         -> do to'   <- first ProgramErrorExp
+                      $ checkExp frag (ctxExp ctx) to
+
+               While t n to' <$> go stmts
+
+        ForeachInts t n from to stmts
          -> do from' <- first ProgramErrorExp
                       $ checkExp frag (ctxExp ctx) from
                to'   <- first ProgramErrorExp
@@ -97,7 +103,7 @@ checkStatement frag ctx stmt
                requireSame (ProgramErrorWrongType from) tf (FunT [] IntT)
                requireSame (ProgramErrorWrongType to)   tt (FunT [] IntT)
 
-               ForeachInts n from' to' <$> go stmts
+               ForeachInts t n from' to' <$> go stmts
 
         ForeachFacts binds vt lo stmts
          -> ForeachFacts binds vt lo <$> go stmts
@@ -191,7 +197,11 @@ statementContext frag ctx stmt
            ctxX' <- insert (ctxExp ctx) n t
            return (ctx { ctxExp = ctxX' })
 
-    ForeachInts n _ _ _
+    While _ n _ _
+     -> do ctxX' <- insert (ctxExp ctx) n (FunT [] IntT)
+           return (ctx { ctxExp = ctxX' })
+
+    ForeachInts _ n _ _ _
      -> do ctxX' <- insert (ctxExp ctx) n (FunT [] IntT)
            return (ctx { ctxExp = ctxX' })
 
--- a/src/Icicle/Avalanche/Check.hs
+++ b/src/Icicle/Avalanche/Check.hs
@@ -86,10 +86,17 @@ checkStatement frag ctx stmt
                Let n x' <$> go stmts
 
         While t n to stmts
-         -> do to'   <- first ProgramErrorExp
-                      $ checkExp frag (ctxExp ctx) to
+         -> do to'  <- first ProgramErrorExp
+                     $ checkExp frag (ctxExp ctx) to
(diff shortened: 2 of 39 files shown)


src/Icicle/Avalanche/Check.hs
```diff
@@ -88,7 +88,15 @@ checkStatement frag ctx stmt
...
                Let n x' <$> go stmts
 
-        ForeachInts n from to stmts
+        While t n from to stmts
+         -> do from' <- first ProgramErrorExp
+                      $ checkExp frag (ctxExp ctx) from
+               to'   <- first ProgramErrorExp
+                      $ checkExp frag (ctxExp ctx) to
+
+               While t n from' to' <$> go stmts
```
amosr: hmm. if "n" is supposed to start with some value and be modified, this should probably take an accumulator name, rather than a scalar name.
that way you can just insert "n" into the accumulator map, and inside it can do "Write n".

tranma: yeah, when I actually use this construct I just init and modify the acc manually

amosr: it's a bigger change to typecheck what you're doing now though: you'd need to make the rest of the typechecking return the environment of bound scalars, and check that it has the same type as "from" etc.
whereas if you use an accumulator, you can just check that from and to types are the same, then insert "n" as an accumulator of that type and recurse

tranma:
@amosr wanna have another look at this?

src/Icicle/Sea/Eval.hs
```diff
@@ -353,6 +354,16 @@ assemblyOfPrograms psv programs = do
...
 
+irOfPrograms
+  :: (Show a, Show n, Pretty n, Eq n)
+  => Psv
+  -> [(Attribute, Program (Annot a) n Prim)]
+  -> EitherT SeaError IO Text
+irOfPrograms psv programs = do
+  code    <- hoistEither (codeOfPrograms psv programs)
+  options <- getCompilerOptions
+  firstEitherT SeaJetskiError (compileIR options code)
```
jystic: Awesome 👍

src/Icicle/Avalanche/Check.hs
```diff
@@ -88,7 +88,13 @@ checkStatement frag ctx stmt
                    $ checkExp frag (ctxExp ctx) x
                Let n x' <$> go stmts
 
-        ForeachInts n from to stmts
+        While t n to stmts
+         -> do to'   <- first ProgramErrorExp
+                      $ checkExp frag (ctxExp ctx) to
+
+               While t n to' <$> go stmts
```
amosr: I guess this should get `n` from the accumulator map and check that `to'` has the same type

amosr: oh, never mind, I forgot that they were torn apart

amosr: ok. so if the way you plan on using this is:
```
Init n = initx
While n to
  ...
  Write n updx
```
this should look up `n` in the accumulator environment, and check that the while expression `to` has the same type as `n`. It doesn't need to add `n` to the expression environment below

tranma: yeah, I forgot about this. This should have a requireSame here, and the insert n below no longer applies.

amosr: ah. but if the equality check is hidden in the `While` expression itself, it will not be melted. so I suppose it is better off restricting it to primitives. or maybe just restricting it to a `Bool`.

tranma: d'oh, I already fixed it but haven't pushed

tranma: oh, I didn't think of that. I did this:

```
        Write n x
         -> do x'   <- first ProgramErrorExp
                     $ checkExp frag (ctxExp ctx) x

               vt   <- maybeToRight (ProgramErrorNoSuchAccumulator n)
                     $ Map.lookup n $ ctxAcc ctx

               let t = annType (annotOfExp x')

               requireSame (ProgramErrorWrongType x) t (FunT [] vt)
               return (Write n x')
```

but I suppose that wouldn't work it `vt` needs to be melted

amosr: ah, good. yeah, we can probably ignore the melting thing, since While loops will only be introduced by us on melting

tranma:
fixed up and squashed

main/icicle-repl.hs
```diff
@@ -214,6 +216,9 @@ readSetCommands ss
     ("+c-assembly":rest)   -> (:) (ShowSeaAssembly True)  <$> readSetCommands rest
     ("-c-assembly":rest)   -> (:) (ShowSeaAssembly False) <$> readSetCommands rest
 
+    ("+c-llvm-ir":rest)   -> (:) (ShowSeaLLVMIR True)  <$> readSetCommands rest
+    ("-c-llvm-ir":rest)   -> (:) (ShowSeaLLVMIR False) <$> readSetCommands rest
+
```
amosr: nice

src/Icicle/Avalanche/Check.hs
```diff
@@ -88,7 +88,20 @@ checkStatement frag ctx stmt
...
+         -> do to'  <- first ProgramErrorExp
+                     $ checkExp frag (ctxExp ctx) to
+               vt   <- maybeToRight (ProgramErrorNoSuchAccumulator n)
+                     $ Map.lookup n $ ctxAcc ctx
+
+               let tt = annType (annotOfExp to')
+
+               requireSame (ProgramErrorWrongType to) tt (FunT [] vt)
+
+               While t n to' <$> go stmts
```
amosr: yep, this looks good

src/Icicle/Avalanche/Eval.hs
```diff
@@ -138,19 +142,36 @@ evalStmt evalPrim now xh values bubblegum ah stmt
...
+                check WhileNe = (/=)
+            let evalLoop curr@(ah',out,bg) end
+                 = do accv <- maybeToRight (RuntimeErrorLoopAccumulatorBad acc)
+                            $ Map.lookup acc ah'
+                      if check t accv end
+                      then do next <-  appendOutputs out bg
+                                   <$> go xh ah' stmts
+                              evalLoop next end
+                      else return curr
+            evalLoop (ah, mempty, mempty) tov
```
amosr: cool

src/Icicle/Avalanche/Prim/Eval.hs
```diff
@@ -137,6 +137,25 @@ evalPrim p vs
...
+       $ VArray
+       $ let low   = min ix1 ix2
+             high  = min ix2 ix2
+             xl    = varr List.!! low
+             xh    = varr List.!! high
+             varr1 = List.take (low - 1) varr
+             varr3 = List.drop high      varr
+             varr2 = List.drop low
+                   $ List.take (high - 1) varr
+         in varr1 <> [xl] <> varr2 <> [xh] <> varr3
```
amosr: 👍

src/Icicle/Avalanche/Statement/Flatten/Exp.hs
```diff
@@ -118,8 +126,232 @@ flatX a_fresh xx stm
...
+             n_acc_arr      <- freshPrefix "sort_acc_array"
+             n_acc_heapSize <- freshPrefix "sort_acc_heap_size"
+             sort   <- heapSort n_acc_arr n_acc_heapSize
+             sorted <- stm (xVar n_loc_arr)
+             return
+               $ initArr t n_acc_arr array'
+               $ Block
+               [ sort
+               , readArr t n_loc_arr n_acc_arr sorted
+               ]
```
amosr: this is really scary/impressive

src/Icicle/Avalanche/Statement/Simp.hs
```diff
@@ -210,11 +211,16 @@ forwardStmts a_fresh statements
       If x ss es
        -> subS e x $ \x' -> If x' ss es
 
-      ForeachInts n from to ss
+      While t n end ss
+       -> do end'  <- sub e end
+             let e' = Map.delete n e
+             return (e', While t n end' ss)
```
amosr: I am not sure, but I do not think this needs to remove `n` from the map any more, since `n` is an accumulator, and `n` is not bound here anyway

src/Icicle/Avalanche/Statement/Simp/Dead.hs
```diff
@@ -59,11 +59,15 @@ deadS us statements
         in  if   usedX n sU
             then (mconcat [xU, sU], sK, Let n x sS)
             else (sU, sK, sS)
-    ForeachInts n from to ss
+    While t n end ss
+     -> let (sU, sK, sS) = deadLoop us ss
+            eU           = usageX end
+        in (addUsedA n $ mconcat [sU, eU], sK, While t n end sS)
```
amosr: I think this is equivalent to `mconcat [sU, eU, usageA n]`. is this faster?
(nitpick)

src/Icicle/Avalanche/Statement/Simp.hs
tranma: yeah, `n` isn't local anymore, thanks for the catch

src/Icicle/Avalanche/Statement/Simp/Dead.hs
tranma: maybe not faster but probably cleaner

src/Icicle/Avalanche/Statement/Simp/Mutate.hs
```diff
@@ -111,8 +114,11 @@ livevarsStmt ss = case ss of
   Let _ x s
    -> livevarsExp x <> livevarsStmt s
 
-  -- loop counter is live at the beginning and inside the loop.
-  ForeachInts n _ end body
+  -- loop counter is live after it is assigned the initial value,
+  -- and live inside the loop.
+  While _ n end body
+   -> Set.insert n (livevarsExp end <> livevarsStmt body)
```
amosr: not sure about this since it's an accumulator now?

amosr: (ignore me if it is still correct)

src/Icicle/Sea/FromAvalanche/Analysis.hs
```diff
@@ -129,20 +133,21 @@ readsOfProgram = readsOfStatement . statements
...
-     SaveResumable _ _     -> Map.empty
-     Output _ _ _          -> Map.empty
-     KeepFactInHistory _   -> Map.empty
+     Block []                -> Map.empty
+     Block (s:ss)            -> readsOfStatement s `Map.union`
+                                readsOfStatement (Block ss)
+     Let _ _ ss              -> readsOfStatement ss
+     If _ tt ee              -> readsOfStatement tt `Map.union`
+                                readsOfStatement ee
+     While          _ _ _ ss -> readsOfStatement ss
```
amosr: in some sense this is actually reading from the accumulator every time. I am not sure what `readsOfStatement` is actually used for though, so it might not be necessary to add it here

amosr:
looks great. 💊 😮 🎉 
only thing - I think the "Sort" primitive might need to be added to the Source arbitrary generator

src/Icicle/Avalanche/Statement/Simp/Mutate.hs
tranma: so I think the `acc` is live here, because it's used in the while body. For this case:

```
init acc = some array value V
let local = put_immutable x acc
while (acc == something) {...}
```

I don't think we want the put_immutable to be changed into a put_mutable?

src/Icicle/Sea/FromAvalanche/Analysis.hs
tranma: It doesn't seem to be used for dead code, but to be safe I should insert `n` here

tranma: ah, gotta change While to carry around the ValType though

src/Icicle/Avalanche/Statement/Simp/Mutate.hs
amosr: ah ok, good point

src/Icicle/Sea/FromAvalanche/Analysis.hs
amosr: might not be worth worrying about then, if it doesn't seem to be necessary

jystic: it's used just once, to declare C variables:
```hs
    , "#line 1 \"compute function" <+> seaOfStateInfo state <> "\""
    , "void " <> pretty (nameOfProgram state) <> "(" <> pretty (nameOfStateType state) <+> "*s)"
    , "{"
    , indent 4 . vsep
               . fmap defOfAccumulator
               . Map.toList
               $ accumsOfProgram program `Map.union`
                 readsOfProgram  program
```

amosr: ok, it's not important to mark the while loop as a read then, because the accumulator will have already added it

PR #388, opened 2016-04-21
Fix reify bug

amosr:
> Tran Ma·9:19 AM convert_is_well_typed failed for this:
```
feature gonzofeaturename
~> fold1 gonzo = 59.14305720951913 : (let kermit = 2
                                      ~> gonzo)
~> Right gonzo
```

@tranma


tranma:
looks good 👍 
that was a subtle bug

PR #389, opened 2016-04-21
fix 7.10 build

tranma:
I'm not sure why we don't see these errors on 7.8?

@jystic

### DIFF
--- a/src/Icicle/Avalanche/Check.hs
+++ b/src/Icicle/Avalanche/Check.hs
@@ -1,6 +1,5 @@
 -- | Evaluate Avalanche programs
 {-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE DeriveGeneric     #-}
 module Icicle.Avalanche.Check (
     checkProgram
   , ProgramError(..)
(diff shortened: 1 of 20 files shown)


jystic:
Ok with this if we put bangs on all the data types that are deriving `NFData` so that `seq` is the same as `deepseq`

olorin:
> I'm not sure why we don't see these errors on 7.8?

https://ghc.haskell.org/trac/ghc/wiki/Migration/7.10#deepseq-1.4.0.0deepseq-1.4.1.0

jystic:
Maybe it's better to make everything strict and do:
```
instance NFData Foo where rnf x = seq x ()
```
instead of `deriving Generic`, at least it's guaranteed to be the same.

tranma:
yeah, I'm leaning towards removing derive Generic completely. Don't want to make the AST too strict

jystic:
```hs
class MFData where
  mediumSeq :: a -> b -> b
```

tranma:
`($!.)` instead of `($!!)`?

tranma:
it's half a bang

jystic:
or `($!?)`

it's unknown how much more than whnf we seq'd

amosr:
does `($¡)` work?

tranma:
should I change all the NFData and `$!!` to MedData and `$?!`?

tranma:
tested this with `ice` on our Big Data Case (tm).
Compilation time is about the same. Space wise, looks like it used a bit more memory than before, but still made it until the GC cleared up some 30GB, so that's still ok.

jystic:
Looks good to me ✌️

PR #391, opened 2016-05-09
Even more savage builtins

tranma:
Some baked-in map and array Source "data types", to make queries like `days spent in some state` and `quantiles` possible.

`now` is now `Pure`.


src/Icicle/Sea/FromAvalanche/Prim.hs
```diff
@@ -165,6 +165,8 @@ seaOfPrimArray p
       -> PDAlloc (prefixOfValType (ArrayT t) <> "put_immutable") Nothing
      PrimArraySwap t
       -> PDFun   (prefixOfValType (ArrayT t) <> "swap") Nothing
+     PrimArrayDel t
+      -> PDFun   (prefixOfValType (ArrayT t) <> "delete") Nothing
```
amosr: is this C function implemented?

data/libs/prelude.icicle
```diff
@@ -112,6 +112,52 @@ numflips' v t
...
+                                 | Some nv'
+                                 -> map_insert v (nv' + n) m
+                                end
+                              )
+                 ~> (Some (x, tx), 0, m')
+                 )
+             end
+          )
+      end
+ ~> snd ds.
```
amosr: this is amazing. would be good to have some test cases for these two, partly for sanity and partly as examples

src/Icicle/Sea/FromAvalanche/Prim.hs
tranma: I forgot, it should be :) Turns out we don't need map_delete for days_in but might as well add it

amosr: should be able to do it with a memcpy, or just the simplest possible way otherwise. I guess it has to copy and be immutable? hm

tranma: I'm inclined to make it mutable

amosr: ok- and then I guess flattening should convert `delete x` to `delete (copy x)` or something?

tranma: hm, actually, now I'm not sure. Maybe we should have both `delete_immutable` and `delete_mutable`, like for `insert`

amosr: yeah, though I wonder if instead of `put_immutable` we could just have `put_immutable = put_mutable . copy`. then we'd only need one version of each mutable array operation. would that complicate mutation recovery though?

tranma: that should be ok, mutation recovery is done in one pass over the AST

tranma: hm, I would prefer `put_immutable` to be the default like it currently is though. Just in case the mutation recovery is incorrect. If `mutable` is the default and we try to find places where we should not mutate, it's a bit more scary.

amosr: oh, I don't mean we should mutate by default: just that we define put_immutable in terms of mutate and copy?

tranma: ah right. I'm not sure, it makes the primitives simpler but maybe also more verbose and messier?

amosr: yeah, perhaps just make `delete` immutable for now and don't worry about a mutable one

tranma:
@HuwCampbell do you think `quantile_on_sorted` and `days_in` look ok?

jystic:
No idea about the quantile stuff, but the rest looks good 🌴

HuwCampbell:
I've been trying, but I have not been able to get sensible values out in the repl for the quantiles feature.

data/libs/prelude.icicle
```diff
@@ -112,6 +112,62 @@ numflips' v t
...
+# (might need a function unbox :: Sum Error a -> Option a)
+quantile_on_sorted k q array
+ =  let len          = length array
+ ~> let actual       = (double k) / (q * double (len - 1))
+ ~> let actual_ceil  = ceil  actual
+ ~> let actual_floor = floor actual
+ ~> let floor_weight = double actual_ceil - actual
+ ~> let val_floor    = index array actual_floor
+ ~> let val_ceil     = index array actual_ceil
+ ~> double val_floor + double val_ceil * double (1 - floor_weight).
```
HuwCampbell: think you need to multiply val_floor by floor weight.

data/libs/prelude.icicle
```diff
@@ -112,6 +112,62 @@ numflips' v t
...
  ~> case s | (s0,i) -> i end.
 
+#######################
+### Example queries ###
+#######################
+
+# Quantiles from a sorted array of ints, same as in R.
+# TODO safe array_index in Source that maps to unsafe index in Avalanche
+# (might need a function unbox :: Sum Error a -> Option a)
+quantile_on_sorted k q array
```
HuwCampbell: Probably don't need to indicies. A double between 0 and 1 should do.

test/cli/repl/t31-builtin/expected
```diff
@@ -60,4 +60,25 @@ ok, loaded test/cli/repl/data.psv, 13 rows
 - Core evaluation:
 [homer, ["arm","head","torso"]]
 
-> > 
\ No newline at end of file
+> > -- Baked-in maps and arrays
+> > ok, time set to 2016-01-01
+> > - C evaluation:
+[homer, -10835.166666666666]
```
HuwCampbell:
```
> feature injury ~> let x = (latest 4 ~> days between now time) ~> quantile_on_sorted 1 2 (sort x) , x
- Core evaluation:
[homer, (-10835.166666666666,[-9496,-8035,-6209,-2191])]
```

tranma:
@HuwCampbell do you mean you get tombstone or NaN values? (or unsafe indexing exception? I'm not too sure what the quantile should be in that case)

HuwCampbell:
Should be the median.

tranma:
@HuwCampbell
```
> feature injury ~> let x = (latest 4 ~> days between time now) ~> quantile_on_sorted 1 2 (sort x) , x
[homer, (2860.6666666666665, [9496,8035,6209,2191])]
```

HuwCampbell:
2860 is not the median of those numbers.

tranma:
@HuwCampbell sorry, my bad, I messed up the definition.

```
[homer, (7122.0
        ,[9496,8035,6209,2191])]
```

HuwCampbell:
Wait, I think you interpreted my statement above, that the value in the test should be the median as something else.

HuwCampbell:
Asking for nonsense quantiles should yield a missing value.

tranma:
@HuwCampbell ok, I should have paid more attention :) you meant quantile of 1/2 is the median right? I'll just return a None in the error case

HuwCampbell:
That's right. I also reckon that you could use a double instead of k and q, but that's up to you.

tranma:
yeah I agree.

```
feature injury ~> let x = (latest 4 ~> days between time now) ~> quantile_on_sorted 0.5 (sort x), x
[homer, (7122.0
        ,[9496,8035,6209,2191])]

feature injury ~> let x = (latest 4 ~> days between time now) ~> quantile_on_sorted 5.0 (sort x), x
[homer, (tombstone
        ,[9496,8035,6209,2191])]
```

PR #392, opened 2016-05-11
Get rid of MapT in Source

tranma:
Changed my mind, there should be no distinction between GroupT (from `group`) and MapT (from user maps). The type checker checks the context, not the type, to ensure there is no Group in Group anyway.

Should have listened to my manager :)


tranma:
Also fixed this inconsistency between C and Core eval:

```
> feature injury ~> newest fields
- C error:
Cannot convert type Struct [(action,Option String)
                           ,(admitted,Option Time)
                           ,(location,String)
                           ,(severity,Int)]

- Core evaluation:
[homer, [(action,tombstone)
        ,(admitted,2010-01-02)
        ,(location,"torso")
        ,(severity,1)]]
```

Now Sea allows outputting structs too.

@jystic

jystic:
> Now Sea allows outputting structs too.

Woo, that's awesome!

This looks good 💾

PR #393, opened 2016-05-12
Remove duplicate work in constant folding

tranma:
This reduces the compilation time of a big dictionary with nested cases from 17057.11s to 257.91s.
@amosr

### DIFF
--- a/src/Icicle/Avalanche/Statement/Simp/Eval.hs
+++ b/src/Icicle/Avalanche/Statement/Simp/Eval.hs
@@ -32,6 +32,11 @@ simpEvalX'  :: (Hashable n, Eq n)
             -> Maybe (Exp a n p)
 simpEvalX' ev ty = go
   where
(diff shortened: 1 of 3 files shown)


amosr:
gosh!

src/Icicle/Avalanche/Statement/Simp/Eval.hs
```diff
@@ -39,18 +39,15 @@ simpEvalX' ev ty = go
...
-             | otherwise
-             -> Nothing
+        -> case takePrimApps x' of
+             Just (prim, as)
+               | Just args <- mapM takeValue as
+               -> case simpEvalP ev ty a prim args of
+                   Just x''
+                     -> return x''
+                   _ -> XApp a <$> mp <*> mq
+             _ -> XApp a <$> mp <*> mq
```
amosr: I think you'll have to explain to me why this is faster: I can't see it. amazing

PR #398, opened 2016-05-18
separate PSV reading parts from Sea input

tranma:
I'm not sure how the binary reading code will look like yet, this PR is just picking the low-hanging fruits in separating PSV-specific code from the rest of Sea input. Will probably need to do more later.

So binary reading code would go in `Icicle.Sea.IO.Binary.Input`. It should define its specific fact value reading functions in the `SeaInput` record. See `Icicle.Sea.IO.Psv.Input` for an example.


tranma:
```
+data InputFormat
 +  = InputPsv Psv.PsvInputConfig Psv.PsvOutputConfig
 +  | InputEnterprise
```

jystic:
> InputEnterprise

lol

tranma:
oops, maybe the PsvOutputConfig shouldn't be inside InputFormat

src/Icicle/Sea/IO/Base/Input.hs
```diff
@@ -0,0 +1,339 @@
...
+
+     t
+      -> Left (SeaUnsupportedInputType t)
+
+--------------------------------------------------------------------------------
+
+type CStmt  = Doc -- lies
+type CBlock = Doc
+type CFun   = Doc
+type CName  = Doc
```
jystic: lol

jystic:
This is looking good, as you say, probably it will change a bit when we know what we want, but this is a much better starting point 👍

PR #403, opened 2016-05-25
Add new Prelude function to compute mode of stream

amosr:
```
# Group by given key, count, and find the most common element.
# If there are multiple elements with the maximum count we return one of those,
# but exactly which one is undefined and deterministic.
mode key
 = group fold (key', count') = (group key ~> count key)
 ~> max_by count' key'.

mode : Element a -> Aggregate Possibly a.
```

### DIFF
--- a/data/libs/prelude.icicle
+++ b/data/libs/prelude.icicle
@@ -120,7 +120,7 @@ numflips' v t
 # Group by given key, count, and find the most common element.
 # If there are multiple elements with the maximum count we return one of those,
 # but exactly which one is undefined and deterministic.
-mode key
+distribution_mode key
  = group fold (key', count') = (group key ~> count key)
  ~> max_by count' key'.
 
--- a/data/libs/prelude.icicle
+++ b/data/libs/prelude.icicle
@@ -116,6 +116,14 @@ numflips' v t
           end
  ~> case s | (s0,i) -> i end.
 
+
+# Group by given key, count, and find the most common element.
+# If there are multiple elements with the maximum count we return one of those,
+# but exactly which one is undefined and deterministic.
+mode key
+ = group fold (key', count') = (group key ~> count key)
+ ~> max_by count' key'.
+
 #######################
 ### Example queries ###
 #######################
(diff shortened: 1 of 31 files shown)


tranma:
hm, why is this the "mode" of stream?

amosr:
I think it's a common statistics term, though maybe it isn't a good name given our use of mode-al types

https://en.wikipedia.org/wiki/Mode_(statistics)

erikd-ambiata:
Not a blocker, but is it possible to write code to update the `test/.../expected` files at test time so they don't need to be checked into the repo? This would drasitcally reduce the noise in commits like this.

amosr:
it would probably be easier to just remove the "ok, loaded N functions" statement.
I might be misunderstanding, but the expected files are checked against each the actual test output: lazy testing with diff

tranma:
@amosr ah right, I did indeed think it had something to do with our modal types, was very confused :) Maybe call it distribution_mode, or probablity_mode, or something?

amosr:
yep, good plan. will rename. and change the example to be a bit better (it should be dividing by 10,000 instead of 1,000)

amosr:
renamed… @tranma

jystic:
I'm fairly sure your average Icicle user won't be versed in modal types and would expect `mean`, `median` and `mode` to be statistical functions. The name `mode` might be confusing for us, but we're not the target audience.

/cc @blever @markhibberd

amosr:
that is true. but does is hurt the average user to have a slightly longer name?

tranma:
💯 for distribution_mode

jystic:
Do you have something in mind that you would prefer to save `mode` for?

I think `mode` is better, but I don't really mind, so if you two feel strongly the other way then go ahead 👍

markhibberd:
> does is hurt the average user to have a slightly longer name?

I think it does when it is replacing a standard term from their domain with something else. I am very much on the side of using "mode", happy to talk about it, but we need to do a sanity check on the prelude etc... before we put this in hands of users anyway, so it isn't something we can't fix/change mind on (yet).

amosr:
ok, good point. `mode` it is

Issue #401, opened 2016-05-24
Stop ignoring namespaces


tranma:
Plumb the namespace through to output. Spec:

  * Each feature name still needs to be globally unique, i.e. this is illegal:

```
[feature.foo]
  namespace="A"
[feature.foo]
  namespace="B"
```

  * The feature name needs to be included in the output formation dictionary (this only affects dense PSV output), e.g.

```json
attributes":[{"name":"A:foo","encoding{"primitive":"double"},"index":1}]
```

  * Nothing changes for the actual snapshot output, e.g. we would still have

```
ID000|foo|0
```

@markhibberd

markhibberd:
💯

tranma:
this is done. Example:

```
title = "Dense input example dictionary"

version = 1

chapter = []

namespace = """default"""

tombstone = """NA"""

[feed.person]
  columns = ["misery", "state", "injury"]

[fact.person]
  [fact.person.encoding]
    misery="int"
    state="string"
    injury="(location:string,severity:double)"

[feature.newest_misery]
  namespace  = "miserables"
  expression = "feature person ~> newest misery"

[feature.mean_misery_by_state]
  expression = "feature person ~> group state ~> mean misery"

[feature.mean_misery_by_injury_location]
  namespace  = "warzone"
  expression = "feature person ~> group injury.location ~> mean misery"
```

output formation dictionary:
```
{"global_properties":{"missing_value":"NA"},"entity_id":{"encoding":"string","index":0},"attributes":[{"name":"warzone:mean_misery_by_injury_location","encoding":{"listof":{"pairof":[{"primitive":"string"},{"primitive":"double"}]}},"index":1},{"name":"default:mean_misery_by_state","encoding":{"listof":{"pairof":[{"primitive":"string"},{"primitive":"double"}]}},"index":2},{"name":"miserables:newest_misery","encoding":{"primitive":"int"},"index":3}],"version":"1","encoding_version":"1"}
```

markhibberd:
💯 Thanks.

PR #404, opened 2016-05-25
namespace plumbing

tranma:
as in https://github.com/ambiata/icicle/issues/401


amosr:
looks ok. an example with different namespaces wouldn't go astray
👍

amosr:
never mind example comment, just saw the one in ice

PR #405, opened 2016-05-30
fix off-by-one in PSV execution: was throwing away every 128th row

amosr:


tranma:
I thought this might have been introduced in recent commits but the check has always missed the set count: https://github.com/ambiata/icicle/blob/a4f68dde5061b6e33bf3e04063653e62b62cd69e/src/Icicle/Sea/Psv/Input.hs#L333

amosr:
hmm that's a bit scary

tranma:
🚀 it's a bit scary indeed

jystic:
💯

PR #407, opened 2016-06-07
Avalanche binary search map lookup

tranma:
Sorry for the big PR, a lot of it is tests.

- Bake a binary search into Avalanche flattening, used in `MapLookup` and `MapInsertOrUpdate`. When insert/update, if the key does not exist, put the key/val pair at the end of the two arrays and just call sort.
- Add an option to typecheck after each simpFlattened transform 
- Fix a name shadowing: simpFlattened used a fresh namer "simp", which shadows some existing variables. I think this is only a problem now because I generated a fresh Avalanche name and tried to pass it off as a Core name (in `MapInsertOrUpdate`: `flatX (xApp upd new_name)`)
- Make sure the while loop condition doesn't get dead-code'd


data/sea/30-array.h
```diff
@@ -203,15 +203,16 @@ static ARRAY_T(t) INLINE ARRAY_FUN(t,put_immutable)
...
-      x = arr;                                                                  \
+        ARRAY_T(t) arr = (ARRAY_T(t))imempool_alloc(pool, bytes_old);           \
+        memcpy(arr, x, bytes_old);                                              \
+        x = arr;                                                                \
+        x->count = ix + 1;                                                      \
                                                                                 \
     }                                                                           \
                                                                                 \
-    x->count = ix + 1;                                                          \
     ARRAY_PAYLOAD(t,x)[ix] = v;                                                 \
```
amosr: this should allocate and memcpy regardless, it shouldn't modify the existing array

main/icicle-repl.hs
```diff
@@ -86,49 +89,53 @@ runRepl inits
...
+   , hasAvalancheEval   :: Bool
+   , hasFlattenSimp     :: Bool
+   , hasFlattenNoSimp   :: Bool
+   , hasSeaPreamble     :: Bool
+   , hasSea             :: Bool
+   , hasSeaAssembly     :: Bool
+   , hasSeaLLVMIR       :: Bool
+   , hasSeaEval         :: Bool
+   , doFlattenSimpCheck :: Bool
+   , doCoreSimp         :: Bool }
```
jystic: Maybe one day we should make this a `Data.Set` of enabled flags

src/Icicle/Avalanche/Simp.hs
```diff
@@ -107,3 +187,9 @@ simpFlattened a_fresh p
    -- finish a-normalisation by taking lets from inside expressions to statements.
    >>= return . pullLets
 
+  constantFold
+   = transformX return (return . simpEvalX Flat.evalPrim Flat.typeOfPrim)
+
+  ctx    = Check.initialContext p
+  check1 = check (optCheck (simpOptsCheckSimp   opts) ctx)
+  check2 = check (optCheck (simpOptsCheckCrunch opts) ctx)
```
jystic: woah decimal counting system :nextlevel:

test/cli/repl/t31-builtin/expected
```diff
@@ -22,7 +22,7 @@ ok, loaded test/cli/repl/data.psv, 13 rows
                                    \             (
                                     \_____________)
 > > - C evaluation:
-[homer, ["head","arm","torso"]]
+[homer, ["arm","head","torso"]]
```
jystic: 👏

src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
```diff
@@ -0,0 +1,747 @@
...
+initBool acc init
+  = InitAccumulator (Accumulator acc BoolT init)
+
+readBool :: Name n -> Name n -> Statement a n p -> Statement a n  p
+readBool local acc
+  = Read local acc BoolT
+
+-- Create a let binding with a fresh name
+slet_
+  :: (Hashable n, Monad m)
```
jystic: `::` on a different line to `slet_` makes grep a sad panda

jystic:
`Icicle.Avalanche.Statement.Flatten.Algorithms` is very clean code 👍

data/sea/30-array.h
jystic: I think we've always modified the existing array?

amosr: ah, yeah, so there are two versions now: put mutable and put immutable. put immutable is very rarely used, and most cases are optimised to mutable ones. the immutable versions were added because we now expose map and array update from the source language, so some of those uses end up being immutable.
we probably haven't actually encountered any cases where it would end up using this though..

src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
```diff
@@ -0,0 +1,747 @@
...
+--
+-- binary_search key array
+--   found = false
+--   m = -1
+--
+--   low  = 0
+--   high = size array - 1
+--   end = false
+--
+--   while end == false
```
amosr: what is the loop invariant? tee hee hee.

src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
```diff
@@ -0,0 +1,747 @@
...
+  -> FlatM a n
+avalancheMapInsertUpdate a_fresh stm flatX tk tv upd ins key map
+  = flatX' map $ \map' -> flatX' key $ \key' -> do
+      n_acc_keys <- freshPrefix "map_insert_acc_keys"
+      n_acc_vals <- freshPrefix "map_insert_acc_vals"
+      n_acc_idx  <- freshPrefix "map_insert_acc_bs_index"
+      n_loc_keys <- freshPrefix "map_insert_loc_keys"
+      n_loc_vals <- freshPrefix "map_insert_loc_vals"
+      n_loc_idx  <- freshPrefix "map_insert_loc_bs_index"
+      --n_loc_new  <- freshPrefix "map_insert_loc_new"
```
jystic: cruft?

src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
```diff
@@ -0,0 +1,747 @@
...
+               [ Write   n_acc_mid (xMid v_low v_high)
+               , readInt n_loc_mid  n_acc_mid
+               $ Let  n_loc_x   (xIndex v_array v_mid)
+               $ If ( relEq t v_x v_key       )
+                    ( Block [ Write n_acc_end   xTrue
+                            , Write n_acc_found xTrue ])
+               $ If ( relLt t v_x v_key )
+                    ( Write n_acc_low  (xPlusOne  v_mid) )
+                    ( Write n_acc_high (xMinusOne v_mid) )
+               ]
```
amosr: this looks good

amosr:
super cool!

data/sea/30-array.h
tranma: yeah, I need to increment the count for the new array

src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
tranma: yeah just don't want the type signature to lean too much to the right. Does something along the lines of `grep -e "slet_$^ .::"` work?

jystic: possibly, but I have a hotkey setup to grep word under cursor, I guess I could bind something separate which searches for only type signatures

jystic: yeah that doesn't seem to work, it's single line only :(

tranma: hm, damn. I guess I'll just have to live with right-leaning type signatures. Haskell tooling is so bad

data/sea/30-array.h
amosr: I don't understand what you mean. the immutable version should allocate a new array and copy the results in regardless, because at the moment it's mutating the input array if the index is already in the array

tranma: sorry, that's what I meant.

src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
tranma: something like, `forall i. A[0]...A[low] <= key, key <= A[high]...A[n-1]`?

data/sea/30-array.h
```diff
@@ -192,28 +192,9 @@ Immutable_put(arr, ix, val)
...
-                                                                                \
-    }                                                                           \
-                                                                                \
-    x->count = ix + 1;                                                          \
-    ARRAY_PAYLOAD(t,x)[ix] = v;                                                 \
-    return x;                                                                   \
+    ARRAY_T(t) arr = ARRAY_FUN(t,copy)(pool, x);                                \
+    arr = ARRAY_FUN(t,put_mutable)(pool, arr, ix, v);                           \
+    return arr;                                                                 \
 }
```
tranma: @amosr I should have just done this from the beginning, sorry! ^

amosr: oh yeah! cool 👍

tranma:
gonna merge this, that failed boris build seems unrelated

PR #408, opened 2016-06-07
Crazy Avalanche EDSL

amosr:
sketch of an idea that might be useful for writing Avalanche programs @tranma @jystic

### DIFF
--- a/src/Icicle/Avalanche/Statement/Edsl.hs
+++ b/src/Icicle/Avalanche/Statement/Edsl.hs
@@ -1,0 +1,150 @@
+-- | Crazy monadic combinators for writing Avalanche programs
+
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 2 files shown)


tranma:
avalanche snowballing!

src/Icicle/Avalanche/Statement/Edsl.hs
```diff
@@ -0,0 +1,150 @@
...
+ = X (Exp () n Prim)
+
+unsafeX :: Exp a n Prim -> X n t
+unsafeX = X . reannotX (const ())
+
+
+-- ----------------------------------------
+-- Monads!
+-- The actual magic is a fresh monad with a continuation of what to do with the *rest* of the statement.
+-- ----------------------------------------
```
tranma: this replaces passing the `stm` function around, right?

amosr: oh, I hadn't thought that far ahead. I was thinking it would just be useful for the definitions of sort and binary search

PR #410, opened 2016-06-08
map_insert without sort

tranma:
@amosr I think the reason it was slow was because map_update used array_put_immutable rather than array_put_mutable. By binding a new name here, the mutation recovery can change it to a put_mutable. Do you think it's better if the Avalanche programmer uses mutable/immutable explicitly?

```
      if (map_insert_loc_bs_index$flat$11$simpflat$40)
      {
        read update_acc$flat$27 = map_insert_acc_vals$flat$7 [Array String];
        let simpflat$66 = Array_length#@{String} update_acc$flat$27;
        -- this was previously a put_immutable
        write map_insert_acc_vals$flat$7 = Array_put_mutable#@{String} update_acc$flat$27 simpflat$66 ("foo"@{String});
      }
```

### DIFF
--- a/src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
+++ b/src/Icicle/Avalanche/Statement/Flatten/Algorithms.hs
@@ -83,8 +83,14 @@ forI :: (Hashable n, Monad m, IsString n)
      => a -> Exp a n p -> (Exp a n p -> FreshT n m (Statement a n p))
      -> FreshT n m (Statement a n p)
 forI a_fresh to ss
+ = forI_ a_fresh ForeachStepUp (XValue a_fresh IntT (VInt 0)) to ss
+
+forI_ :: (Hashable n, Monad m, IsString n)
+     => a -> ForeachType -> Exp a n p -> Exp a n p -> (Exp a n p -> FreshT n m (Statement a n p))
+     -> FreshT n m (Statement a n p)
+forI_ a_fresh t from to ss
  = do n  <- freshPrefix "for_counter"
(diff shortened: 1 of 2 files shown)


jystic:
> Do you think it's better if the Avalanche programmer uses mutable/immutable explicitly?

"Avalanche programmer" 😆

amosr:
I'm not sure. doesn't it depend on how/when map_insert is used in the source program, whether it should be a put_mutable or immutable?

tranma:
@amosr I made map_insert inserts into the appropriate index instead of inserting at the end, then sort:

```
./dist/build/icicle-bench/icicle-bench slowmaps.toml inc.psv foo.psv foo.c 2016-01-01     
icicle-bench: starting compilation
icicle-bench: compilation time = 0.87s
icicle-bench: starting snapshot
icicle-bench: snapshot time   = 0.04s
icicle-bench: total entities  = 50
icicle-bench: total facts     = 499950
icicle-bench: fact throughput = 13.07 million facts/s
icicle-bench: byte throughput = 634.14 MB/s
```

re. array put mutable/immutable, you're right, the mutation recovery is a bit broken :( I'll have another look at it

amosr:
those numbers look very good to me. I guess this data is the 'best case' for the shuffling along, but in real data the insert should be rather rare anyway.
👍

amosr:
I'll try on some larger datasets, and ones going in the reverse order, but it might not be an issue if mutation recovery is incomplete, for now

tranma:
@amosr shall I merge this? Haven't tested linear insert vs. this binary insert for large data sets yet, for small data the gains are pretty minimal. Anyhow it's better than the first attempt using push + sort.

amosr:
yes, merge it. 👍

PR #411, opened 2016-06-08
Fix tiny leak in mempool: iblock_free was not freeing block itself

amosr:


jystic:
Looks legit 👍 

How did you find this?

amosr:
purely by accident: I just noticed it when I was trying to figure out how to write that debug function. it would have taken a long time before it actually caused any issues

PR #412, opened 2016-06-09
Add flattened Avalanche mutation recovery test

tranma:
This should technically be already covered by the simp_flattened tests. Just making sure we hit array update cases more often.
I don't know how to test for completeness so safety would have to do for now.


amosr:
who could say no to more tests
🔭

amosr:
I don't know how to test for completeness either, without either reimplementing the whole thing, or trying to solve something undecidable

PR #414, opened 2016-06-16
Limit the number of facts per entity-attribute

tranma:
- This will drop facts when the count exceeds a limit per entity-attribute. In this PR the already computed values stay, it just doesn't read any more facts after the limit. In the next PR I'll drop these results from the output as well.
- The limit is global for now, but can be made to vary for each attribute.
- The dropped entity/attribute are recorded:  `<output-dir>/drop_001.txt` contains the dropped entity/attributes from output `<output-dir>/part_001.psv`


data/sea/52-psv.h
```diff
@@ -107,7 +123,72 @@ static int INLINE psv_compare (const char *xs, size_t xs_size, const char *ys, s
...
+
+        n_ptr = memchr (line_ptr, '\n', bytes_remaining);
+
+        if (n_ptr == 0) {
+            s->input_remaining = bytes_remaining;
+            return 1;
+        }
+
+        const char  *entity_ptr = line_ptr;
+        const int    new_entity = psv_compare (entity_cur, entity_cur_size, entity_ptr, bytes_remaining);
```
amosr: I believe entity_cur contains the pipe after the entity name, so this will still count an entity 'ABCD' after 'AB' as a new entity - does that sound right?
might be worth writing a test for that case, perhaps just a bash script that runs bench with the max count set to 2?

amosr: does what I said make sense? a test something like:

in.psv
```
AB|attr|v1|2000-01-01
AB|attr|v2|2000-01-01
ABCD|attr|v3|2000-01-01
```

dict.toml:
```
feature attr ~> latest value
```

run and check:
```
> bench --max-count 2
throw away AB
ABCD|v3
```

data/sea/52-psv.h
```diff
@@ -107,7 +123,72 @@ static int INLINE psv_compare (const char *xs, size_t xs_size, const char *ys, s
...
+
+    iint_t  fact_count      = s->fact_count;
+
+    for (;;) {
+        const size_t bytes_remaining = end_ptr - line_ptr;
+
+        n_ptr = memchr (line_ptr, '\n', bytes_remaining);
+
+        if (n_ptr == 0) {
+            s->input_remaining = bytes_remaining;
```
amosr: should this read a new block from the file if there is no `\n`?

data/sea/52-psv.h
```diff
@@ -107,7 +123,72 @@ static int INLINE psv_compare (const char *xs, size_t xs_size, const char *ys, s
...
+
+            return 1;
+        } else {
+            fact_count++;
+        }
+
+        line_ptr = n_ptr + 1;
+    }
+
+    s->fact_count = fact_count;
```
amosr: unreachable code?

data/sea/52-psv.h
amosr: ahh, I think I understand now. this will return, and psv_snapshot will loop and read more of the buffer, so if the next one has the same entity it will call this again. cool!

amosr:
some more comments might help to explain the need for the is_same check and the entity_dropped field, but it looks good
🔮

data/sea/52-psv.h
tranma: yeh I did try this out in ice, but hasn't gotten around to writing a cli test, and haven't tried this case. Will do

data/sea/52-psv.h
tranma: yeah it's a bit savage because it relies on the specific control flow of psv_snapshot, this is just the easiest way

amosr: it's good, it just took me a while to realise what was happening

src/Icicle/Sea/IO/Psv/Input.hs
```diff
@@ -224,20 +231,31 @@ seaOfReadAnyFactPsv opts config states = do
...
+                , "   , curr_time_ptr"
+                , "   , last_time_size"
+                , "   , last_time_ptr );"
+                ])
+                ( vsep
+                [ "return ierror_loc_tag_format"
+                , "   ( IERROR_LIMIT_EXCEEDED"
+                , "   , entity_ptr + entity_size"
+                , "   , entity_ptr"
+                , "   , \"%.*s - %.*s facts were dropped starting from here\\n\""
```
jystic: This doesn't really say why they were dropped, not sure what to say instead, maybe something like:
`$entity contained an unreasonable number of $attribute facts, they have been dropped`

tranma:
@amosr @jystic mind having another look at this? see the bench tests for an example

data/sea/52-psv.h
```diff
@@ -107,7 +122,71 @@ static int INLINE psv_compare (const char *xs, size_t xs_size, const char *ys, s
...
+                                          , s->entity_dropped, s->entity_dropped_size));
+
+    if (new_dropped) {
+        ierror_msg_t msg           = ierror_loc_pretty (error, s->fact_count);
+        size_t       bytes_written = write (drop_fd, msg, error_msg_size);
+
+        if (bytes_written == 0)
+          return error;
+
+        s->entity_dropped = calloc (s->entity_cur_size, 1);
```
jystic: Where does this memory get freed? (couldn't see it)

tranma: oops, it doesn't :) thanks

tranma:
gonna merge this when it builds

data/sea/52-psv.h
```diff
@@ -449,6 +576,8 @@ void psv_snapshot (psv_config_t *cfg)
     free (input_ptr);
     free (entity_cur);
     free (output_ptr);
+
+    if (!state.entity_dropped) free (state.entity_dropped);
```
amosr: should that `!` be there?

tranma: oh right, I can't C :( sorry

data/sea/52-psv.h
```diff
@@ -107,7 +122,75 @@ static int INLINE psv_compare (const char *xs, size_t xs_size, const char *ys, s
...
+        const int    new_entity  =  (drop_size != entity_size)
+                                 || (psv_compare ( drop_me, drop_size
+                                                 , entity_ptr, entity_size ) != 0);
+
+        if (new_entity) {
+            /* start the input buffer here */
+            return 0;
+        } else {
+            /* just skip this fact */
+            line_ptr = n_ptr + 1;
```
amosr: would it be a big change to count the number of dropped facts, and output that to the drop file at the end? just an idea

tranma: number of dropped facts per entity-attribute?

amosr: that would be neat - or a total would be fine too. I don't know, it doesn't really matter

tranma: sounds good esp for debugging, I'll add it

amosr: whatever is easiest

Issue #138, opened 2015-10-14
Inlining causes side effects


HuwCampbell:
If there's a context in an inline in a function, the results can be somewhat unexpected. With function:
`filterSome a b = filter (case a | Some a' -> True | None -> False end) ~> b.`
One would expect that b would be evaluated after the filter, when we desugar:

```
> feature injury ~> filterSome admitted (latest 4 ~> admitted)
- Desugar:
feature injury
~> let a$inline$0 = admitted
~> let b$inline$1 = (latest 4 ~> admitted)
~> filter case a$inline$0
         | (Some desugar$0) -> True
         | None -> False
       end
~> b$inline$1
```

HuwCampbell:
This is probably more a question of how to include and use contexts within the prelude, which is at the moment quite difficult.

amosr:
this is very important, and I have this sinking feeling that the answer is going to involve "we need higher order functions" because I imagine both behaviours will be desirable, but I think we can work around it for now.

obvious next question: what if an argument were used in two contexts, could this cause duplicate computation?
```
filter2 v f = (filter v > 50 ~> f), (filter v > 100 ~> f)
```

```
feature salary ~> filter value (really_expensive_sum value)
```
but maybe that duplication is actually what is meant. ooh

amosr:
this is sort of two things: we have an implicit stream threaded through with ``~>``, as well as the lack of higher order functions.

this program:
```
feature salary
~> let x = sum value
~> filter pred
~> let y = sum value
~> x, y
```
actually means something like
```
s1  = salary
x   = sum (map value s1)
s2  = filter pred s1
y   = sum (map value s2)
return (x, y)
```

and, when functions are inlined, we introduce lets to avoid duplication of work. I guess we could just remove those lets

amosr:
I have come around to changing the behaviour of function application to duplicate computation à la call-by-name, since our let bindings are 'strict' and we can simulate call-by-value by adding let bindings.
this *is* a change in the semantics, but should affect very few programs; only those with functions with filters and groups.
I will coordinate with @jystic to make this change at the same time as the syntax overhaul.

amosr:
this is the default now

PR #416, opened 2016-06-18
Safe substitution on Source expressions. Fix substitution.

amosr:


tranma:
does this fix https://github.com/ambiata/icicle/issues/138?

amosr:
@tranma not yet, but it is the first step to implementing it. I will add an option to `inlineTransform` which is whether to add lets or substitute directly, which should be fairly simple to do

jystic:
😋 looks good

amosr:
will merge & implement inlineTransform change as separate PR

tranma:
wow so this substitution code works first try, great job @amosr 👍 👏 👏 👏

jystic:
💯

PR #417, opened 2016-06-21
Don't write partial results for dropped facts to regular output

tranma:
this is pretty boring. Just writing the results when we drop facts to the drop log rather than normal output.

e.g.

```
line 2: bar|salary|100|1989-12-18
        ^~~~ entity bar had too many salary facts, they have been dropped starting from here
bar|newest_salary|5
total: 1 facts for bar were dropped.

line 4: foo|salary|10|1989-12-18
        ^~~~ entity foo had too many salary facts, they have been dropped starting from here
foo|newest_salary|0
total: 3 facts for foo were dropped.

line 6: foooo|salary|40|1989-12-20
        ^~~~~~ entity foooo had too many salary facts, they have been dropped starting from here
foooo|newest_salary|20
total: 2 facts for foooo were dropped.
```


data/sea/52-psv.h
```diff
@@ -20,7 +20,8 @@ typedef struct {
     iint_t       entity_count;
 
     /* stuff */
-    const size_t psv_max_ent_attr_count;
+    const size_t facts_limit;   /* maximum number of facts per entity-attribute */
+    const int    discard_limit; /* whether to write discarded facts to output or drop log */
```
amosr: `ibool_t`?

amosr:
this is great. looks like all the information you could want. does it work with dense too?

test/cli/bench/t2a-expected.psv
```diff
@@ -1,3 +0,0 @@
-bar|newest_salary|5
-foo|newest_salary|0
-foooo|newest_salary|20
```
jystic: Is it expected that there are no entities here?

amosr: I think "bar" and "foo" are the entities?

jystic: just that it's highlighted red, is that being removed by this PR?

tranma: sorry, it's out of date. Updated now.

tranma:
It only works with dense PSV in the sense that dense facts are considered as one whole struct, i.e. it can't count the dense fields separately

amosr:
oh right, that sounds like the only reasonable thing for dense

test/cli/bench/run
```diff
@@ -6,8 +6,8 @@ t1_dict=test/cli/bench/t1-dictionary.toml
 t1_in=test/cli/bench/t1-in.psv
 
 t1_expected=test/cli/bench/t1-expected.psv
-t1_out_c=`mktemp -t icicle-bench-t1-c`
-t1_out_psv=`mktemp -t icicle-bench-t1-out`
+t1_out_c=`mktemp -t icicle-bench-t1-c-XXXXXX`
+t1_out_psv=`mktemp -t icicle-bench-t1-out-XXXXXX`
```
jystic: mktemp is different on os/x / linux, fwiw mafia does this https://github.com/ambiata/mafia/blob/master/script/mafia#L21

tranma:
shall I merge this? It has a dropping test for both sparse/dense psv by the way.

jystic:
Yeah go for it 👍

PR #418, opened 2016-06-22
minor fixes

tranma:
Sorry, made a late change last PR and didn't notice it broke.

### DIFF
--- a/data/sea/52-psv.h
+++ b/data/sea/52-psv.h
@@ -106,10 +106,18 @@ static ierror_msg_t INLINE psv_write_outputs
   , size_t entity_size
   , ifleet_t *fleet );
 
+static ierror_msg_t INLINE psv_write_to_output
+  ( psv_state_t *s
+  , const char *entity
+  , const size_t entity_size );
+
 static ierror_msg_t INLINE psv_write_to_drop
   ( psv_state_t *s
   , const char *entity
   , const size_t entity_size );
+
+static ierror_msg_t INLINE psv_flush_to_output
+  ( psv_state_t *s );
 
 static ierror_msg_t INLINE psv_flush_to_drop
   ( psv_state_t *s );
@@ -336,6 +344,10 @@ static ierror_loc_t psv_read_buffer (psv_state_t *s, const size_t facts_limit)
                                      , entity_cur
                                      , entity_cur_size
                                      , s->fleet );
+                /* we need to flush the output immediately because it might be the case that
+                   the result for the new entity will need to be flushed to drop_fd, while the output
+                   for the old entity here needs to be flushed to output_fd as usual. */
+                msg = psv_flush_to_output (s);
 
                 if (error) {
                     error = ierror_loc_format (0, 0, "%s", msg);
@@ -462,7 +474,7 @@ static ierror_msg_t INLINE psv_flush_to_output (psv_state_t *s)
 
 static ierror_msg_t INLINE psv_flush_to_drop (psv_state_t *s)
 {
-  return psv_output_flush (s->output_fd, s->output_start, &s->output_ptr);
+  return psv_output_flush (s->drop_fd, s->output_start, &s->output_ptr);
 }
 
 #define ENSURE_SIZE(bytes_required)                                   \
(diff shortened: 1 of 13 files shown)


data/sea/52-psv.h
```diff
@@ -106,11 +106,19 @@ static ierror_msg_t INLINE psv_write_outputs
...
+  ( psv_state_t *s
+  , const char *entity
+  , const size_t entity_size );
+
 static ierror_msg_t INLINE psv_write_to_drop
   ( psv_state_t *s
   , const char *entity
   , const size_t entity_size );
 
+static ierror_msg_t INLINE psv_flush_to_output
```
amosr: I don't really understand why this returns a message?

amosr: oh, never mind me…

tranma: it's just when the call to `write` fails for whatever reason

data/sea/52-psv.h
```diff
@@ -336,6 +344,10 @@ static ierror_loc_t psv_read_buffer (psv_state_t *s, const size_t facts_limit)
                                      , entity_cur
                                      , entity_cur_size
                                      , s->fleet );
+                /* we need to flush the output immediately because it might be the case that
+                   the result for the new entity will need to be flushed to drop_fd, while the output
+                   for the old entity here needs to be flushed to output_fd as usual. */
+                msg = psv_flush_to_output (s);
```
amosr: should this and the call to `psv_write_outputs` above be assigning to `error` rather than `msg`? and should this only occur if msg/error is null?

tranma: yeah good catch, should check `msg` for failure here, not `error`

amosr:
so this is flushing the output buffer for every new entity? that sounds fine to me, @jystic you don't think that will introduce performance problems will it?

tranma:
yeah it's flushing every single time it sees a new entity

tranma:
So I make the Sea pretty printer prints out NaN and Infinity, but there only seem to be NaNs:

```
> feature person ~> newest (1/0)
- C evaluation:
[homer, NaN,marge, NaN]

- Core evaluation:
[homer, Infinity,marge, Infinity]

> feature person ~> newest (sqrt(-1))
- C evaluation:
[homer, NaN,marge, NaN]

- Core evaluation:
[homer, NaN,marge, NaN]
```

Not sure if this is ok.

src/Icicle/Sea/FromAvalanche/Program.hs
```diff
@@ -252,6 +265,10 @@ seaOfXValue v t
  where
   prim p args
    = seaOfPrimDocApps (seaOfXPrim p) args
+  nan
+   = "0.0/0.0"
+  inf
+   = "sqrt(-1.0)" -- assuming math.h
```
amosr: haha I don't know about this…

amosr: add some static constants to one of the sea files?

HuwCampbell: Aren't `INFINITY` and `NAN` `math.h`?

HuwCampbell: Maybe it depends on the implementation...

amosr: yeah, surely in there somewhere

tranma:
@amosr Core simpX had the same problem as Avalanche simpEvalX, fixed here: https://github.com/ambiata/icicle/pull/393

The fix is identical to that PR. It brought compilation time for our big dictionary from half hour to about 5m, most of which is spent in clang.

amosr:
fantastic! 👍 💯 :1000:

src/Icicle/Sea/FromAvalanche/Program.hs
tranma: yeah gonna assume math.h has INFINITY and NAN

olorin: C99 requires INFINITY; it does not require NAN.

PR #421, opened 2016-06-27
boris.toml / master.toml

jystic:
/cc @nhibberd @amosr @tranma


nhibberd:
looks good :+1:

master.toml
```diff
@@ -1,38 +1,51 @@
...
 
 [build.appshop]
   MAFIA_ALEX = "true"
   MAFIA_HAPPY = "true"
-  PUBLISH_APPSHOP="icicle-repl"
+  PUBLISH_APPSHOP = "icicle-repl"
 
 [build.appshop.master]
-  MAFIA_ALEX = "true"
-  MAFIA_HAPPY = "true"
```
tranma: we don't need this?

jystic: Not in this table, it goes in the `build.appshop` table

PR #422, opened 2016-06-27
Minor yak shaving: clean up compiler pipeline

tranma:
More yak shaving on `Pipeline`. It should be easier to find what we need and not write redundant compilation functions now?

I think the type aliases in Pipeline helps.

### DIFF
--- a/src/Icicle/Benchmark.hs
+++ b/src/Icicle/Benchmark.hs
@@ -97,7 +97,9 @@ createBenchmark mode dictionaryPath inputPath outputPath dropPath packedChordPat
   (dictionary, input') <- firstEitherT BenchDictionaryImportError $ inputCfg input
   let output'           = outputCfg output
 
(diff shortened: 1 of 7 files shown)


jystic:
`IsName` is cool but not sure if I like the other ones, I guess there's a naming system so you can predict what the type would be?

tranma:
I think at least having aliases for the two annotation types reduce the noise. The other aliases might be unnecessary, they can go

jystic:
I'm kind of on the fence, have had bad experiences with lots of aliases (trying to figure out what the real types are), I don't mind if you think it's good

tranma:
hm I think `*_Type` and `*_Unit` as shorthands for `Blah (Common.Annot ()) foo var` and `Blah () foo bar` is good, but the others can probably go. Don't really need an alias to tell you that `Program Flat.Prim` is using Flat primitives

jystic:
Cool 🚎

PR #425, opened 2016-07-06
Lexer / Parser / Pretty for Sorbet (the new syntax)

jystic:
The is the first step towards the new syntax which looks a lot more like the syntax presented in Amos and Ben's paper and in addition has whitespace sensitive case expressions and let blocks.

To get an idea of what it looks like I've converted the prelude and put it in a new file `prelude-new.icicle`.

I started by documenting the existing grammar (in `icicle-v0.bnf`) and used that as a basis for the sorbet grammar (in `sorbet.bnf`).

The concrete syntax types match the grammar exactly 1:1 so you can read them as the grammar itself. This makes the parser really easy, it's just applicative over each constructors. The plan is to have an elaboration phase which converts the concrete syntax to the abstract syntax (i.e. the current icicle source AST).

/cc @amosr @tranma @markhibberd


doc/internals/icicle-v0.bnf
```diff
@@ -0,0 +1,304 @@
...
+                |  <digit>
+                |  _
+                |  $
+                |  '
+<var-tails>    ::= <var-tails> <var-tail>
+                |  ε
+
+<con-id>       ::= <con-head> <con-tails>
+<con-head>     ::= <upper>
+                |  _
```
amosr: I don't think constructors can start with underscore, and I guess single "_" case is just a var

jystic: This is the v0 (current) syntax where I think this is actually the case, the new syntax is in `doc/internals/sorbet.bnf`

src/Icicle/Sorbet/Concrete/Annotation.hs
```diff
@@ -0,0 +1,237 @@
...
+  , annotOfConId
+  ) where
+
+import           Icicle.Sorbet.Concrete.Syntax
+
+
+annotOfRepl :: Repl a -> a
+annotOfRepl = \case
+  From a _ _ ->
+    a
```
amosr: are annotations comonads?

jystic: Discussed IRL, looks like they are!

doc/internals/sorbet.bnf
```diff
@@ -0,0 +1,369 @@
...
+                    |  of
+                    |  if
+                    |  then
+                    |  else
+                    |  for
+                    |  fold
+                    |  fold1
+                    |  window
+                    |  group
+                    |  fold
```
HuwCampbell: fold is mentioned twice

amosr:
🍧 so good

doc/internals/sorbet.bnf
jystic: ta, fixed

PR #427, opened 2016-07-18
Split out icicle-data, icicle-source, icicle-core

tranma:
Crude split into a few subprojects. This is already too big, more PRs will follow.

- icicle-data: common, data definitions
- icicle-source: source language parser, type checker, transforms, and conversion to core
- icicle-core: core exp, stream, program definitions, and transforms
- icicle-compiler: the rest of the icicle blob

@amosr @jystic


amosr:
looks good 👞

Issue #424, opened 2016-07-05
Discuss splitting into multiple modules.


markhibberd:
It would be nice to split out source level parsing / type-checking without some of the dependencies on c/code-gen. Case I need is embedding it in an upstream service where I want to be able to parse, typecheck, validate and pretty print for online editing of expressions. Not sure how possible this is or how much impact it would have, but it would make a big difference to dev'ing other projects.

markhibberd:
@jystic @tranma @amosr

tranma:
This shouldn't be too bad, splitting into something like: `icicle-source`, `icicle-core`, `icicle-avalanche`, `icicle-sea` (DDC style) sounds good to me.

amosr:
it sounds relatively straightforward.
if we rename `Source.ToCore.*` into `Core.FromSource`, we should be able to split out to a few projects: `icicle-common` (`Icicle.Common.*`, `Icicle.Data.*`), `icicle-source` (`Icicle.Source.*`, `Icicle.Storage.*`) and `icicle-meat` (everything else, could be divided further).

then your case would just rely on `icicle-common` and `icicle-source`, and I think the dependencies for that would end up as something like this. is `jetski` your main concern?
```
                       base
                     , ambiata-p
                     , ambiata-x-eithert
                     , annotated-wl-pprint
                     , bifunctors
                     , containers 
                     , dates -- not sure about this one
                     , deepseq
                     , hashable
                     , lens -- damn
                     , scientific
                     , text -- probably, not certain
                     , thyme
                     , time -- not sure
                     , time-locale-compat -- not sure
                     , transformers
                     , unordered-containers
                     , validation -- hmm
                     , vector
                     , vector-space -- used one place
                     , parsec
```

amosr:
actually maybe it's better off leaving it as-is and not moving `Source.ToCore`, since the core language shouldn't really rely on the source language should it?

tranma:
- while we are at it, splitting out the TOML stuff into a separate package would eliminate many of those questionable dependencies from the other packages.
- I think it makes sense for "downstream" languages to have a dependency on higher level ones.

amosr:
ok, so talking to jacob, we agreed to leave `Icicle.Source.ToCore` where it was, and have `icicle-source` depend on `icicle-core` rather than the other way around

amosr:
I believe this is done

PR #428, opened 2016-07-21
Split out icicle-repl & Add icicle-repl-source

tranma:
The `icicle-repl-source` REPL only typechecks Source expressions, which should fit the use case in https://github.com/ambiata/icicle/issues/424

@amosr @jystic


tranma:
packages so far

- `icicle-data`: common
- `icicle-source`: source, encoding
- `icicle-core`: core defs and transforms
- `icicle-compiler`: rest of the blob
- `icicle-repl-data`: common repl bits, mostly terminal fiddling
- `icicle-repl-source`: source repl
- `icicle-repl`: same old repl

amosr:
looks groovy 💃

jystic:
wondering if we need the `icicle-repl-source` executable? reading https://github.com/ambiata/icicle/issues/424 it seems as though @markhibberd just needs a library? /notsure

tranma:
yeh a library is probably enough, but a REPL might make it easier esp. if they want to use it in scripts /notsure
happy to remove it if Mark only needs a library

jystic:
I don't mind, although better to have less code if we can manage it I reckon

tranma:
removed the source REPL for now

PR #430, opened 2016-08-08
Fact modes

tranma:
* Fact ingestion now distinguishes between fact modes.
* This doesn't affect compute in anyway.
* For sparse states mode, a running state is kept and updated with each new fact.
* At the start of a window, a fact state has the value it did before the window.
* If there is no state update within the window, a fact state still has the value it did before the window.

(see the bench and REPL tests for examples)

@amosr @jystic


amosr:
looks good to me. 👍 
how is it working on icicle now that it's split up - is it any more annoying to build?

HuwCampbell:
What happens if a new sparse state is ingested before the starting edge of the window after a snapshot has been taken?

jystic:
do the existing psv property tests cover the modes?

tranma:
> What happens if a new sparse state is ingested before the starting edge of the window after a snapshot has been taken?

you mean like in big data mode? This is only in small data mode, so it assumes data is given since he start of time, regardless of how big the window is. I'm not too sure how big data mode should handle this fact mode.

> do the existing psv property tests cover the modes?

it does generate random modes

> is it any more annoying to build?

more annoying to build the REPL :(

jystic:
Cool, looking good 👍

HuwCampbell:
> you mean like in big data mode? This is only in small data mode, so it assumes data is given since the start of time, regardless of how big the window is. I'm not too sure how big data mode should handle this fact mode.

Yeah, because the fact state saved in `seaOfSaveFactState` doesn't appear to be commutative.

tranma:
@HuwCampbell It's not commutative, but I'm not sure if "commutative state updates" (for sparse state facts) makes sense. I'll write this up in the doc but not sure how to handle it here.

doc/user/modes.md
```diff
@@ -0,0 +1,27 @@
...
+This is the default fact mode. Each fact is ingested as a single value.
+
+Sparse States
+-------------
+
+Facts are treated as state updates. A running state is kept and updated with each new fact.
+
+### Notes
+
+1. Number of facts: one extra fact with the current fact state is inserted with the snapshot/chord date.
```
markhibberd: I actually think it needs to keep its original date as well (or only the original date).

markhibberd:
@tranma @jystic Semantics/examples all look pretty much right to me, there are two things we should talk through though:
 - mode being an artefact of the fact and not the feature, this isn't something that clicked for me until I walked through this code, it may not be avoidable but would want to talk it through a bit more so I understand limitations
 - semantics of bubblegum / saving-state / etc... there are some subtleties here that probably need to be worked out/captured that aren't at the moment, I realise I probably left out some of the edge cases in previous explanation, but i hope they won't be too hard to sort out.

tranma:
yeah fair enough, I didn't really think about big data mode until Huw pointed it out. I'll write up the doc with more details and examples, then catch up IRL at some stage?

doc/user/modes.md
tranma: if it only kept the original date, it won't be counted in the window. So if the input only has:

```
bulbasaur|pokemon|{"location":"pallet"}|2014-01-01
```

and we ask for `feature pokemon ~> windowed  2 weeks ~> newest fields` at snapshot date `2016-1-1`, we will get nothing for bulbasaur.

On the other hand, if we add an extra fact at `2016-1-1`, we will get bulbasaur still being at pallet, which sounds right to me.

tranma:
Superseded by facts refutation, in a different PR.

PR #433, opened 2016-08-26
Crude stab at a refute-by for facts

tranma:
For now we will associate a "nub-by" key with each fact, rather than with features. The key is any Icicle expression e.g. `key = "time"`

```
bulbasaur|pokemon_id|1|2014-01-01
bulbasaur|pokemon_id|3|2014-01-01
bulbasaur|pokemon_id|5|2016-01-01
charmander|pokemon_id|0|2014-01-01
charmander|pokemon_id|2|2014-01-01
```

If keyed by `time`:

```
> feature pokemon_id ~> newest value
[bulbasaur, 5,charmander, 0]
```

If unkeyed:

```
> feature pokemon_id ~> newest value
[bulbasaur, 5,charmander, 2]
```

@jystic @amosr


icicle-source/src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -593,3 +602,99 @@ convertReduce xx
...
+convertKey :: (Hashable n, Eq n)
+           => ConcreteKey (Annot a n) n
+           -> CoreBinds () n
+           -> ConvertM a n (CoreBinds () n)
+convertKey (ConcreteKey Nothing)  core = return core
+convertKey (ConcreteKey (Just k)) core = do
+  let a'k    = annotOfExp k
+  k'        <- convertExp k
+  t'k       <- convertValType (annAnnot a'k) (annResult a'k)
+  let t'key  = T.OptionT t'k
```
amosr: maybe worth mentioning that the nub key has to be monotonic / sorted?

icicle-source/src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -593,3 +602,99 @@ convertReduce xx
...
+               $ CE.xLet n'acc'old a
+               $ CE.xLet n'key'new (CE.makeApps () (someX t'k        ) [k'             ])
+               $ CE.xLet n'acc'new xx'
+               $ CE.makeApps ()
+                   ( CE.xPrim (C.PrimFold C.PrimFoldBool (T.PairT t'key t'acc)) )
+                   [ CE.xVar n'input
+                   , CE.makeApps () (pairX t'key t'acc) [ CE.xVar n'key'new, CE.xVar n'acc'new ]
+                   , CE.makeApps () (eqX   t'key      ) [ CE.xVar n'key'old, CE.xVar n'key'new ] ]
+
+        pure ( x', Map.singleton n'input a )
```
amosr: I was imagining you could add this new_key and fold_bool as two extra folds right at the start, then just change each fold to refer to the fold_bool. rather than adding these to each fold?

icicle-source/src/Icicle/Source/ToCore/ToCore.hs
```diff
@@ -593,3 +602,99 @@ convertReduce xx
...
+        let s = C.SFilter x fs'
+        pure (s : ss, sub <> subs)
+
+      nub [] = pure ([], Map.empty)
+
+  (streams', subs) <- hoist runFreshIdentity . nub
+                    . streams $ core
+  postcomps'       <- lift . runFreshIdentity
+                    . mapM (\(n, x) -> (n,) <$> CE.subst () subs x)
+                    . postcomps $ core
```
amosr: I didn't realise you could do it without modifying convertQuery/convertReduce, so this is really cool

icicle-source/src/Icicle/Storage/Dictionary/Toml.hs
```diff
@@ -199,7 +206,31 @@ checkDefs checkOpts d defs
...
+                   (SQ.Prim p (SQ.Lit (SQ.LitInt 0))))
+
+  (checked, _)  <- first DictionaryErrorCompilation $ do
+    _ <- P.sourceDesugarQT q
+    P.sourceCheckQT checkOpts d q
+
+  case contexts . query $ checked of
+    SQ.Distinct _ xx' : _
+      -> Right . ConcreteKey . Just $ xx'
+    _ -> Left DictionaryErrorImpossible
```
amosr: haha nice

amosr:
I would like a test with a group or something because the folds act slightly differently, but it looks correct

🎍

jystic:
This is awesome 🎁 

Would like to see a test / example where a struct field is used as the key, something like `transaction_id`

icicle-source/src/Icicle/Source/ToCore/ToCore.hs
tranma: probably won't make too much a difference for the generated core (I hope) but it might make this bit cleaner

doc/user/refutation.md
```diff
@@ -0,0 +1,86 @@
...
+zubat|capitalism|{"transaction_id":3,"dollarydoos":1}|2016-01-01
+zubat|capitalism|{"transaction_id":3,"dollarydoos":9}|2016-01-05
+```
+
+```
+> feature capitalism ~> count dollarydoos
+[zubat, 2]
+
+> feature capitalism ~> newest dollarydoos
+[zubat, 1]
```
markhibberd: `[zubat, 9]`? I would expect last to win anyway, is it first wins?

doc/user/refutation.md
```diff
@@ -0,0 +1,86 @@
...
+[blatoise, [("foo14",2014-01-01)
+           ,("foo15",2015-01-01)
+           ,("foo16",2016-01-01)]]
+```
+
+### Struct Values
+
+```
+[fact.capitalism]
+  key = "transaction_id"
```
markhibberd: Would normally always use time, an example of struct field + time, would be good.

doc/user/refutation.md
tranma: it's first-wins, should it be configurable, or last-wins instead?

jystic: The idea was that the highest priority facts will come first, this makes it easier to implement sparse states because you don't have to undo a field that you may have already set, if that makes sense?

tranma: so this is hinging on the assumption that first-wins vs. last-wins doesn't matter, because we shouldn't have a fact with the same key and priority, but different values.

markhibberd: Yep, that is good. Makes sense. That interaction with priority is the main thing.

tranma:
@jystic @amosr redid the liveness analysis to be betterer. Unfortunately now the mutation recovery is more restrictive, for example when the local name shadows the accumulator name (`Write n <expression using n>`).

tranma:
this is right but exposed cases where array mutation breaks, I'll fix that before merge

icicle-compiler/src/Icicle/Avalanche/Statement/Simp/Mutate.hs
```diff
@@ -1,86 +1,70 @@
...
-
-
-mutateExp :: (Hashable n, Eq n) => Set (Name n) -> Exp a n F.Prim -> Exp a n F.Prim
+mutateStmt :: (Hashable n, Eq n, Pretty n)
+           => Set (Name n) -- live vars from other nodes that are dependent this node
+                           -- but not reachable from the AST definition.
+                           -- this is a hack so we don't have to construct a proper CFG
+           -> Statement a n F.Prim
+           -> Statement a n F.Prim
+mutateStmt children ss = case ss of
```
amosr: what case was it incorrectly changing to a mutate before?

tranma: no actually... I went over the examples thoroughly, the mutate transform is correct, it's not the problem (C result /= Core result). Still I rewrote it with a proper cfg anyway, we might be able to use that code for other analysis in future.

amosr:
I'm a bit confused. does this change introduce cases that the mutation recovery doesn't pick up? was it incorrect before?

tranma:
merging this as-is (added fact key, but no more array mutation -- that will come back in a separate branch)

PR #434, opened 2016-10-19
Output tombstone

tranma:
PSV output was not writing out tombstones (it just doesn't write out anything, e.g. "homer|test|"). Here is a quick fix that uses the tombstone value defined for that fact/namespace (if exists, otherwise the default tombstone "NA" is used)

```
namespace = """default"""
tombstone = "foo"
[fact.injury]
  tombstone  = "none"
  encoding = "(location:string,severity:double,action:string*)"
[feature.test]
   expression = "feature injury ~> newest action"
```

```
homer|injury|{"location":"torso","severity":3,"action":"ignore"}"|1999-01-01
homer|injury|{"location":"torso","severity":1}"|2010-01-01
```

```
homer|test|none
```

! @jystic

### DIFF
--- a/icicle-compiler/src/Icicle/Benchmark.hs
+++ b/icicle-compiler/src/Icicle/Benchmark.hs
@@ -32,7 +32,6 @@ import           Control.Monad.IO.Class (liftIO)
 
 import           Data.Map (Map)
 import qualified Data.Map as Map
-import           Data.Set (Set)
 import qualified Data.Text.Lazy.IO as TL
 import           Data.Time (NominalDiffTime, getCurrentTime, diffUTCTime)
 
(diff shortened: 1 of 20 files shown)


jystic:
So the tombstones are ordered now and it picks the first one?

tranma:
picks the last one, yeah, as it's the one "most recently defined"

jystic:
Cool, looks good 🍇

tranma:
@raronson this will fix your ice issue, will be piggybacked on jystic's ice changes

jystic:
will update my PR to reference this

markhibberd:
This isn't quite right, missing value on output should not be related to input tombstone. Needs to be a uniform token for all values.

jystic:
@markhibberd see the new PR

PR #435, opened 2016-10-19
Use one output tombstone value

tranma:
! @jystic 
tombstone will be an ice argument, or defaults to NA

### DIFF
--- a/icicle-compiler/src/Icicle/Benchmark.hs
+++ b/icicle-compiler/src/Icicle/Benchmark.hs
@@ -31,6 +31,7 @@ import           Icicle.Storage.Dictionary.Toml
 import           Control.Monad.IO.Class (liftIO)
 
 import           Data.Map (Map)
+import           Data.Set (Set)
 import qualified Data.Map as Map
 import qualified Data.Text.Lazy.IO as TL
 import           Data.Time (NominalDiffTime, getCurrentTime, diffUTCTime)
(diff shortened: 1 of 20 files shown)


icicle-compiler/src/Icicle/Benchmark.hs
```diff
@@ -103,7 +104,7 @@ createBenchmark mode dictionaryPath inputPath outputPath dropPath packedChordPat
 
   let cfg = HasInput
           ( FormatPsv (PsvInputConfig  mode input')
-                      (PsvOutputConfig mode output'))
+                      (PsvOutputConfig mode output' defaultOutputTombstone))
```
jystic: As per @raronson's comments in the ice PR, we should call this missing value

jystic:
Looks good 👍

tranma:
@raronson note that this applies to arrays/maps/structs too, so `feature injury ~> latest 2 ~> action` gives
`homer|test|["ignore",NA]`
rather than
 `homer|test|["ignore"]`

given
```
homer|injury|{"location":"torso","severity":3,"action":"ignore"}"|1999-01-01
homer|injury|{"location":"torso","severity":1}"|2010-01-01
```

jystic:
> @raronson note that this applies to arrays/maps/structs too

I'm not sure if this is right, as the result won't be valid JSON anymore

tranma:
hipspace discussion - leaving this as-is until we figure out how we want to let the user control the behaviour of tombstones (bottom/maybe/combination of both, etc.)

PR #436, opened 2016-10-20
Temporality constraints can be strict

tranma:
In Source, our constraints on `Temporality` are more like sub-typing, i.e.

```
  requireTemporality ty tmp
   | (tmp',_,_) <- decomposeT ty
   = case tmp' of
      Just TemporalityPure -> return []
      Nothing              -> return []
      Just tmp''
       -> return $ require a (CEquals tmp'' tmp)
```

This is fine, but it means we have more programs to cover unnecessarily in ToCore. I think we can just add a `requireTemporalityStrict` where `ty` must have a temporality exactly as required. This can be used for things like `GroupBy` keys (see the diff).

I'm happy to consider other solutions if there are better ways.

### DIFF
--- a/icicle-source/src/Icicle/Source/Checker/Constraint.hs
+++ b/icicle-source/src/Icicle/Source/Checker/Constraint.hs
@@ -263,7 +263,7 @@ generateQ qq@(Query (c:_) _) env
 
             let tkey = annResult $ annotOfExp x'
 
-            consT          <-  (<>) <$> requireTemporality tkey TemporalityElement
+            consT          <-  (<>) <$> requireTemporalityStrict tkey TemporalityElement
                                     <*> requireAgg tval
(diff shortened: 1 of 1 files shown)


tranma:
! @amosr

amosr:
hmm, so what's the benefit of this? that something like
```
filter True
group 5
```
will be outlawed?

I think I need more context - what is this trying to solve?

tranma:
@amosr trying to outlaw expressions like `group True ~> ...`

tranma:
hm, on second thought, I think this isn't right after all. There should be no reason why the sub-typing relation isn't appropriate.

tranma:
yeah, this is wrong.
I was having problems with expressions like `group .. ~> box (Some None)` where conversion to Core failed, but that's a problem with the conversion, not with the Source. Nvm this change

PR #438, opened 2016-10-25
Fix convert fold patttern

tranma:
This was picked up by quickcheck:

```
feature foo ~> group anything ~> case something_pure | Some x -> Right x | .. end

Variable not bound: desugar_q$0$conv$11
```

It should insert a Unit into the map, rather than try to use the unbound `x`

! @amosr


amosr:
🏖

Issue #439, opened 2016-10-31
Retrieve the input filepath from the user


tranma:
We need a proper way to ask another process (the user) for the path and offset of some icicle input. This will be useful for error reporting.

PR #440, opened 2016-10-31
Use readlink to get input filepath from the fd

tranma:
This is a dirty stop-gap solution to find out the input filename when an error occurs. It only works on Linux (tested on CentOS 6), should eventually be replaced when we get to https://github.com/ambiata/icicle/issues/439. I hope this isn't too savage for now.

! @jystic @olorin

### DIFF
--- a/data/sea/05-error.h
+++ b/data/sea/05-error.h
@@ -65,47 +65,49 @@ typedef struct ierror_loc {
     const char   *line_end;
     const char   *span_start;
     const char   *span_end;
+    iint_t        fd;
     ierror_tag_t  tag;
     char          message[0];
 } *ierror_loc_t;
 
-static ierror_loc_t NOINLINE v_ierror_loc_tag_format (const ierror_tag_t tag, const char *start, const char *end, const char *fmt, va_list args)
+static ierror_loc_t NOINLINE v_ierror_loc_tag_format (const ierror_tag_t tag, const iint_t fd, const char *start, const char *end, const char *fmt, va_list args)
 {
-  const  size_t message_size = 4 * 1024;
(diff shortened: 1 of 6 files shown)


olorin:
:cat: after length fix.

data/sea/05-error.h
```diff
@@ -121,6 +123,19 @@ static ierror_msg_t NOINLINE ierror_loc_pretty (ierror_loc_t loc, iint_t line)
     char *p  = msg_text;
     char *pe = msg_text + msg_size;
 
+    char *filename = calloc (255, 1);
```
tranma: yup, thanks

tranma:
This will not necessarily make it to master. I'll help in finding problems in datasets but I'd rather hold it off for now if it can be helped.

amosr:
does this work when you have multiple files being merged? I thought the file descriptors would just be an unnamed pipe in that case

tranma:
@amosr should work in icicle-bench, not ice?

amosr:
yeah I'd believe that.. for finding parse errors maybe could run icicle-bench with no features over each file individually?

tranma:
@amosr yeh that's what I thought too, but when people tried running ice over each file in a shell loop apparently it didn't fail - so I just cooked up this savage solution to help debug.

data/sea/05-error.h
```diff
@@ -121,6 +123,19 @@ static ierror_msg_t NOINLINE ierror_loc_pretty (ierror_loc_t loc, iint_t line)
...
+    char *proc     = calloc (269, 1);
```
jystic: is the intention here not to bother freeing the memory because it's only a temporary fix?

jystic:
This was just a temporary debug thing, closing

PR #441, opened 2016-11-04
skip_to_next_entity bug

tranma:
skip_to_next_entity marks part of the input buffer as skipped (by setting state->input_ptr), but this needs to be reset before each read_whole_buffer.

we didn't encounter this before because we didn't test with big data so the input always fit into the read buffer, even when some of it is skipped.

! @amosr @jystic


erikd-ambiata:
Well done! Is it possible to write a test for this?

tranma:
yeah, perhaps we should make the buffer sizes configurable, and run quickcheck with random sizes

erikd-ambiata:
@tranma    `s/perhaps//`

amosr:
super 🗳

PR #443, opened 2016-11-07
Allow unlimited number of facts per entity

tranma:
Just what it says on the tin. That was silly to begin with... Turns out users don't usually set the limit.

! @amosr

### DIFF
--- a/data/sea/52-psv.h
+++ b/data/sea/52-psv.h
@@ -84,7 +84,7 @@ static ierror_loc_t INLINE psv_read_fact
   , const char   *time_ptr
   , const size_t  time_size
   , ifleet_t     *fleet
-  , const size_t  max_ent_attr_count);
+  , const size_t  facts_limit);
 #else
 static ierror_loc_t INLINE psv_read_fact
   ( const char   *entity_ptr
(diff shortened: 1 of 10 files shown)


jystic:
Don't we usually want a limit though? (maybe I missed something last week)

markhibberd:
I don't understand this either, there needs to be a limit unless I am missing something obvious.

tranma:
There still can be a limit, the user just needs to set it explicitly in ice/icicle-bench. Currently if nothing is provided, icicle uses 1024*1024 per entity, which is very low. (I saw several entities exceeding this limit)

tranma:
Having an `Unlimited` option is still good I think, I can just make `Limited 1million` the default again if that's preferred

markhibberd:
Maybe the limit is a bit low, (but I am not sure that a million facts in 2 years should be common unless there is an issue that we are missing elsewhere). But there should always be a limit, it should be safe and predictable by default and we can choose to lift limit if we hit it.

tranma:
Ack, the default will still be 1 million then.

markhibberd:
As long as prod will _never_ run with unlimited by accident.

jystic:
We decided this isn't worth it, will revisit if necessary

PR #444, opened 2016-11-10
Array update REPL tests

tranma:
Just for reference, these examples will fail if all array updates are mutable.

! @amosr @jystic


amosr:
I stole these

amosr:
https://github.com/ambiata/icicle/pull/445

PR #445, opened 2016-11-10
Brave New Master

amosr:
So this should let us leave everything else as-is, while making the map_insert copy once then do in-place mutates?

! @tranma @jystic

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ gen
 *.hi
 *.chi
 *.chs.h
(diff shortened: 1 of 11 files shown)


amosr:
lol no tests failed

tranma:
yep, using mutate_put here is fine. Should be easy enough to add a copy primitive to Flat too.

tranma:
🅰️ gile

PR #446, opened 2016-11-11
After writing the last entity, flush the drop buffer

tranma:
! @amosr


jystic:
This is in https://github.com/ambiata/icicle/pull/445 now

PR #448, opened 2016-11-11
Array delete

amosr:
! @jystic @tranma

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
+++ b/icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
@@ -151,7 +151,7 @@ seaOfWriteProgramOutput config state = do
 seaOfWriteOutputSparse :: Doc -> Int -> OutputName -> ValType -> [ValType] -> Either SeaError Doc
 seaOfWriteOutputSparse struct structIndex outName@(OutputName name _) outType argTypes
   = do (m, f, body, _, _)
-          <- seaOfOutput NotInJSON struct structIndex outName PsvDrop Map.empty outType argTypes id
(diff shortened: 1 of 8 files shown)


icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
```diff
@@ -319,15 +321,15 @@ seaOfOutput isJSON struct structIndex outName@(OutputName name _) missing env ou
...
                    , outputMissing, body', ix, ts )
 
        -- Base
        _
         | (t  : ts) <- argTypes
         , (mx : _)  <- members
-        , mx'       <- transform mx
+        , mx'       <- transform t mx
         -> do d <- seaOfOutputBase' isJSON t mx'
               pure (Nothing, Nothing, d, structIndex + 1, ts)
```
tranma: checked out :chipmunk:

icicle-repl/test/cli/repl/t60-map-delete/script
```diff
@@ -0,0 +1,10 @@
+:set +core-eval
+:set +c-eval
+
+feature injury ~> fold x = map_create : case map_lookup (Some location) x | None -> map_insert (Some location) (Some severity) x | Some _ -> map_delete (Some location) x end ~> x
+-- fold x = map_create :
+--  case map_lookup location x
+--  | None -> map_insert location severity x
+--  | Some _ -> map_delete location x
+--  end
```
tranma: could we have this test case in bench as well? It's an interesting output type

tranma:
☔️

icicle-repl/test/cli/repl/t60-map-delete/script
amosr: done

PR #451, opened 2016-11-11
Test with random buffer sizes

tranma:
! @amosr @jystic

### DIFF
--- a/icicle-compiler/test/Icicle/Test/Sea/Psv.hs
+++ b/icicle-compiler/test/Icicle/Test/Sea/Psv.hs
@@ -42,7 +42,8 @@ import           System.IO
 import           System.IO.Temp (createTempDirectory)
 import           System.Directory (getTemporaryDirectory, removeDirectoryRecursive)
 
-import           Test.QuickCheck (Gen,arbitrary, elements, suchThat)
+import           Test.QuickCheck (Gen, Arbitrary(..), elements, suchThat)
+import           Test.QuickCheck (getPositive)
 import           Test.QuickCheck (Property, (==>), property, counterexample)
 import           Test.QuickCheck.Property (succeeded, failed)
 import           Test.QuickCheck.Monadic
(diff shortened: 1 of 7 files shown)


icicle-compiler/test/Icicle/Test/Sea/Psv.hs
```diff
@@ -174,16 +179,39 @@ failWithError' prints wt = \case
...
+    psv <- genPsvOpts wt
+    return $ WellTypedPsv wt psv
+
+genPsvOpts :: WellTyped -> Gen PsvOpts
+genPsvOpts wt = do
+  -- maximum number of rows to read before compute
+  let row x = x + 1
+  maxRowCount <- row . getPositive <$> arbitrary
+  -- the buffer needs to be at least as large as a single line
+  let str x = x + longestLine wt + 4
```
jystic: do we ever want to test what happens when the buffer isn't long enough? /not sure

jystic:
Looks good to me 👍

icicle-compiler/test/Icicle/Test/Sea/Psv.hs
tranma: I'm not sure what the sensible behaviour in that case is. Right now I believe ice/icicle-bench segfaults.

jystic: Yeah we probably shouldn't do that :) I think either we should drop the row, or fail the whole job with an error. Dropping the row is probably the nicest, not sure how hard that is though.

amosr: ice merge should blow up with a reasonable error message in this case, so I think it's adequate for now

tranma: cool, will rebase and merge this

jystic:
Bump, should be good to go 👍

PR #462, opened 2016-11-18
Loop over map arrays in reverse so that delete on an index is safe

jystic:
! @amosr @tranma


tranma:
good after tests updated 🏂

PR #464, opened 2016-11-24
group ~> group

tranma:
just the GroupBy first

! @amosr

### DIFF
--- a/icicle-source/src/Icicle/Source/Checker/Invariants.hs
+++ b/icicle-source/src/Icicle/Source/Checker/Invariants.hs
@@ -33,10 +33,11 @@ invariantQ ctx (Query [] x)
 invariantQ ctx (Query (c:cs) xfinal)
  = case c of
     Windowed{}
-     | allowWindowsOrGroups inv
-     -> go
+     | allowWindows inv
+     -> goBanWindowAndGroupFold
      | otherwise
-     -> errBanGroup "Consider moving the window to the start of the query"
+     -> errBanWindow "Consider moving the window to the start of the query"
+
     Latest{}
      | allowLatest inv
      -- XXX: ban latest inside latests.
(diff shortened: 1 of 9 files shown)


icicle-source/src/Icicle/Source/Checker/Invariants.hs
```diff
@@ -77,12 +74,12 @@ invariantQ ctx (Query (c:cs) xfinal)
...
   goBanAll
      = invariantQ
-        (ctx{ checkInvariants = inv{ allowLatest = False, allowWindowsOrGroups = False }})
+        (ctx{ checkInvariants = inv{ allowLatest = False, allowWindowsOrGroups = True }})
         q'
 
   goBanGroup
      = invariantQ
-        (ctx{ checkInvariants = inv{ allowWindowsOrGroups = False }})
+        (ctx{ checkInvariants = inv{ allowWindowsOrGroups = True }})
```
amosr: is `allowWindowsOrGroups` still necessary then? maybe these should remain `False` and just rename it to `allowWindows`

icicle-source/src/Icicle/Source/ToCore/Fold.hs
```diff
@@ -297,11 +303,76 @@ convertFold q
...
+                 , CE.xVar n'x'
+                 , CE.xVar n'k'
+                 , CE.xVar n'r' ]
+
+           let xtra
+                = CE.xPrim (C.PrimFold (C.PrimFoldMap t'k t'fold) t'sum)
+                  CE.@~ ins
+                  CE.@~ x'sumzero
+
+           return $ ConvertFoldResult kons zero xtra t'map t'sum
```
amosr: yep

amosr:
looks good. can you add group-inside-group to the arbitraries, or will that happen because of the type checker changes?

tranma:
@amosr that's already happening yeah

tranma:
this is good enough for nested groups and distincts

amosr:
oh yeah I meant 👍

PR #467, opened 2016-12-12
copy zebra entities into the void

tranma:
this is just for reference for now


jystic:
Looking good

amosr:
I think just remove the printfs in compute and fix the failing tests before merging. then do the rest of the cleanup as a separate PR?

data/sea/54-zebra.h
```diff
@@ -0,0 +1,236 @@
+#include "53-zebra_data.h"
```
tranma: not much to see outside of this file

icicle-compiler/src/Icicle/Sea/IO/Zebra.hs
```diff
@@ -4,12 +4,110 @@ module Icicle.Sea.IO.Zebra (
     seaOfZebraDriver
   ) where
```
tranma: this one too

tranma:
anything outside of 54-zebra.h and Icicle/Sea/IO/Zebra.hs is pretty much inconsequential

data/sea/54-zebra.h
```diff
@@ -0,0 +1,236 @@
...
+            return 1;
+
+        case ZEBRA_DOUBLE:
+            *dst = data.d + elem_start;
+            return 1;
+
+        case ZEBRA_ARRAY:
+            {
+                istring_t *strings = imempool_alloc (mempool, count * sizeof (istring_t));
+                int offset = 1;
```
jystic: It might be worth having a comment as to why this gets set to 1, and what the rough limitations are

icicle-compiler/main/icicle-bench.hs
```diff
@@ -17,132 +18,119 @@ import           Icicle.Sea.Eval
 import           P
 
 import           System.FilePath (replaceExtension)
-import           System.Environment (getArgs)
 import           System.IO (IO, FilePath, putStrLn, print)
 
 import           Text.Printf (printf)
 
 import           X.Control.Monad.Trans.Either
+import           X.Options.Applicative
```
jystic: sorry :( should have been using this from the start

tranma:
I'll merge this as it's essentially the new master now. Cleanups will follow.

jystic:
💯

PR #473, opened 2016-12-21
WIP: upgrade zebra: lengths as prefix scan

amosr:
not tested yet


tranma:
this should be ok. There is also a zebra cli test on master now, just for sanity check

amosr:
This is superseded by #496

PR #476, opened 2017-01-04
Use new jetski instead of libffi

jystic:
! @amosr @tranma


jystic:
Depend on https://github.com/ambiata/jetski/pull/36

amosr:
👍

PR #481, opened 2017-01-09
some clean up after zebra changes

tranma:
just a bit less savagery first

! @amosr @jystic

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/Eval/Base.hs
+++ b/icicle-compiler/src/Icicle/Sea/Eval/Base.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE DoAndIfThenElse #-}
 module Icicle.Sea.Eval.Base (
     MemPool
   , SeaState
@@ -41,6 +42,7 @@ module Icicle.Sea.Eval.Base (
 
 import           Control.Monad.Catch (MonadMask(..))
 import           Control.Monad.IO.Class (MonadIO(..))
+import           Control.Monad.Morph
 
 import qualified Data.List as List
 import           Data.Map (Map)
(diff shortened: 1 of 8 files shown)


icicle-compiler/src/Icicle/Sea/Eval/Base.hs
```diff
@@ -255,64 +257,67 @@ seaCompile' options cache input programs chords = do
...
-                     $ mergeBlocks (MergeOptions puller step' 100) pullid
+            ret <- runEitherT . joinErrors show show
+                 $ mergeBlocks (MergeOptions puller' step' 100) pullid
 
-                case ret of
-                  Left e ->
-                    Savage.error $ "error merge: " <> show e
-                  Right _ ->
-                    return ()
+            -- need to error straight from IO as piano is tied to IO
```
jystic: not sure I understand this, why can't we use `EitherT`?

tranma: `withPiano` is the continuation that will be given to `withCPiano :: ForeignPiano -> (CPiano -> IO a) -> IO a`

tranma: maybe I could return the error rather than () though

jystic: yeah you can maybe do something like this:
```hs
withCPiano' :: ForeignPiano -> (CPiano -> EitherT x IO a) -> EitherT x IO a
withCPiano' fp io =
  EitherT . withCPiano fp $ runEitherT . io
```

tranma: I changed sfSnapshot was it was being liftIO'd anyway

jystic:
👍

Just a heads up, blizzard will not have a `FilePath` but a `Conduit` connected to S3. So we'll need to change this a bit when we do that integration. I suspect we'll need to pass in the puller in order to make that work, not sure.

Issue #468, opened 2016-12-15
fact identifiers inside groups


tranma:
I will fix this soon (promise).

```
=== prop_convert_is_well_typed from test/Icicle/Test/Source/Convert.hs:30 ===
*** Failed! Falsifiable (after 966 tests):
QueryWithFeature:
  Query: group fold (statler
           ,kermit) = (group kermitnow
                       ~> "cartman")
~> (latest 1 ~> kermitnow)
  Now: kermitnow
  Output: statler
  Type: Unit
  Name: kermitfeaturename
  Time: kermittimename
  Key: "token"
feature kermitfeaturename
~> group fold (statler
              ,kermit) = (group kermitnow
                          ~> "cartman")
~> (latest 1 ~> kermitnow)
Program (core$0 : Unit, core$1 : FactIdentifier, core$2 : Time, core$3 : SNAPSHOT_TIME)
Precomputations:


Streams:
  STREAM_FOLD (core$6 : (Option String, Map Time Unit))
    INIT:
      pair# (None) (Map [])
    KONS:

        let core$24 = fst# core$6
        let core$22 = snd# core$6
        let core$25 = some# ("token")
        let core$23 = Map_insertOrUpdate#
          (\core$8 ()) (()) core$3 core$22
         in if# core$6 (pair# core$25 core$23) (eq# core$24 core$25)


Postcomputations:
  core$17              = Map_fold#
                           (\core$14 \core$9 \core$11 Map_insertOrUpdate#
                             (\core$13 core$13) ("cartman") core$9 core$14) (Map []) (snd# core$6)
  core$4               =
                           let core$21 = Map_fold#
                             (\core$5 \statler$core$18 \kermit$core$19 Latest_push# core$5 core$1 core$3) (Buf []) core$17
                            in Latest_read# core$21

Returning:
  statler              = core$4

Left (ProgramErrorPost (ExpErrorVarNotInEnv (Name {nameHash = 8985580063069689063, nameBase = NameMod (Variable "core") (NameBase (Variable "1"))})))
```

tranma:
simpler example:

```
> feature cylon ~> group fold (k,v) = (group time ~> 0) ~> newest v
- Core (simplified):
Program (conv$0 : (Sum Error Int), conv$1 : FactIdentifier, conv$2 : Time, conv$3 : SNAPSHOT_TIME)
Precomputations:


Streams:
  STREAM_FOLD (conv$6 : Map Time Unit)
    INIT:
[...]

PR #488, opened 2017-01-11
Revert "Disable prop_convert_is_well_typed"

jystic:
I'm just leaving this here so we don't forget to enable that (important!) test again

! @amosr @tranma


jystic:
Re-enable once https://github.com/ambiata/icicle/issues/468 is fixed

tranma:
this was done

jystic:
\o/

PR #491, opened 2017-01-19
Remove Cabal from lockfiles

amosr:
This is breaking build for me - I think we discussed this before and realised it had been added accidentally, but I don't really remember the details.

! @jystic @tranma


tranma:
👍

PR #492, opened 2017-01-19
use anemone_mempool (the savage way)

tranma:
I think this breaks icicle-make

! @jystic @amosr


jystic:
Don't you need to copy the definitions from `anemone_mempool.h` in to `10-mempool.h` or something similar? (the same as with piano`

`anemone/anemone_mempool.h` won't be available on client machines

tranma:
@jystic it's very savage. `anemone_mempool.h` is concat'd into the preamble (see Sea.Preamble) and the include in mempool.h is just there for editor integration, it's stripped out afterwards :)

data/sea/51-piano.h
```diff
@@ -1,13 +1,13 @@
...
-typedef void error_t;
+typedef void piano_error_t;
 typedef struct piano piano_t;
 
 int64_t piano_max_count (
     piano_t *piano
   );
 
-error_t piano_lookup (
+piano_error_t piano_lookup (
```
jystic: This can be `error_t` and you can delete the `typedef`, piano uses the anemone error

icicle-compiler/src/Icicle/Sea/Preamble.hs
```diff
@@ -26,16 +26,37 @@ import           X.Data.FileEmbed (embedWhen)
...
  = vsep
- $ fmap (uncurry seaOfExternal) files
+ ( fmap (uncurry seaOfExternal) externals <> fmap (uncurry seaOfHeader) files )
  where
   files
    = List.sortBy (compare `on` fst)
    $ $(embedWhen ((== ".h") . takeExtension) "data/sea/")
+  externals
+   = let base    = $(embedWhen ((== "anemone_base.h"))    "data/csrc/anemone")
+         mempool = $(embedWhen ((== "anemone_mempool.h")) "data/csrc/anemone")
```
jystic: Why is `data/csrc/anemone` different to `../../lib/anemone/csrc/anemone_mempool.h` ?

tranma: it's a symlink, looks like it's missing in the commit -- sorry about that

data/csrc/anemone/anemone_mempool.c
```diff
@@ -0,0 +1 @@
+../../../lib/anemone/csrc/anemone_mempool.c
```
jystic: is this actually required?

icicle-compiler/src/Icicle/Sea/Preamble.hs
amosr: could you just symlink the files as `00-anemone_mempool.h` instead, so they came first?

data/csrc/anemone/anemone_mempool.c
tranma: no, I just think it's cleaner than putting this path in Preamble -- happy to just do that

icicle-compiler/src/Icicle/Sea/Preamble.hs
tranma: I wanted to be explicit in case those headers don't have same structure as icicle sea headers (ie the file doesn't start with the #includes)

icicle-compiler/src/Icicle/Sea/Preamble.hs
```diff
@@ -26,16 +26,37 @@ import           X.Data.FileEmbed (embedWhen)
...
+ ]
+ where
+  (_, file)
+   = case T.decodeUtf8' bs of
+       Left err ->
+         Savage.error $
+           "Icicle.Sea.Preamble.seaOfExternal: failed to decode: " <> path <> "\n" <>
+           show err
+       Right txt ->
+         List.partition (T.isPrefixOf "#include \"") $ T.lines txt
```
amosr: this is probably better - might as well remove `seaOfHeader` and replace it with this?

tranma: yeah, I guess the line number isn't all that useful anyway

icicle-compiler/src/Icicle/Sea/Preamble.hs
tranma: meh, I'll just symlink them into `data/sea` as well

icicle-compiler/src/Icicle/Sea/Preamble.hs
amosr: oh, I didn't notice. you could still put the `#line` in here couldn't you?

amosr: ah never mind. I was not reading properly

amosr: I think we might be able to keep correct line numbers in both cases.. maybe?
```
lines = zipWith mkLine (T.lines txt) [1...]
mkLine line lineNo
 | T.isPrefixOf "#include \""...
 = "#line blah " <> show lineNo
 | otherwise
 = line
```

data/csrc/anemone/anemone_mempool.c
tranma: oh, I misread -- no the .c doesn't need to be symlinked

icicle-compiler/src/Icicle/Sea/Preamble.hs
tranma: it's then used to show where the source starts after the includes, so I don't think so...? like with

```
#ifdef ...
#include...
#include...
static const...
```

after stripping the includes, there isn't really a right number you can use for the rest of the blob:

```
#ifdef ...
static const...
```

tranma:
no, this doesn't break icicle-make

data/sea/10-mempool.h
```diff
@@ -66,66 +18,42 @@ static void imempool_add_block (imempool_t *pool)
         return ptr;                                                                                    \
     }
 
-static void * NOINLINE imempool_alloc_block (imempool_t *pool, size_t num_bytes)
+static void * NOINLINE imempool_alloc_block (anemone_mempool_t *pool, size_t num_bytes)
```
jystic: I'm a little confused about keeping these functions around instead of just using the anemone ones directly, is this for tests or something?

data/sea/10-mempool.h
```diff
@@ -1,56 +1,8 @@
...
-    const iblock_t *prev = block->prev;
-
-    free (block->ptr);
-    free ((iblock_t *) block);
-
-    iblock_free (prev);
-}
-
-static void imempool_add_block (imempool_t *pool)
+static void INLINE imempool_add_block (anemone_mempool_t *pool)
```
jystic: cruft? (empty function)

tranma: yeah thanks

jystic:
> no, this doesn't break icicle-make

is this because we include the anemone_mempool.c in the externals too?

tranma:
> is this because we include the anemone_mempool.c in the externals too?

yeah

data/sea/10-mempool.h
```diff
@@ -1,131 +1,47 @@
...
+static void * NOINLINE imempool_alloc_block (anemone_mempool_t *pool, size_t num_bytes)
 {
-    imempool_add_block (pool);
-
-    TRY_ALLOC (imempool_alloc_block);
-
-    /* Couldn't allocate even after adding a new block to the pool, we will
-     * never be able to service this request. */
-    fprintf (stderr, "oh my, someone's a bit greedy\n");
-    fprintf (stderr, "did you really need %zu bytes\n", num_bytes);
```
jystic: 🎈

data/sea/10-mempool.h
tranma: I was thinking we might want to roll the alloc check into these functions. But maybe that's a bad idea

jystic: The greedy check?

tranma: the alloc for each entity check. I was considering checking the alloc'd amount somewhere in here, and return an error. But I think the idea discussed today (checking before/after compute in psv and zebra) is better anyway. So these functions can go

jystic: cool 👍

data/sea/10-mempool.h
```diff
@@ -1,131 +1,17 @@
...
-void imempool_free (imempool_t *pool)
-{
-#if ICICLE_DEBUG
-    fprintf (stderr, "imempool_free: %p\n", pool);
-    imempool_debug_block_usage (pool);
-#endif
-
-    iblock_free (pool->last);
-    free(pool);
+    fprintf (stderr, "anemone_mempool_debug_block_usage: %" PRIu64 " blocks, %" PRIu64 " bytes\n", blocks, size);
```
jystic: is this just calculating the stuff which anemone already keeps track of now?

tranma: I think this also counts the number of blocks

jystic: ah cool

data/sea/54-zebra.h
```diff
@@ -11,8 +11,12 @@ typedef struct {
 
     /* outputs */
     const char  *error;
-    iint_t       fact_count;
-    iint_t       entity_count;
+    iint_t fact_count;
+    iint_t entity_count;
+
+    /* options */
+    const size_t fact_limit;
```
jystic: is this change supposed to be here? or the beginning of something else?

tranma: yeah, this should be the alloc limit rather than fact limit, it will come later as a separate branch

jystic:
Cool, this is looking good 🏎

data/sea/10-mempool.h
amosr: yeah - it is far less useful now (and actually size is wrong since blocks might be larger) but I like keeping these debugging functions around

amosr: wait. I think this function (or a similar one) is also defined in anemone_mempool.c, and I was getting them confused

tranma: oh yeah, it's there too. Can get rid of this module entirely then.

PR #494, opened 2017-01-23
zebra stats

tranma:
This was just missing before. Changing to dropping based on alloc limit will be a separate PR.

```
icicle: total entities  = 55670
icicle: total facts     = 114464
```

! @jystic @amosr

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/Eval.hs
+++ b/icicle-compiler/src/Icicle/Sea/Eval.hs
@@ -81,8 +81,8 @@ seaZebraSnapshotFd :: SeaFleet ZebraState
                    -> EitherT SeaError IO ZebraStats
 seaZebraSnapshotFd fleet input output mchords = do
   withWords 6 $ \pState -> do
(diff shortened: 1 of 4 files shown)


icicle-compiler/src/Icicle/Sea/Eval.hs
```diff
@@ -81,8 +81,8 @@ seaZebraSnapshotFd :: SeaFleet ZebraState
                    -> EitherT SeaError IO ZebraStats
 seaZebraSnapshotFd fleet input output mchords = do
   withWords 6 $ \pState -> do
-  a <- liftIO $ newCString input
-  pokeWordOff pState 0 a
+  input_path <- liftIO $ newCString input
```
jystic: memory leak?

tranma: oops, it's just the config but yeah 😁

jystic:
Cool 👍

PR #496, opened 2017-01-27
Read n zebra facts at a time

tranma:
We will read some maximum number of zebra facts per entity before stopping and calling compute. I think this is a good place to check the allocation limit (will be separate PR).

! @jystic @amosr 
/jury approved @amosr

### DIFF
--- a/data/sea/54-zebra.h
+++ b/data/sea/54-zebra.h
@@ -18,20 +18,28 @@ typedef struct {
 } zebra_config_t;
 
 typedef struct zebra_state {
-  /* output buffer */
-  char *output_start;
-  char *output_end;
-  char *output_ptr;
+    /* output buffer */
+    char *output_start;
+    char *output_end;
+    char *output_ptr;
 
-  /* stats */
-  iint_t fact_count;
-  iint_t entity_count;
+    /* stats */
+    iint_t fact_count;
+    iint_t entity_count;
 
-  /* fleet state */
-  ifleet_t *fleet;
+    /* fleet state */
+    ifleet_t *fleet;
 
-  /* output file descriptor */
-  int output_fd;
+    /* output file descriptor */
+    int output_fd;
+
+    /* read state */
+    int64_t  entity_alloc_limit;
+    int64_t  attribute_count;
+    int64_t *entity_fact_offset;
+    int64_t *entity_fact_count;
+    ibool_t *entity_attribute_leftover;
+
 } zebra_state_t;
 
 
@@ -47,19 +55,27 @@ static int64_t zebra_translate_column
   , iint_t elem_start
   , iint_t count
   , void **dst
-
   , const zebra_column_t *src );
 
-static ierror_msg_t zebra_read_entity (zebra_state_t *state, zebra_entity_t *entity);
+static ierror_msg_t zebra_read_entity
+  ( piano_t *piano
+  , zebra_state_t *state
+  , zebra_entity_t *entity );
 
+ierror_msg_t zebra_snapshot_step
+  ( piano_t *piano
+  , zebra_state_t *state
+  , zebra_entity_t *entity );
 
 /* map a zebra entity to sea fleet inputs */
 static ierror_msg_t zebra_translate
-    ( zebra_state_t        *state
+    ( piano_t              *piano
+    , zebra_state_t        *state
+    , zebra_entity_t       *entity
     , anemone_mempool_t    *mempool
     , int                   attribute_ix
     , itime_t               chord_time
-    , iint_t               *new_count
+    , iint_t               *new_fact_count
     , ierror_t            **new_tombstone
     , itime_t             **new_fact_time
     , iint_t                columns_to_fill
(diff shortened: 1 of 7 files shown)


data/sea/54-zebra.h
```diff
@@ -3,6 +3,9 @@
 #if !ICICLE_NO_INPUT
 #if ICICLE_ZEBRA
 
+/* Number of facts to read before starting compute and checking allocation limit */
+static const iint_t zebra_chunk_fact_count = 2;
```
amosr: is it worth putting this in `zebra_config_t` to make it easier to quickcheck with different limits?

data/sea/54-zebra.h
```diff
@@ -79,6 +103,7 @@ static ierror_msg_t zebra_translate
     /* temporary fix until icicle uses the same time as zebra */
     itime_t zebra_chord_time = itime_to_epoch_seconds (chord_time);
 
+    /* the number of facts before chord time for this entity */
     for (iint_t i = 0; i != row_count; ++i) {
```
amosr: this is counting all the facts for every call, but could you just loop from `state->entity_fact_offset` up to `state->entity_fact_offset + zebra_chunk_fact_count`?

data/sea/54-zebra.h
```diff
@@ -87,25 +112,38 @@ static ierror_msg_t zebra_translate
         fact_count++;
     }
 
+    /* the number of unread facts for this entity */
+    iint_t fact_offset    = state->entity_fact_offset;
+    iint_t fact_remaining = fact_count == 0 ? 0 : fact_count - fact_offset;
```
amosr: I would have thought `fact_count = 0 ==> state->entity_fact_offset = 0` anyway

data/sea/54-zebra.h
```diff
@@ -87,25 +112,38 @@ static ierror_msg_t zebra_translate
...
+        fact_count = fact_remaining;
+        state->entity_fact_offset = 0;
+    }
+
+    state->fact_count += fact_count;
+
     itime_t *sadface_times = anemone_mempool_alloc (mempool, fact_count * sizeof (itime_t));
 
     for (iint_t i = 0; i != fact_count; ++i) {
         sadface_times[i] = itime_from_epoch_seconds (attribute->times[i]);
```
amosr: `attribute->times[i + fact_offset]`

data/sea/54-zebra.h
```diff
@@ -87,25 +112,38 @@ static ierror_msg_t zebra_translate
...
     }
 
-    *new_count = fact_count;
-    *new_tombstone = (ierror_t*)attribute->tombstones;
-    *new_fact_time = sadface_times;
-
-    zebra_translate_table (mempool, 0, fact_count, dst, &attribute->table);
+    *new_fact_count = fact_count;
+    *new_tombstone  = (ierror_t*) attribute->tombstones;
```
amosr: `attribute->tombstones + fact_offset`, I think

data/sea/54-zebra.h
```diff
@@ -116,14 +154,15 @@ static int64_t zebra_translate_table
...
-        offset += zebra_translate_column (mempool, elem_start, count, dst + offset, cols + i);
+        offset += zebra_translate_column (mempool, fact_offset, elem_start, count, dst + offset, cols + i);
     }
 
     return offset;
 }
 
 static int64_t zebra_translate_column
     ( anemone_mempool_t *mempool
+    , iint_t fact_offset
```
amosr: looks like you should be able to just use `elem_start` instead of adding this new parameter

data/sea/54-zebra.h
```diff
@@ -156,7 +195,7 @@ static int64_t zebra_translate_column
                 int offset = 1;
                 for (int i = 0; i != count; ++i) {
                     int len = data.a.n[i];
-                    offset = zebra_translate_table (mempool, elem_start, len, (void**)(strings + i), &data.a.table);
+                    offset = zebra_translate_table (mempool, fact_offset, elem_start, len, (void**)(strings + i), &data.a.table);
```
amosr: this was already half-wrong - it didn't support nested arrays properly which wasn't a problem, but now we need to slice out of the nested arrays

so you need to loop through from `[0,elem_start)` summing up the lengths and use that as `elem_start` for the recursive calls.
but this loop wouldn't be necessary after #473 since latest zebra stores them as prefix sum
anyway.

we might need another function to sum the offsets of a table without doing any translation - because at the moment if the array is empty the offset isn't being computed properly and just set to 1. that's not directly related to this PR but it's probably worth working this out at the same time if you want to go through it together tomorrow

data/sea/54-zebra.h
```diff
@@ -204,22 +246,38 @@ void zebra_collect_state (zebra_config_t *cfg, zebra_state_t *state)
...
+
+    ierror_loc_t  error_loc = psv_configure_fleet ((char*) entity->id_bytes, entity->id_length, piano, state->fleet);
+    if (error_loc) return ierror_loc_pretty (error_loc, 0);
+
+    ierror_msg_t error = zebra_read_entity (piano, state, entity);
+    if (error) return error;
+
+    /* keep going if we have not finished reading all facts for this entity */
+    if (state->entity_fact_offset != 0) {
+        zebra_read_step (piano, state, entity);
```
amosr: this should return the result in case it's an error

data/sea/54-zebra.h
tranma: yeah - I had that originally, but thought this is clearer. You think I should just roll it in to `elem_start`?

data/sea/54-zebra.h
tranma: yeah, this was wrong -- `fact_count == 0` should only be when we have not read any fact for this entity. That should be an invariant actually.

data/sea/54-zebra.h
tranma: sure, we have this in the config for psv too.

data/sea/54-zebra.h
amosr: I think it's cleaner to just have one, that way it's just performing a slice from `[elem_start,elem_start+fact_count)` or something

data/sea/54-zebra.h
```diff
@@ -68,44 +88,72 @@ static ierror_msg_t zebra_translate
...
+        /* the number of facts before chord time for this entity */
+        for (iint_t i = 0; i != row_count; ++i) {
+            itime_t time = attribute->times[i];
+            if (zebra_chord_time <= time) {
+                break;
+            }
+            fact_count++;
         }
-        fact_count++;
+        state->entity_fact_count = fact_count;
```
jystic: so this just runs the first time in here for each entity?

tranma: yeah

data/sea/54-zebra.h
```diff
@@ -68,44 +88,72 @@ static ierror_msg_t zebra_translate
...
 
+    /* number of unread facts for this entity */
+    iint_t fact_offset    = state->entity_fact_offset;
+    iint_t fact_remaining = fact_count - fact_offset;
+
+    if (fact_remaining > zebra_chunk_fact_count) {
+        fact_count = zebra_chunk_fact_count;
+        state->entity_fact_offset += fact_count;
+    } else {
+        fact_count = fact_remaining;
```
jystic: I don't like how this variable is written to so many times throughout the function, makes it a bit hard to follow what the meaning of it is.

jystic: as in maybe it could get a fresh name at this point, rather than re-using the `fact_count` you were incrementing above, like `fact_chunk_count` or something? /not sure

tranma: yeah fair enough

jystic: i just think we should try to write our C like Haskell where possible, pretend it doesn't have mutable variables where possible

tranma:
this is ok but I'll just leave it until there are "peek input" tests for psv and zebra.

data/sea/54-zebra.h
```diff
@@ -206,22 +270,42 @@ void zebra_collect_state (zebra_config_t *cfg, zebra_state_t *state)
...
+    ierror_loc_t  error_loc = psv_configure_fleet ((char*) entity->id_bytes, entity->id_length, piano, state->fleet);
+    if (error_loc) return ierror_loc_pretty (error_loc, 0);
+
+    ierror_msg_t error = zebra_read_entity (piano, state, entity);
+    if (error) return error;
+
+    /* keep going if we have not finished reading all facts for this entity */
+    for (int64_t i = 0; i != state->attribute_count; ++i) {
+        if (state->entity_fact_offset[i] != state->entity_fact_count[i]) {
+            return zebra_read_step (piano, state, entity);
```
amosr: I think this would be clearer as a loop

data/sea/54-zebra.h
```diff
@@ -68,38 +84,63 @@ static ierror_msg_t zebra_translate
...
+        sadface_times[i] = itime_from_epoch_seconds (attribute->times[i + fact_offset]);
+    }
 
-    zebra_translate_table (mempool, 0, fact_count, dst, &attribute->table);
+    *new_fact_count = fact_to_read;
+    *new_tombstone  = (ierror_t*) attribute->tombstones + fact_offset;
+    *new_fact_time  = sadface_times;
 
-    state->fact_count += fact_count;
+    zebra_translate_table (mempool, fact_offset, fact_to_read, dst, &attribute->table);
```
amosr: this function looks right

data/sea/54-zebra.h
```diff
@@ -152,12 +209,13 @@ static int64_t zebra_translate_column
...
                 istring_t *strings = anemone_mempool_alloc (mempool, count * sizeof (istring_t));
-                int offset = 1;
                 for (int i = 0; i != count; ++i) {
-                    int len = data.a.n[i];
-                    offset = zebra_translate_table (mempool, elem_start, len, (void**)(strings + i), &data.a.table);
-                    elem_start += len;
+                    int64_t index = i + elem_start;
+                    int64_t start = data.a.n[index]   - data.a.n[0];
+                    int64_t len   = data.a.n[index+1] - data.a.n[index];
+                    offset = zebra_translate_array (mempool, start, len, (void**)(strings + i), &data);
```
amosr: this function looks right

amosr: can we add an assertion that

```
data.a.table.columns_count == 1 &&
data.a.table.columns[0].type == ZEBRA_BYTE
```

and if it's false just die really badly saying we don't support that input format

data/sea/54-zebra.h
```diff
@@ -206,22 +270,42 @@ void zebra_collect_state (zebra_config_t *cfg, zebra_state_t *state)
     free (state);
```
amosr: don't need to free the buffer, entity_fact_offset etc? (I don't think this is important)

data/sea/54-zebra.h
tranma: yeah, that sounds good

data/sea/54-zebra.h
```diff
@@ -206,22 +270,42 @@ void zebra_collect_state (zebra_config_t *cfg, zebra_state_t *state)
...
+    /* keep going if we have not finished reading all facts for this entity */
+    for (int64_t i = 0; i != state->attribute_count; ++i) {
+        if (state->entity_fact_offset[i] != state->entity_fact_count[i]) {
+            return zebra_read_step (piano, state, entity);
+        }
+        state->entity_fact_count [i] = 0;
+        state->entity_fact_offset[i] = 0;
+    }
+
+    return 0;
```
amosr: yeah this is weird control flow, but I think something like this would be cleaner:

```
while (true) {
  read_entity
  all = true;
  for (...) if (fact_offset != fact_count) all = false;
  if (!all) break;
}

for (...) fact_offset = fact_count = 0;
```

amosr: you've hand fused the two loops!

jystic:
How's this one going?

tranma:
@jystic it's ok on its own but I'd like to finish & merge the zebra quickcheck tests in `topic/input-test` first

data/sea/54-zebra.h
tranma: haha I guess it was a bit too haskelly. I made it 2 loops

tranma:
this should be ok now with the tests in

amosr:
👍

PR #497, opened 2017-02-01
Use the right attribute index for Zebra

tranma:
The attribute index given to Zebra was just the ordering in `[SeaProgramState]`, which is compiled from `Map Attribute AvalancheProgram`. If some attribute is never used in a virtual definition, we will get a zebra "columns mismatched" error, or a zebra translate error, e.g.

```
[feature.a_thing]
expression="feature a ~> .."

[feature.c_thing]
expression="feature c ~> .."

[fact.a]
encoding="..."

[fact.b]
encoding="..."

[fact.c]
encoding="..."
```

This just passes the list of attributes along so the zebra code can figure out the right index.

! @amosr @jystic


icicle-compiler/src/Icicle/Sea/IO/Zebra.hs
```diff
@@ -29,30 +31,37 @@ seaOfZebraDriver _ states = return $ seaOfDefRead states
...
+  , "            , fleet->mempool"
+  , "            , chord_count"
+  , "            , " <> i
+  , "            , fleet->iprogram_" <> n
+  , "            , entity);"
   , "if (error) return error;"
   ]
   where
-    i = pretty ix
+    n = pretty (stateName state)
```
jystic: fancy 👍

jystic: is `stateName` sanitised to be C identifier compatible anywhere?

tranma: it's an int

jystic: oh right hah ok :)

jystic:
📻

PR #499, opened 2017-02-21
quickcheck zebra reading

tranma:
This adds a test that just savages some random Icicle facts to Zebra, reads them into an input struct, and peeks the result.

depends on https://github.com/ambiata/zebra/pull/64

! @jystic @amosr 
/jury approved @jystic @amosr


icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,374 @@
...
+
+        SumT ErrorT t -> do
+          (bools, vs) <- List.unzip <$> mapM (fromVSumError (tableOf t)) rows
+          pure $ [ Zebra.IntColumn (Storable.fromList bools) ] <> concat vs
+
+        SumT l r -> do
+          (bools, ls, rs) <- List.unzip3 <$> mapM (fromVSum (tableOf l) (tableOf r)) rows
+          lcols <- mergeValues l ls
+          rcols <- mergeValues r rs
+          pure $ [ Zebra.IntColumn (Storable.fromList bools) ] <> lcols <> rcols
```
jystic: nice 👍

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,374 @@
...
+        --       ]
+
+        OptionT t -> do
+          (bools, vs) <- List.unzip <$> mapM (fromVOption (tableOf t)) rows
+          cols <- mergeValues t vs
+          pure $ [ Zebra.IntColumn (Storable.fromList bools) ] <> cols
+
+        SumT ErrorT t -> do
+          (bools, vs) <- List.unzip <$> mapM (fromVSumError (tableOf t)) rows
+          pure $ [ Zebra.IntColumn (Storable.fromList bools) ] <> concat vs
```
jystic: this is how icicle does it, but not sure that this specialisation should end up in a zebra file

should probably be `(tags, errs, vs)` instead?

jystic: or just get rid of this case and let `SumT` handle it maybe

icicle-compiler/test/test.hs
```diff
@@ -65,6 +66,7 @@ main
         , Icicle.Test.Avalanche.Melt.tests
 
         , Icicle.Test.Sea.Psv.tests
+--        , Icicle.Test.Sea.Zebra.tests
```
jystic: not enabled?

jystic:
It might be easier to convert the icicle type to a `Schema` then convert the icicle value to a zebra `Value`, then you can use:
```hs
fromRow :: Schema -> Value -> Either (TableError Schema) (Table Schema)
```

might be a safer test because otherwise you're testing your encoding of tables vs the icicle generated code.

tranma:
true, that is less savage actually -- I'll convert to a `Schema`

tranma:
using zebra Schema now

icicle-compiler/test/Icicle/Test/Arbitrary/Core.hs
```diff
@@ -178,19 +178,20 @@ instance Arbitrary ValType where
...
-         , TimeT
-         , StringT ]
+   oneof_sized
+         [ pure IntT
+         , pure UnitT
+         , pure BoolT
+         , pure TimeT
+         , pure StringT
+         , StructT <$> arbitrary
+         , OptionT <$> arbitrary
```
jystic: the stuff in the "small" list should probably be non-recursive, otherwise you might generate `OptionT (OptionT (OptionT etc` and never terminate

tranma: right ok. This was because I wanted more structs

jystic: they should all be equally likely except when the size is 1 or less

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+                             , argPtr (Zebra.unCEntity c_entity) ]
+
+                     when (e2 /= nullPtr) $ do
+                       err <- peekCString (castPtr e2)
+                       fail $ "failed to read entity: " <> err
+
+                 -- iprogram: { *mempool, input, ... }
+                 -- input: { *chord_time, *fact_count, *tombstone, *input_start, ... }
+                 programs0_ptr <- peekWordOff fleet_ptr 4
+                 tombstones_ptr <- peekWordOff programs0_ptr 3
```
jystic: Some serious pointer twiddling :D

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+    pure Schema.Int
+
+  StringT ->
+    pure $ Schema.Array Schema.Byte
+
+  ErrorT ->
+    pure Schema.Int
+
+  UnitT ->
+    Nothing
```
jystic: This can be mapped to `Schema.Struct Boxed.empty`

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+  StructT struct
+    -- An empty icicle struct is not unit (unlike zebra), it has no input variable.
+    | fields <- getStructType struct
+    , not (Map.null fields) ->
+        let
+          fieldOf (f, t) =
+            Schema.Field (Schema.FieldName (nameOfStructField f)) <$> schemaOfType t
+        in Schema.Struct . Boxed.fromList <$> mapM fieldOf (Map.toList fields)
+
+  _ -> Nothing
```
jystic: I think we should be able to map all of `ValType` to zebra? probably doesn't matter for a test I guess

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+zebraOfValue :: ValType -> BaseValue -> Either TestError Zebra.Value
+zebraOfValue ty val = case val of
+  VInt x ->
+    pure . Zebra.Int . fromIntegral $ x
+
+  VDouble x ->
+    pure . Zebra.Double $ x
+
+  VUnit ->
+    pure . Zebra.Int $ 0
```
jystic: same as above

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+
+  VBool x ->
+    pure . Zebra.Bool $ x
+
+  VTime x ->
+    pure . Zebra.Int . fromIntegral . Icicle.packedOfTime $ x
+
+  VString x ->
+    pure . Zebra.ByteArray . ByteString.pack . Text.unpack $ x
```
jystic: s/`ByteString.pack . Text.unpack`/`Data.Text.Encoding.encodeUtf8`/ ?

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+    | SumT t _ <- ty
+    -> Zebra.Enum 0 <$> zebraOfValue t x
+
+  VRight x
+    | SumT _ t <- ty
+    -> Zebra.Enum 1 <$> zebraOfValue t x
+
+  VNone
+    | OptionT t <- ty
+    -> Zebra.Enum 0 <$> zebraOfValue t (defaultOfType t)
```
jystic: This should be `Zebra.Enum 0 (Zebra.Struct Boxed.empty)`

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+    a' <- schemaOfType a
+    b' <- schemaOfType b
+    pure . Schema.Struct . Boxed.fromList $
+      [ Schema.Field (Schema.FieldName "fst") a'
+      , Schema.Field (Schema.FieldName "snd") b' ]
+
+  OptionT t ->
+    let
+      noneOf x =
+        Schema.Variant (Schema.VariantName "none") <$> schemaOfType x
```
jystic: This should be:
`Schema.Variant (Schema.VariantName "none") (Schema.Struct Boxed.empty)`
or
`Schema.Variant (Schema.VariantName "none") <$> schemaOfType UnitT`

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+    | StructT struct <- ty
+    , types <- getStructType struct
+    -> do let vs = Map.elems (Map.intersectionWith (,) types xs)
+          Zebra.Struct . Boxed.fromList <$> mapM (uncurry zebraOfValue) vs
+  VBuf xs
+    | BufT _ t <- ty
+    -> Zebra.Array . Boxed.fromList <$> mapM (zebraOfValue t) xs
+
+  VFactIdentifier x ->
+    pure . Zebra.Int . fromIntegral . getFactIdentifierIndex $ x
```
jystic: This case (fact identifier) is missing from `schemaOfType`

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+      rightOf x =
+        Schema.Variant (Schema.VariantName "right") <$> schemaOfType x
+    in Schema.Enum <$> leftOf a <*> (Boxed.singleton <$> rightOf b)
+
+  MapT k v -> do
+    k' <- schemaOfType' k
+    v' <- schemaOfType' v
+    pure . Schema.Struct . Boxed.fromList $
+      [ Schema.Field (Schema.FieldName "keys") (Schema.Array k')
+      , Schema.Field (Schema.FieldName "vals") (Schema.Array v') ]
```
jystic: This is cool for a test, but maybe we want `Array (Struct [("key", k'), ("val", v')])` for production? not sure just wanted to bring it up for discussion. I can imagine when zebra is pretty printed having the key/value pair together would be useful.

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,449 @@
...
+    | MapT tk tv <- ty
+    -> do keys <- mapM (zebraOfValue tk) (Map.keys x)
+          vals <- mapM (zebraOfValue tv) (Map.elems x)
+          pure . Zebra.Struct . Boxed.fromList $
+            [ Zebra.Array (Boxed.fromList keys), Zebra.Array (Boxed.fromList vals) ]
+  VStruct xs
+    | StructT struct <- ty
+    , types <- getStructType struct
+    -> do let vs = Map.elems (Map.intersectionWith (,) types xs)
+          Zebra.Struct . Boxed.fromList <$> mapM (uncurry zebraOfValue) vs
```
jystic: Might be good to have a guard saying `length vs == length types` or similar

jystic:
💯 for the test

👍 after we resolve the mapping from icicle types to zebra, happy to discuss IRL, maybe it's not so important, but I can see that testing stuff becoming the basis for our production mapping

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
tranma: this is a savage way to let you comment out cases for debugging :D

data/sea/54-zebra.h
```diff
@@ -155,7 +156,7 @@ static int64_t zebra_translate_column
                 istring_t *strings = anemone_mempool_alloc (mempool, count * sizeof (istring_t));
                 int offset = 1;
                 for (int i = 0; i != count; ++i) {
-                    int64_t index = i + elem_start;
+                    int64_t index = i;
```
amosr: I'm not sure about this. I *think* this should include the `elem_start`, but that `elem_start` should not have added `+= len` below, and also the recursive call should use `start` instead of `elem_start`.

the `elem_start` refers to the current level - that is the indices/lengths array. then `start` refers to the nested values, so you need to use it when calling `zebra_translate_table`

I think the current code *would* pass as long as `elem_start = 0` for nested arrays (strings)

does that make sense?

tranma: yes, I think I did that in the other branch -- (reading zebra facts in chunk)

tranma: like this?

```
                      int64_t index = i + elem_start;
                      int64_t start = data.a.n[index]   - data.a.n[0];
                      int64_t len   = data.a.n[index+1] - data.a.n[index];
 -                    offset = zebra_translate_table (mempool, elem_start, len, (void**)(strings + i), &data.a.table);
 -                    elem_start += len;
 +                    offset = zebra_translate_array (mempool, start, len, (void**)(strings + i), &data);
```

amosr: example probably better

```
nested array of strings
  ["beef", "helps", "keep", "order"]

lengths:
  [4, 5, 4, 5]
scan (data.a.n):
  [0, 4, 9, 13, 18]
values (data.a.table):
  "beefhelpskeeporder"

zebra_translate_table(elem_start = 0, count = 1, table)
= ["beef"]

zebra_translate_table(elem_start = 0, count = 2, table)
= ["beef", "helps"]

zebra_translate_table(elem_start = 2, count = 2, table)
= ["keep", "order"]
```

the last one is the most interesting because it needs to start looking at scan array 2 and 3 (`data.a.n`), so to pull out the value "beef" it looks up `scan[2] = 9` for the start index, and `scan[3] = 13` for the end index, meaning the length is `13-9=4`.

icicle-compiler/src/Icicle/Sea/FromAvalanche/Type.hs
```diff
@@ -75,10 +75,16 @@ expandType t
...
  = Set.fromList
+ . auxillary
  . concatMap expandType
  . Set.toList
  . typesOfProgram
 
+auxillary :: [ValType] -> [ValType]
+auxillary [] = []
+auxillary (t@(BufT _ a) : ts) = (ArrayT a : t : auxillary ts)
+auxillary (t : ts) = t : auxillary ts
```
amosr: silly question, no chance of bufs inside other types at this point?

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -0,0 +1,453 @@
...
+                     tombstone <- liftIO $ peekElemOff tombstones_ptr fact_i
+
+                     x <- case errorOfWord tombstone of
+                        ExceptNotAnError -> do
+                          liftIO $ slice buf fact_i
+                          VRight . snd <$> peekOutput buf 0 ty
+                        e ->
+                          pure (VLeft (VError e))
+
+                     peekInputs (x:xs) (index - 1)
```
amosr: this `peekInputs` might be worth moving in to sea eval, wherever `peekOutput` is

data/sea/54-zebra.h
amosr: sorry I wrote that example comment above before I saw your reply

amosr:
:1st_place_medal: wow

tranma:
I'm going to rebase and merge this, ignoring the unmelted value failure (see #500)

PR #502, opened 2017-03-06
Fix melt bugs

tranma:
this fixes two melt/constructor/deadcode bugs:

- don't replace `prim_eq unit unit` with `unit` (needs a default case for `fold1`)
- when we have a binding that doesn't get used and is not removed by dead code, melt will not melt it and so constructor won't get rid of it, and we are left with an unmelted value in the end (see #500)

ideally we should have a pass right after `FromCore` that freshify every name

! @amosr @jystic 
/jury approved @amosr @jystic


amosr:
👍

jystic:
:shipit:

PR #503, opened 2017-03-06
Update zebra

jystic:
! @amosr @tranma 
/jury approved @tranma @amosr

### DIFF
--- a/icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
+++ b/icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
@@ -25,7 +25,7 @@ import           Foreign.C.String
 
 import           System.IO
 
-import qualified Prelude as Savage 
+import qualified Prelude as Savage
 
 import           P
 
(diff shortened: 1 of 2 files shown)


icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -301,73 +297,77 @@ schemaOfType ty = case ty of
...
 
   VLeft x
     | SumT t _ <- ty
-    -> Zebra.Enum 0 <$> zebraOfValue t x
+    -> Value.Enum 0 <$> zebraOfValue t x
 
   VRight x
     | SumT _ t <- ty
-    -> Zebra.Enum 1 <$> zebraOfValue t x
+    -> Value.Enum 1 <$> zebraOfValue t x
```
jystic: We should really add `either`, `left`, `right` combinators to zebra

tranma:
👍

amosr:
👍

PR #505, opened 2017-03-09
Zebra alloc limit

tranma:
This allows an allocation limit (per entity) to be set for Zebra. If the limit is exceeded, we stop and write the entity to a log file. It will not appear in the snapshot/chord.

```
./dist/build/Icicle/icicle query --dictionary-toml test/cli/icicle/t6-zebra.toml --input-zebra test/cli/icicle/t6-input --snapshot 2017-1-1 --output-code foo/code.c --output-dense-psv foo/output.psv --zebra-chunk-size 2 --zebra-alloc-limit 10

$ cat foo/output_dropped.txt
ID00000001
```

! @jystic @amosr 
/jury approved @amosr

### DIFF
--- a/data/sea/54-zebra.h
+++ b/data/sea/54-zebra.h
@@ -8,6 +8,7 @@ typedef struct {
     const char *input_path;
     iint_t output_fd;
     iint_t chord_fd;
+    iint_t drop_fd;
 
     /* outputs */
     const char  *error;
@@ -36,7 +37,7 @@ typedef struct zebra_state {
 
     /* output file descriptor */
     int output_fd;
-    int dropped_entity_fd;
+    int drop_fd;
 
     /* read state */
     int64_t  attribute_count;
@@ -224,10 +225,12 @@ zebra_state_t *zebra_alloc_state (piano_t *piano, zebra_config_t *cfg, int64_t a
     int fd;
     int output_fd = (int)cfg->output_fd;
     int chord_fd  = (int)cfg->chord_fd;
+    int drop_fd   = (int)cfg->drop_fd;
 
     /* System.IO.Handles are in non-blocking mode by default */
     psv_set_blocking_mode (output_fd);
     psv_set_blocking_mode (chord_fd);
+    psv_set_blocking_mode (drop_fd);
 
     /* If we have a piano, we know we are playing a chord */
     int64_t max_chord_count;
@@ -248,6 +251,8 @@ zebra_state_t *zebra_alloc_state (piano_t *piano, zebra_config_t *cfg, int64_t a
     state->output_ptr   = output_ptr;
     state->output_fd    = output_fd;
 
+    state->drop_fd      = drop_fd;
+
     state->attribute_count    = attribute_count;
     state->entity_fact_offset = calloc(attribute_count, sizeof (int64_t));
     state->entity_fact_count  = calloc(attribute_count, sizeof (int64_t));
@@ -263,13 +268,16 @@ void zebra_collect_state (zebra_config_t *cfg, zebra_state_t *state)
 {
     cfg->entity_count = state->entity_count;
     cfg->fact_count = state->fact_count;
-    free (state->output_start);
     free (state->entity_fact_offset);
     free (state->entity_fact_count);
     free (state);
 }
 
 void zebra_write_dropped_entity (zebra_state_t *state, zebra_entity_t *entity) {
+    char *msg = calloc (error_msg_size, 1);
+    const size_t msg_size = snprintf (msg, error_msg_size, "%.*s\n", entity->id_length, entity->id_bytes);
+    write (state->drop_fd, msg, msg_size);
+    free (msg);
 }
 
 ibool_t zebra_limit_exceeded (zebra_state_t *state) {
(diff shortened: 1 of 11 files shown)


icicle-compiler/src/Icicle/Sea/IO/Offset.hs
```diff
@@ -0,0 +1,228 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
+-- * Offsets into statically defined or generated C structs.
+module Icicle.Sea.IO.Offset where
```
jystic: great idea 👍

data/sea/54-zebra.h
```diff
@@ -120,8 +128,8 @@ static ierror_msg_t zebra_translate
     iint_t fact_remaining = fact_count - fact_offset;
     iint_t fact_to_read   = 0;
 
-    if (fact_remaining > zebra_chunk_fact_count) {
-        fact_to_read = zebra_chunk_fact_count;
+    if (fact_remaining > state->chunk_size) {
+        fact_to_read = state->chunk_size;
```
amosr: sorry for being pedantic but I preferred "chunk_fact_count": size makes me think of bytes or continuous things like centimetres

data/sea/54-zebra.h
```diff
@@ -241,23 +251,39 @@ zebra_state_t *zebra_alloc_state (piano_t *piano, zebra_config_t *cfg, int64_t a
...
 
+void zebra_write_dropped_entity (zebra_state_t *state, zebra_entity_t *entity) {
+    char *msg = calloc (error_msg_size, 1);
+    const size_t msg_size = snprintf (msg, error_msg_size, "%.*s\n", entity->id_length, entity->id_bytes);
+    write (state->drop_fd, msg, msg_size);
+    free (msg);
+}
+
+ibool_t zebra_limit_exceeded (zebra_state_t *state) {
+    return state->entity_alloc_count > state->alloc_limit;
```
amosr: this seems very roundabout. I'm guessing you can't just do `state->fleet->mempool` or something because it's not defined at this point?

icicle-compiler/main/icicle.hs
```diff
@@ -194,6 +196,20 @@ pFlagDrop =
     long "drop-to-output" <>
     help "write partial results to dropped-X.txt or normal output"
 
+pZebraChunkSize :: Parser ZebraChunkSize
+pZebraChunkSize =
+  flip option (long "zebra-chunk-size" <> value defaultZebraChunkSize) $
+    readerAsk >>= \s -> case readMaybe s of
+      Just i  -> return (ZebraChunkSize i)
+      Nothing -> readerError "--chunk-size NUMBER_BYTES"
```
amosr: should this be `NUMBER_FACTS`?

icicle-compiler/main/icicle.hs
```diff
@@ -194,6 +196,20 @@ pFlagDrop =
...
+
+pZebraAllocLimit :: Parser ZebraAllocLimit
+pZebraAllocLimit =
+  flip option (long "zebra-alloc-limit" <> value defaultZebraAllocLimit) $
+    readerAsk >>= \s -> case readMaybe s of
+      Just i  -> return (ZebraAllocLimit i)
+      Nothing -> readerError "--zebra-alloc-limit NUMBER_BYTES"
```
amosr: I think bytes is too fine-grained.. in zebra I parsed a double and used that as gigabytes. I also kind of like having the measure in the argument: eg `--zebra-alloc-limit-gb 2`.

just a suggestion

icicle-compiler/src/Icicle/Sea/Eval.hs
```diff
@@ -64,36 +65,45 @@ data PsvStats = PsvStats {
...
+  withWords 10 $ \pState -> do
   input_path <- liftIO $ newCString input
   pokeWordOff pState 0 input_path
   pokeWordOff pState 1 output
   pokeWordOff pState 2 (fromMaybe 0 mchords)
-  pokeWordOff pState 6 (defaultPsvOutputBufferSize)
+  pokeWordOff pState 3 drop_fd
+  pokeWordOff pState 7 (defaultPsvOutputBufferSize)
+  pokeWordOff pState 8 (unZebraChunkSize . zebraChunkSize $ conf)
+  pokeWordOff pState 9 (unZebraAllocLimit . zebraAllocLimit $ conf)
```
amosr: `Offset.zebraConfigAllocLimit` ?

amosr:
👍 comments just suggestions

data/sea/54-zebra.h
tranma: yeah fair enough, `count` is good

data/sea/54-zebra.h
tranma: yeah that's right. There might be a less roundabout way of doing this though

amosr: cool - I was just confused at first but makes sense. I don't see any better way to do it

icicle-compiler/main/icicle.hs
tranma: yeah good idea

tranma:
@nhibberd icicle will be able to drop zebra entities following this PR.

nhibberd:
> icicle will be able to drop zebra entities following this PR.

Thanks

PR #507, opened 2017-03-10
Relax restrictions on nested groups

tranma:
I don't remember why it was restricted like this before (we support nested groups now). Some existing dictionaries won't compile with these restrictions.

! @amosr

### DIFF
--- a/icicle-source/src/Icicle/Source/Checker/Invariants.hs
+++ b/icicle-source/src/Icicle/Source/Checker/Invariants.hs
@@ -34,7 +34,7 @@ invariantQ ctx (Query (c:cs) xfinal)
  = case c of
     Windowed{}
      | allowWindows inv
-     -> goBanWindowAndGroupFold
+     -> goBanWindow
      | otherwise
      -> errBanWindow "Consider moving the window to the start of the query"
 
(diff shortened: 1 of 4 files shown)


icicle-source/src/Icicle/Source/Checker/Invariants.hs
```diff
@@ -48,16 +48,13 @@ invariantQ ctx (Query (c:cs) xfinal)
...
 
     Distinct _ x
-     -> goX x >> goBanWindowAndGroupFold
+     -> goX x >> go
 
     GroupFold _ _ _ x
-     | allowWindows inv
-     -> goX x >> goBanWindowAndGroupFold
-     | otherwise
-     -> errBanGroupFold
```
tranma: @amosr I think it should be stricter than this, but what we had before was too strict?

amosr: something about - we need to ban latests inside groups? hmm

amosr: we might need to dig up the change that made it stricter to see why

tranma: the change actually made it less strict already, from the group in group change: https://github.com/ambiata/icicle/commit/88901cd08b51f4988913c000142ee8f74e5f616e#diff-743f2c2eb9de71059af5941e74e658c6

icicle-source/src/Icicle/Source/Checker/Invariants.hs
```diff
@@ -48,16 +48,13 @@ invariantQ ctx (Query (c:cs) xfinal)
      -> errBanLatest
 
     GroupBy _ x
-     -> goX x >> goBanWindowAndGroupFold
+     -> goX x >> go
```
amosr: I think windows still can't be inside groups or distincts

icicle-source/src/Icicle/Source/Checker/Invariants.hs
amosr: ah. so I think we need to ban latests here, #468

amosr: if we ban latests I think we can reenable the test #488

amosr:
is this superseded by #513 ?

tranma:
yeah this is really the segfault-debugging PR

PR #509, opened 2017-03-10
Fission: allow an attribute to have multiple "compute kernels"..

amosr:
Code gen part of splitting out attributes to allow multiple compute functions.
Will implement fission options as separate PR

! @tranma @jystic


amosr:
am going to merge this as-is: it's a fairly mechanical change, and if something were to go wrong *everything* would go wrong.
will add a test and options to control fission as next PR

PR #510, opened 2017-03-16
Add icicle-test package

tranma:
I wanted to test the output dictionary in glacier. Probably overkill as we will eventually switch to zebra output anyway, but we get a piece of mind in case the ordering of dense psv columns ever changes.

! @jystic 
/jury approved @jystic

### DIFF
--- a/icicle-compiler/test/icicle-test.cabal
+++ b/icicle-compiler/test/icicle-test.cabal
@@ -1,0 +1,58 @@
+name:                  icicle-test
+version:               0.0.1
+license:               AllRightsReserved
+author:                Ambiata <info@ambiata.com>
+maintainer:            Ambiata <info@ambiata.com>
+copyright:             (c) 2015 Ambiata.
+synopsis:              icicle test generators
+category:              System
+cabal-version:         >= 1.8
+build-type:            Simple
+description:           icicle test generators.
+
+library
+  ghc-options:         -Wall
+
+  build-depends:
+                       base                            >= 3          && < 5
+                     , icicle-data
+                     , icicle-source
+                     , icicle-core
+                     , icicle
+                     , ambiata-disorder-core
+                     , ambiata-disorder-corpus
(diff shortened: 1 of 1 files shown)


nhibberd:
Probably worth adding a lock file for it as well

icicle-compiler/test/icicle-test.cabal
```diff
@@ -0,0 +1,58 @@
...
+                     , directory
+                     , exceptions                      == 0.8.*
+                     , filepath                        == 1.4.*
+                     , geniplate-mirror                >= 0.7.2      && < 0.8
+                     , megaparsec                      == 5.0.*
+                     , parsec
+                     , pretty-show                     == 1.6.*
+                     , hashable                        == 1.2.*
+                     , unordered-containers            == 0.2.*
+                     , QuickCheck                      == 2.8.*
```
nhibberd: Should this be a range?

icicle-compiler/test/icicle-test.cabal
```diff
@@ -0,0 +1,58 @@
...
+                     , text
+                     , transformers
+                     , unix                            == 2.7.*
+                     , vector                          == 0.11.*
+                     , semigroups                      >= 0.16       && < 0.19
+
+  exposed-modules:
+                       Icicle.Test.Arbitrary
+                       Icicle.Test.Sea.Utils
+                       Icicle.Test.Foreign.Utils
```
nhibberd: /newline

icicle-compiler/test/icicle-test.cabal
jystic: Yeah for test libraries this should be `>= 2.7 && < 2.10`, for test runners anything is ok

jystic:
👍 on fixups

tranma:
on second thought, no -- would like to keep icicle as self-contained as possible (it's a compiler)

PR #511, opened 2017-03-20
 Add CPP flags so anemone header files can be included

tranma:
… when icicle used as a dependency. Any package that depends on `icicle` will need to define:

```
cpp-options: -DICICLE_DEP
```

This is a bit awkward. If I flip the flag around, and have `-DICICLE_NOT_DEP` defined in `icicle.cabal`, I'm not sure how to turn it off when icicle is built as a dependency.

! @jystic @amosr 
/jury approved @jystic


jystic:
👍

PR #512, opened 2017-03-20
Embed anemone sources as strings

tranma:
this will make it easier for other packages to build `icicle` as a library dependency.

! @jystic @amosr 
/jury approved @jystic


jystic:
👍

PR #513, opened 2017-03-22
Source: typo in invariant checker

tranma:
I broke this too many times, sorry
! @amosr 
/jury approved @amosr


icicle-source/src/Icicle/Source/Checker/Invariants.hs
```diff
@@ -54,7 +54,7 @@ invariantQ ctx (Query (c:cs) xfinal)
      -> goX x >> goBanWindowAndGroupFold
 
     GroupFold _ _ _ x
-     | allowWindows inv
+     | allowGroupFolds inv
      -> goX x >> goBanWindowAndGroupFold
```
amosr: I think we need to ban latest here too

tranma:
fixed for real this time

amosr:
👍

PR #514, opened 2017-03-27
Read and output chord names from piano

tranma:
Output the chord names from piano instead of chord times (comparison will still be on chord times of course).

! @jystic 
/jury approved @jystic


icicle-compiler/src/Icicle/Sea/IO/Offset.hs
```diff
@@ -146,14 +149,23 @@ fleetChordCount = 2
...
+
+fleetChordNameLengths :: Int
+fleetChordNameLengths = 5
+
+fleetChordNameData :: Int
+fleetChordNameData = 6
+
 fleetProgramOf :: Int -> Int
-fleetProgramOf ix = 4 + ix
+fleetProgramOf ix = 7 + ix
```
jystic: :agile: structs

tranma:
chords:
```
marge|2014-06-03
marge|2014-06-04
homer|2014-05-03
```

output:
```
homer|[["CA",0.0],["UT",0.0]]|0|2014-05-03
marge|[["CA",8.0],["UT",7.0]]|5|2014-06-03
marge|[["CA",8.0],["UT",7.0]]|5|2014-06-04
```

icicle-compiler/src/Icicle/Sea/IO/Base.hs
```diff
@@ -209,8 +233,11 @@ seaOfAllocFleet states
  , "{"
  , "    ifleet_t *fleet = calloc (1, sizeof (ifleet_t));"
  , ""
- , "    fleet->max_chord_count = max_chord_count;"
- , "    fleet->chord_times     = calloc (max_chord_count, sizeof(itime_t));"
+ , "    fleet->max_chord_count    = max_chord_count;"
+ , "    fleet->chord_times        = calloc (max_chord_count, sizeof(itime_t));"
+ , "    fleet->chord_name_lengths = calloc (max_chord_count, sizeof(int64_t));"
+ , "    fleet->chord_name_offsets = calloc (max_chord_count, sizeof(int64_t));"
+ , "    fleet->chord_name_data    = (uint8_t*) calloc (chord_label_name_data_size, 1);"
```
jystic: probably don't need to allocate these anymore (except for `chord_times`) after your fixup

tranma: yeah

data/sea/51-piano.h
```diff
@@ -1,7 +1,10 @@
-#include "50-chord.h"
+#include "42-text-conversion.h"
 
 typedef struct piano piano_t;
 
+/* FIXME: arbitrary decision for total chord label size (for an entity) */
+static const int64_t chord_label_name_data_size = 1024;
```
jystic: I think this isn't required either anymore

tranma:
this is good to go whenever it's needed

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
```diff
@@ -96,29 +98,29 @@ seaOfWriteFleetOutput config whitelist states = do
...
 
 timeFmt :: Doc
 timeFmt = "%04\" PRId64 \"-%02\" PRId64 \"-%02\" PRId64 \"T%02\" PRId64 \":%02\" PRId64 \":%02\" PRId64 \"Z"
 
 outputChord :: Doc
 outputChord
-  = outputValue "string" ["chord_time", "chord_size"]
-
+  = vsep
+  [ "if (chord_name_length != 0) {"
```
jystic: It might be better to have this as `chord_name != NULL` and change the code above to set `chord_name = NULL`  in the snapshot case.

It probably won't happen, but the empty string seems like a perfectly valid chord name, so it would be odd if that caused the column to be dropped off.

jystic:
👍 after fixing the empty chord name issue

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
tranma: right, good catch!

PR #517, opened 2017-03-31
re-enable put_mutable_diff

tranma:
this test stills sometimes segfault, which is when this assertion is violated:

```
...put_mutable...
    anemone_block_t *thing = pool->last;\
    ARRAY_PAYLOAD(t,x)[ix] = v;\
    assert (thing == pool->last);\
```

pretty weird...
/jury approved @jystic


jystic:
Would be interesting to try:
```
#if ICICLE_ASSERT
if (thing != pool->last) {
  // print out thing + pool->last
}
#endif

amosr:
oh.. so I only enabled icicle assertions (`-DICICLE_ASSERT=1`) for the Sea/Psv tests. might be worth enabling them here and elsewhere if you haven't already

tranma:
this should be good now too (the problem was in the test, poking an array from haskell to C without allocating a rounded up number of bytes)

jystic:
👍

PR #519, opened 2017-04-10
Update to Zebra v3

tranma:
I'll need to make up a new CLI test I think.

! @jystic @amosr 
/jury approved @amosr

### DIFF
--- a/data/sea/54-zebra.h
+++ b/data/sea/54-zebra.h
@@ -51,29 +51,23 @@ typedef struct zebra_state {
 
 static int64_t zebra_attribute_count ();
 
-static int64_t zebra_translate_table
-  ( anemone_mempool_t *mempool
-  , iint_t elem_start
-  , iint_t count
-  , void **dst
-  , const zebra_table_t *table );
-
-static int64_t zebra_translate_column
-  ( anemone_mempool_t *mempool
-  , iint_t elem_start
-  , iint_t count
-  , void **dst
-  , const zebra_column_t *src );
-
 static ierror_msg_t zebra_read_entity
-  ( piano_t *piano
-  , zebra_state_t *state
-  , zebra_entity_t *entity );
+    ( piano_t *piano
+    , zebra_state_t *state
+    , zebra_entity_t *entity );
 
 ierror_msg_t zebra_snapshot_step
-  ( piano_t *piano
-  , zebra_state_t *state
-  , zebra_entity_t *entity );
+    ( piano_t *piano
+    , zebra_state_t *state
+    , zebra_entity_t *entity );
+
+static int64_t zebra_translate_table
+    ( anemone_mempool_t *mempool
+    , int64_t elem_start
+    , int64_t elem_count
+    , void **dst
+    , const zebra_table_t *src );
+
 
 /* map a zebra entity to sea fleet inputs */
 static ierror_msg_t zebra_translate
@@ -83,25 +77,19 @@ static ierror_msg_t zebra_translate
     , anemone_mempool_t    *mempool
     , int                   attribute_ix
     , itime_t               chord_time
-    , iint_t               *new_fact_count
+    , int64_t               *new_fact_count
     , ierror_t            **new_tombstone
     , itime_t             **new_fact_time
-    , iint_t                columns_to_fill
+    , int64_t                columns_to_fill
     , void                **dst
     , const zebra_entity_t *src )
 {
+    const int64_t attribute_count = src->attribute_count;
+    if (attribute_ix >= attribute_count) {
+        fprintf (stderr, "Fatal error: attribute index (%" PRId64 ") is out of bounds for attribute count (%" PRId64 ")\n", attribute_ix, attribute_count);
+    }
+
     zebra_attribute_t *attribute = &src->attributes[attribute_ix];
-    iint_t columns_count = attribute->table.column_count;
-
-    /* zebra columns must match the number of input variables */
-    if (columns_count != columns_to_fill) {
-        return ierror_msg_format
-          ( "zebra_translate: attribute_ix=%d, column_count=%d, columns_to_fill=%d\n"
-          , attribute_ix
-          , columns_count
-          , columns_to_fill );
-     }
-
     zebra_table_t *table = &attribute->table;
     int64_t row_count = table->row_count;
     int64_t fact_count = 0;
@@ -112,7 +100,7 @@ static ierror_msg_t zebra_translate
         itime_t zebra_chord_time = itime_to_epoch_seconds (chord_time);
 
         /* the number of facts before chord time for this entity */
-        for (iint_t i = 0; i != row_count; ++i) {
+        for (int64_t i = 0; i != row_count; ++i) {
             itime_t time = attribute->times[i];
             if (zebra_chord_time <= time) {
                 break;
@@ -125,16 +113,15 @@ static ierror_msg_t zebra_translate
     }
 
     /* number of unread facts for this entity */
-    iint_t fact_offset    = state->entity_fact_offset[attribute_ix];
-    iint_t fact_remaining = fact_count - fact_offset;
-    iint_t fact_to_read   = 0;
+    int64_t fact_offset    = state->entity_fact_offset[attribute_ix];
+    int64_t fact_remaining = fact_count - fact_offset;
+    int64_t fact_to_read   = 0;
 
     if (fact_remaining > state->chunk_fact_count) {
         fact_to_read = state->chunk_fact_count;
     } else {
         fact_to_read = fact_remaining;
     }
-    //printf ("zebra: fact_offset = %d, fact_remaining = %d, fact_to_read = %d\n", fact_offset, fact_remaining, fact_to_read);
 
     state->fact_count                       += fact_to_read;
     state->entity_fact_offset[attribute_ix] += fact_to_read;
@@ -142,7 +129,7 @@ static ierror_msg_t zebra_translate
     /* temporary fix until icicle uses the same time as zebra */
     itime_t *sadface_times = anemone_mempool_alloc (mempool, fact_to_read * sizeof (itime_t));
 
-    for (iint_t i = 0; i != fact_to_read; ++i) {
+    for (int64_t i = 0; i != fact_to_read; ++i) {
         sadface_times[i] = itime_from_epoch_seconds (attribute->times[i + fact_offset]);
     }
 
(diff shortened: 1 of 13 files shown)


data/sea/54-zebra.h
```diff
@@ -74,6 +59,68 @@ ierror_msg_t zebra_snapshot_step
...
+
+
+static uint64_t zebra_translate_table
+    ( anemone_mempool_t *mempool
+    , int64_t elem_start
+    , int64_t elem_count
+    , void **dst
+    , const zebra_table_t *src )
+{
+    const int64_t row_count = src->row_count;
```
amosr:
```
assert (elem_start + elem_count <= row_count)
```
perhaps

data/sea/54-zebra.h
```diff
@@ -74,6 +59,68 @@ ierror_msg_t zebra_snapshot_step
...
+
+        case ZEBRA_TABLE_ARRAY: {
+            int64_t offset = 0;
+            for (int64_t i = 0; i != row_count; ++i) {
+                offset += zebra_translate_column
+                    ( mempool
+                    , elem_start
+                    , elem_count
+                    , dst + offset
+                    , data._array.values + i );
```
amosr: the row_count here is the number of elements, not the number of columns - the table only has one top-level column, and this loop should be in the ZEBRA_COLUMN_STRUCT or something

tranma: ah ok

data/sea/54-zebra.h
```diff
@@ -154,68 +189,164 @@ static ierror_msg_t zebra_translate
...
+            const int64_t elem_index = elem_start + elem;
+            const int64_t tag = src[elem_index];
+            if (tag > bit) {
+                bools[elem] = 1;
+            }
+        }
+
+        *(dst + bit) = bools;
+    }
+    return tag_max;
```
amosr: 🦀

data/sea/54-zebra.h
```diff
@@ -154,68 +189,164 @@ static ierror_msg_t zebra_translate
...
+
+static uint64_t zebra_translate_column_nested
+    ( anemone_mempool_t *mempool
+    , int64_t elem_start
+    , int64_t elem_count
+    , void **dst
+    , const int64_t *indices
+    , const zebra_table_t *src )
+{
+    int64_t offset = 1;
```
amosr: ```assert src->tag == BINARY```
?

data/sea/54-zebra.h
```diff
@@ -154,68 +189,164 @@ static ierror_msg_t zebra_translate
...
+
+        case ZEBRA_COLUMN_ENUM: {
+            uint64_t offset = zebra_translate_column_tags
+                ( mempool
+                , elem_start
+                , elem_count
+                , data._enum.columns.count
+                , dst
+                , data._enum.tags );
+            offset += zebra_translate_named_column
```
amosr: I think this needs to loop over from 0 to column_count

data/sea/54-zebra.h
```diff
@@ -154,68 +189,164 @@ static ierror_msg_t zebra_translate
...
+                ( mempool
+                , elem_start
+                , elem_count
+                , dst + offset
+                , &data._enum.columns );
+            return offset;
+        }
+
+        case ZEBRA_COLUMN_STRUCT: {
+            return zebra_translate_named_column
```
amosr: again, loop from 0 to column_count

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -369,29 +358,29 @@ schemaOfType ty = case ty of
...
 
   PairT a b -> do
     a' <- schemaOfType a
     b' <- schemaOfType b
-    pure . Schema.Struct . Boxed.fromList $
-      [ Schema.Field (Schema.FieldName "fst") a'
-      , Schema.Field (Schema.FieldName "snd") b' ]
+    pure . Schema.Struct . Cons.fromNonEmpty $
+           Schema.Field (Schema.FieldName "fst") a'
+      :| [ Schema.Field (Schema.FieldName "snd") b' ]
```
amosr: maybe ```Cons.from2 (Schema.Field ...fst...) (Schema.Field ...snd...)```

data/sea/54-zebra.h
```diff
@@ -154,68 +189,164 @@ static ierror_msg_t zebra_translate
...
-        //printf ("zebra_translate_table: offset = %d\n", offset);
-        offset += zebra_translate_column (mempool, elem_start, count, dst + offset, cols + i);
+    /* enum B = F | T
+       ==>
+       bool is_true
+       ==>
+       F == tag 0 == 0
+       T == tag 1 == 1
+
+       enum E = A | B | C | D | E
```
jystic: Do we have types like this in icicle at the moment? I wonder if we'd be better off just erroring out if it's not option or either?

I suspect if we wanted to support these kind of enums we would need to change icicle to use the same tagged approach that zebra uses, not sure :man_shrugging: 

The issue is that I can see them getting used for some like state (e.g. WA, NSW, QLD, etc) and this bool encoding wouldn't be so good for something like that.

tranma: we don't have enums like zebra at the moment, just bool flags like this for options and eithers. I don't know whether we want to bake this in from the surface language, or just do this conversion in the convert to C stage...

jystic: I guess I was wondering if we should reject zebra schemas which have sum types with more than two variants for the time being? If we ever added full blown algebraic data types to icicle, I think we'd have to change the runtime representation to tags at that time.

jystic: I think it's ok for zebra to support types which icicle doesn't initally

erikd-ambiata: > I guess I was wondering if we should reject zebra schemas which have sum types with more than two variants for the time being?

That would be OK for zebra as a PSV replacement because `Maybe a` with `a` being a primitive is all we need.

tranma: I don't mind either error-ing out when encountering more than two variants, or putting a FIXME here to change the icicle runtime representation of variants

tranma: I'll error out, this means you can't have the true boolean type `true | false | null` :)

jystic: > this means you can't have the true boolean type `true | false | null` :)

for now :)

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -402,23 +391,29 @@ schemaOfType ty = case ty of
...
-    k' <- schemaOfType' k
-    v' <- schemaOfType' v
-    pure . Schema.Struct . Boxed.fromList $
-      [ Schema.Field (Schema.FieldName "keys") (Schema.Array k')
-      , Schema.Field (Schema.FieldName "vals") (Schema.Array v') ]
+    k' <- schemaOfType k
+    v' <- schemaOfType v
+    pure . Schema.Nested . Schema.Array . Schema.Struct . Cons.fromNonEmpty $
+           Schema.Field (Schema.FieldName "key") k'
+      :| [ Schema.Field (Schema.FieldName "val") v' ]
```
jystic: I wonder if we should represent a Map as a Map now?

tranma: yeah I was wondering that too. I guess we should because otherwise we're not testing the map table type at all

jystic: yeah, I think we might as well, when it comes to outputting zebra, writing out real maps is what we want I think

data/sea/54-zebra.h
```diff
@@ -154,68 +185,142 @@ static ierror_msg_t zebra_translate
...
+/* Translate Zebra enum tags to Icicle bool flags, e.g. is_some, is_right, etc.
+   We only allow at most two variants since Icicle's runtime representation
+   of enums is unsuitable for more.
+ */
+static uint64_t zebra_translate_column_tags
+    ( const int64_t variant_count
+    , void **dst
+    , const int64_t *src )
+{
+    assert (variant_count <= 2);
```
jystic: Is the intention that this is `assert` instead of `IASSERT` ?

If the intention is `assert` maybe we'd be better off saying `if (variant_count <= 2)` then printing a descriptive error and exiting.

amosr: I think these should always be on, but I suggest worry about cleaning up messages in next PR? this is a big enough chunk as-is

jystic: I just worry it will get forgotten, and someone will get a cryptic error in production.

I wasn't thinking of anything too rigorous, just something like this:
```c
if (variant_count > 2) {
    fprintf (stderr, "Fatal error: found an enum in the zebra input with more than two variants (it had %" PRId64 " variants)\n", variant_count);
    exit (1);
}
```

tranma: yeah the intention is to always check, I'll `exit` rather than `assert`

data/sea/54-zebra.h
```diff
@@ -154,68 +185,142 @@ static ierror_msg_t zebra_translate
...
+
+    for (int64_t i = 0; i != columns_count; ++i) {
+        offset += zebra_translate_column (mempool, elem_start, elem_count, dst + offset, columns + i);
     }
 
     return offset;
 }
 
-static int64_t zebra_translate_column
+static uint64_t zebra_translate_column
```
jystic: I think you should probably stick with signed numbers, unsigned arithmetic is not a fun place to be

tranma: yeah agreed, I changed my mind almost as soon as I made these changes

icicle-compiler/test/Icicle/Test/Sea/Zebra.hs
```diff
@@ -402,23 +392,27 @@ schemaOfType ty = case ty of
         Schema.Variant (Schema.VariantName "left")  <$> schemaOfType x
       rightOf x =
         Schema.Variant (Schema.VariantName "right") <$> schemaOfType x
-    in Schema.Enum <$> leftOf a <*> (Boxed.singleton <$> rightOf b)
+    in do
+      a' <- leftOf a
+      b' <- rightOf b
+      pure . Schema.Enum . Cons.fromNonEmpty $
+        a' :| [b']
```
jystic: I think all of these should be `Cons.from2`

data/sea/54-zebra.h
```diff
@@ -194,16 +194,15 @@ static int64_t zebra_translate_column_tags
     , void **dst
     , const int64_t *src )
 {
-    if (variant_count <= 2) {
-        fprintf (stderr, "Fatal error: found an enum with %" PRId64 " variants. Icicle does not support Zebra enums with more than two variants.\n");
+    if (variant_count > 2) {
+        fprintf (stderr, "Fatal error: found an enum with %lld variants. Icicle does not support Zebra enums with more than two variants.\n", variant_count);
```
jystic: not sure if `%lld` will work on the build box?

tranma: ah didn't think of that, I just liked the look better

jystic: yeah it sucks 100% having to use `PRId64` I wish they would add something nicer :(

data/sea/54-zebra.h
```diff
@@ -154,68 +185,129 @@ static ierror_msg_t zebra_translate
...
+
+        case ZEBRA_COLUMN_REVERSED: {
+            return zebra_translate_column
+                ( mempool
+                , elem_start
+                , elem_count
+                , dst
+                , data._reversed.column );
+        }
+
```
jystic: Maybe we should crash if we ever get here?

jystic:
Haskell code looks good after changing all the `Cons.fromNonEmpty`

data/sea/54-zebra.h
tranma: right, I guess merge should never leave this

tranma:
this should be good now btw. I'll abstain from zebra cli tests now as we have some quickcheck for it, and psv will eventually go via zebra so psv cli tests can become zebra tests then.

data/sea/54-zebra.h
```diff
@@ -154,68 +190,125 @@ static ierror_msg_t zebra_translate
...
+   We only allow at most two variants since Icicle's runtime representation
+   of enums is unsuitable for more.
+ */
+static int64_t zebra_translate_column_tags
+    ( const int64_t variant_count
+    , void **dst
+    , const int64_t *src )
+{
+    if (variant_count > 2) {
+        fprintf (stderr, "Fatal error: found an enum with %lld variants. Icicle does not support Zebra enums with more than two variants.\n", variant_count);
```
amosr: (`PRId64`)

data/sea/54-zebra.h
```diff
@@ -154,68 +190,125 @@ static ierror_msg_t zebra_translate
...
 
+    int64_t offset = 1;
+
+    for (int i = 0; i != elem_count; ++i) {
+        const int64_t elem_index = elem_start + i;
+        const int64_t table_start = indices[elem_index] - indices[0];
+        const int64_t table_row_count = indices[elem_index + 1] - indices[elem_index];
+
+        /* read the inner tables */
+        void **target = anemone_mempool_alloc (mempool, elem_count * 8);
```
amosr: hmm, should this allocation be outside the loop?

amosr:
confused / not sure about the nested case, but rest looks good to me

data/sea/54-zebra.h
tranma: yeah it should... this is just allocating the pointers for the nested table, right now we only have the binary case anyway

tranma:
my (un-PR'd) zebra output branch is on top of this so will push this through first

data/sea/54-zebra.h
```diff
@@ -155,73 +142,181 @@ static ierror_msg_t zebra_translate
...
+                ( mempool
+                , elem_start
+                , elem_count
+                , dst
+                , data._nested.indices
+                , &data._nested.table );
+        }
+
+        case ZEBRA_COLUMN_REVERSED: {
+            fprintf (stderr, "Fatal error: encountered a reversed Zebra column.\n");
```
amosr: this should be trivial to implement, I think - just
```
return zebra_translate_column(... data. _reversed.column);
```

icicle-compiler/test/cli/zebra/run
```diff
@@ -6,7 +6,7 @@ echo "Separate compilation + Zebra"
 
 common_args="--snapshot 2010-01-01"
 dict=test/cli/zebra/dictionary.toml
-in=test/cli/zebra/input.zbin2
+in=test/cli/zebra/input.zbin3
```
amosr: I think, according to what @jystic said, you should be able to test with both zbin2 and zbin3 now?

data/sea/54-zebra.h
tranma: zebra decode shouldn't let this get here anyway

icicle-compiler/test/cli/zebra/run
tranma: ah right, I'll run both then

data/sea/54-zebra.h
```diff
@@ -155,73 +142,181 @@ static ierror_msg_t zebra_translate
...
-        offset += zebra_translate_column (mempool, elem_start, count, dst + offset, cols + i);
+    /* Only read nested tables of type binary now.
+       FIXME: slice out nested tables, when we need to support them. */
+    if (src->tag != ZEBRA_TABLE_BINARY) {
+        fprintf (stderr, "Fatal error: found a nested table with tag %ud. Icicle only supports nested binary table for now.\n", src->tag);
+        exit(1);
     }
 
+    int64_t offset = 1;
+    void **target = anemone_mempool_alloc (mempool, elem_count * 8);
```
amosr: should this be an ```iarray_istring_create (elem_count)``` so it stores the size in the header?

data/sea/54-zebra.h
```diff
@@ -155,73 +142,181 @@ static ierror_msg_t zebra_translate
...
+
+    for (int i = 0; i != elem_count; ++i) {
+        const int64_t elem_index = elem_start + i;
+        const int64_t table_start = indices[elem_index] - indices[0];
+        const int64_t table_row_count = indices[elem_index + 1] - indices[elem_index];
+
+        /* read the inner tables */
+        zebra_translate_table (mempool, table_start, table_row_count, target + i, src);
+        *dst = target;
```
amosr: move this statement outside the loop?

data/sea/54-zebra.h
amosr: unless it should actually just be allocating `elem_count + 1` bytes and copying them over? I'm confused

data/sea/54-zebra.h
amosr: why not?.. I thought if the column schema was reversed, it would have a reversed here

data/sea/54-zebra.h
amosr: right. never mind this

data/sea/54-zebra.h
tranma: afaik Reversed was meant for the merge, when Icicle gets the entity it should be in the reversed order already

data/sea/54-zebra.h
tranma: these are just strings, we copy each of the strings with `zebra_translate_table`, e.g.:

```
indices: [6,12,15]
bytes: string1string2foo
```

just relying on the `ZEBRA_TABLE_BINARY` case in `zebra_translate_table` to copy the strings `string1`, `string2` and `foo`

data/sea/54-zebra.h
tranma: yup, good catch

amosr:
👍

data/sea/54-zebra.h
```diff
@@ -155,73 +142,186 @@ static ierror_msg_t zebra_translate
...
+        case ZEBRA_COLUMN_REVERSED: {
+            return zebra_translate_column
+                ( mempool
+                , elem_start
+                , elem_count
+                , dst
+                , &data._reversed );
+        }
+
     }
```
jystic: There's a few places where we are switching on a `tag` and not handling the case where we don't know what the tag is. We need to return an error code in that case, or at the very least print an error and call `exit(1)`.

tranma: will do, sorry for the savage C :)

PR #520, opened 2017-04-12
Fission: add test, fix ups, hardcode limit to 10 queries per kernel

amosr:


icicle-compiler/src/Icicle/Compiler.hs
```diff
@@ -214,12 +214,13 @@ sourceConvert d q
 
 coreOfDictionary :: Source.IcicleCompileOptions
                  -> Dictionary
-                 -> Either Error (Map Attribute (Source.CoreProgramUntyped Source.Var))
+                 -> Either Error (Map Attribute (NonEmpty (Source.CoreProgramUntyped Source.Var)))
 coreOfDictionary opts dict = do
```
amosr: do any external things use this function?

icicle-compiler/src/Icicle/Sea/FromAvalanche/State.hs
```diff
@@ -220,10 +221,15 @@ defOfFactField (name, ty)
...
- <> defOfVar 0 t     (pretty resPrefix <> pretty n) <> semi
+defOfResumable :: SeaProgramCompute -> (Text, ValType) -> Doc
+defOfResumable compute (n, t)
+ =  defOfVar 0 BoolT (pretty hasPrefix <> nameOfResumable compute (pretty n)) <> semi <> line
+ <> defOfVar 0 t     (pretty resPrefix <> nameOfResumable compute (pretty n)) <> semi
+
+nameOfResumable :: SeaProgramCompute -> Doc -> Doc
+nameOfResumable compute n
+ = let (i,j) = stateComputeName compute
+   in  pretty i <> "_" <> pretty j <> "_" <> n
```
amosr: I renamed the resumables to have the index of the attribute and compute kernel. the attribute is superfluous but it was already there.
so the attribute info now has something like:
```
ibool_t has_0_1_fridge;
istring_t res_0_1_fridge;
```
instead of just
```
ibool_t has_fridge;
istring_t res_fridge;
```

I didn't rename the outputs because I believe they should be unique. is that correct?

icicle-source/src/Icicle/Encoding.hs
```diff
@@ -383,7 +385,7 @@ renderOutputValue val
    IntValue v
     -> return $ T.pack $ show v
    DoubleValue v
-    -> return $ T.pack $ show v
+    -> return $ T.pack $ showFFloat Nothing v ""
```
amosr: this is to get the double printing same as PSV output

icicle-source/src/Icicle/Compiler/Source.hs
```diff
@@ -112,12 +114,20 @@ freshNamer prefix
...
 defaultCompileOptions :: IcicleCompileOptions
-defaultCompileOptions = IcicleCompileOptions Inline.defaultInline Check.defaultCheckOptions
+defaultCompileOptions = IcicleCompileOptions Inline.defaultInline Check.defaultCheckOptions defaultFusionOptions
+
+data FusionOptions = FusionOptions
+  { fusionMaximumPerKernel :: Int
+  }
+
+defaultFusionOptions :: FusionOptions
+defaultFusionOptions = FusionOptions 10
```
amosr: hardcoded to 10 for now, but will add command-line options next PR

amosr:
! @jystic @tranma

icicle-compiler/src/Icicle/Compiler.hs
tranma: yes, ice does

tranma:
`prop_psv_fission` looks right. For next PR it would be nice to have it generates an arbitrary number of programs to fuse too (with an arbitrary number kernel limit)
👍

PR #522, opened 2017-04-26
fission

amosr:


amosr:
A surprising amount of time is being spent in the C compiler for all the lines setting has to false (several thousand of them)
```
has_X = ifalse;
has_Y = ifalse;
has_Z = ifalse;
```

I've reorganised the struct so all the has fields are next to each other, which lets us memset them all to false. This shaves six or so minutes off C compiler part for a large-ish file:

```
$ time gcc -O1 -c middata100.c

real	7m17.728s
user	7m14.880s
sys	0m1.848s

$ time gcc -O1 -c middata-memsetout-100.c

real	1m19.556s
user	1m18.687s
sys	0m0.810s
```

tranma:
not sure what impact this change would have on the cache locality as most of the code is `if has_x then res_x = ...` but good idea anyway 👍

amosr:
fission significantly reduces memory pressure in icicle, too:

```
$ icicle compile ... --fuse-maximum-per-kernel 10 +RTS -s
icicle: starting compilation
icicle: compilation time = 22.33s
  33,394,708,120 bytes allocated in the heap
   3,255,607,752 bytes copied during GC
      71,912,256 bytes maximum residency (31 sample(s))
       4,106,304 bytes maximum slop
             365 MB total memory in use (75 MB lost due to fragmentation)

$ icicle compile ... --fuse-maximum-per-kernel 100 +RTS -s

icicle: starting compilation
icicle: compilation time = 59.86s
  43,926,123,384 bytes allocated in the heap
  12,538,360,552 bytes copied during GC
     679,501,640 bytes maximum residency (30 sample(s))
       5,874,184 bytes maximum slop
            1721 MB total memory in use (0 MB lost due to fragmentation)



$ icicle compile ... --fuse-maximum-per-kernel 1000 +RTS -s
icicle: starting compilation
icicle: compilation time = 193.13s
  73,034,194,824 bytes allocated in the heap
  50,933,269,800 bytes copied during GC
   4,741,488,080 bytes maximum residency (45 sample(s))
      38,377,432 bytes maximum slop
           10366 MB total memory in use (0 MB lost due to fragmentation)
```

amosr:
for the C compilation, fission alone made it slightly faster, but not significantly. I didn't notice any cases where compilation was longer with fission

icicle-compiler/src/Icicle/Sea/FromAvalanche/State.hs
```diff
@@ -221,16 +234,29 @@ defOfFactField (name, ty)
...
 
+nameOfResumableHasFlagsStart :: SeaProgramCompute -> Doc
+nameOfResumableHasFlagsStart compute
+ = let (i,j) = stateComputeName compute
+   in  "has_flags_start_" <> pretty i <> "_" <> pretty j
+
+nameOfResumableHasFlagsEnd :: SeaProgramCompute -> Doc
+nameOfResumableHasFlagsEnd compute
+ = let (i,j) = stateComputeName compute
+   in  "has_flags_end_" <> pretty i <> "_" <> pretty j
```
jystic: 😆 *things you can do in C*

PR #525, opened 2017-05-11
Fix Zebra attribute count

tranma:
This value (passed dynamically to Zebra) was wrong when the dictionary is already compiled. A case of a bad, careless default on the Haskell side.

Fixed by setting it in the generated Zebra code.

! @amosr @jystic 
/jury approved @jystic

### DIFF
--- a/data/sea/54-zebra.h
+++ b/data/sea/54-zebra.h
@@ -49,6 +49,7 @@ typedef struct zebra_state {
 
 } zebra_state_t;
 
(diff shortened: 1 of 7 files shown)


jystic:
👍 on 🍏 

anyway we could have caught this with a test?

tranma:
we'd need some testing around the haskell that sets things up for C. There is currently nothing that accomplishes this except a handful of CLI tests

Issue #508, opened 2017-03-10
Source evaluator is ignoring temporality?


amosr:
Source evaluator is rubbish. I thought we changed it to use temporality, but perhaps not.
This failure seems to be because it's evaluating it as an element instead of to an array

```
=== prop_progress_no_values_QWF from test/Icicle/Test/Source/Progress.hs:51 ===
*** Failed! Falsifiable (after 563 tests):
QueryWithFeature:
  Query: sort (latest 4 ~> animalnow)
  Now: animalnow
  Output: fozzy
  Type: String
  Name: animalfeaturename
  Time: animaltimename
  Key: `2038-08-02`
feature animalfeaturename
~> sort (latest 4 ~> animalnow)
Left (EvalErrorPrimBadArgs (Annot {annAnnot = (), annResult = Temporality TemporalityAggregate (ArrayT TimeT), annConstraints = []}) (Fun (BuiltinArray ArraySort)) [VTime (Time (2000 1 1 0 0 0s))])
```

PR #526, opened 2017-05-17
PSV sparse input: drop fact if any attributes have > fact_limit

amosr:
PSV sparse input was not dropping properly because it keeps a separate fact count for each attribute. when it started dropping on attribute 1, then the next buffer starts with attribute 2, it was running over the attribute 2, but never cleaning up afterwards.

the PSV sparse input code goes from:
```
        fleet->icount_0 ++;
        if (fleet->icount_0 > facts_limit)
```
to
```
        fleet->icount_0 ++;
        if (fleet->icount_0 > facts_limit ||  fleet->icount_1 > facts_limit)
```

dense doesn't change.

I tried generating the C locally and running it on redacted and it's didn't segfault

! @tranma @jystic 
/jury approved @jystic

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/IO/Psv/Input.hs
+++ b/icicle-compiler/src/Icicle/Sea/IO/Psv/Input.hs
@@ -99,13 +99,18 @@ seaOfReadTombstone input = \case
 
 seaOfReadNamedFact :: SeaInputError
                    -> InputAllowDupTime
+                   -> [SeaProgramAttribute]
                    -> SeaProgramAttribute
                    -> CStmt
-seaOfReadNamedFact errs allowDupTime state
+seaOfReadNamedFact errs allowDupTime all_states state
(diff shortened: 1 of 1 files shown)


tranma:
Just making a comment here so we don't forget:
It's not worth performing a check before cleaning up (i.e. `if not_dropping attr_0 then save_resumable attr_0`....) because we will be getting rid of psv soon.

👍  on 📗  from me

jystic:
This doesn't affect any of the generated code we have committed in tests?

jystic:
👍 was just curious that none of the `diff` tests failed because of this

tranma:
@jystic nope, our "tests" have one attribute :D

jystic:
Any idea why that test failed?

amosr:
yeah, it looks like a bug in the source evaluator: https://github.com/ambiata/icicle/issues/508

PR #527, opened 2017-05-17
Tests with separate compilation + Zebra test

amosr:
This catches the (now-fixed) attribute_count bug


jystic:
💯 tests 💃

tranma:
👍 thanks

PR #530, opened 2017-05-19
Source evaluator: `latest` returns array sometimes

amosr:
The `latest` returns an array if the input is an element, or a single value if it's an aggregate. The evaluator was not making the check as the type checker, so some cases where type said array but evaluator said single value. They should agree now, because the evaluator is using the result of typechecking to decide.

! @tranma @jystic 
/jury approved @jystic


jystic:
👍

tranma:
sorry, didn't see this. This looks good.
👍

PR #531, opened 2017-05-24
Outputs should be ordered alphabetically within each attribute

jystic:
I think this is the simplest fix, just ensure that outputs are always sorted by `OutputName` within an attribute.

I think ideally this should really be configurable, but this will do for the moment.

! @amosr @tranma
/jury approved @amosr

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/FromAvalanche/State.hs
+++ b/icicle-compiler/src/Icicle/Sea/FromAvalanche/State.hs
@@ -38,6 +38,7 @@ module Icicle.Sea.FromAvalanche.State (
   ) where
 
 import qualified Data.List          as List
+import           Data.Map (Map)
 import qualified Data.Map           as Map
 import qualified Data.Text          as T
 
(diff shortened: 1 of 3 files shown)


icicle-compiler/src/Icicle/Sea/FromAvalanche/State.hs
```diff
@@ -102,6 +104,7 @@ stateOfPrograms name attrib programs@(program :| _)
         , stateInputType      = factType
         , stateInputVars      = fmap (first textOfName) factVars
         , stateComputes       = NonEmpty.zipWith (stateOfProgramCompute name) (0 :| [1..]) programs
+        , stateOutputsAll     = Map.fromList . concatMap outputsOfProgram $ NonEmpty.toList programs
```
jystic: I'm not sure if maybe I should do this below in `Psv/Output.hs` instead of putting something else in the data structure

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
```diff
@@ -143,7 +143,7 @@ seaOfWriteProgramOutput config state = do
 
   let computes = NonEmpty.toList $ stateComputes state
   let resumeables  = fmap clearResumables computes
-  outputs         <- traverse outputState (concatMap stateOutputs computes)
+  outputs         <- traverse outputState . Map.toList $ stateOutputsAll state
```
tranma: 👍

tranma:
I'm happy but this fix but I'd wait until the tests generate WellTyped with more than one attribute first (I'm on that right now)
👍 otherwise

jystic:
> I'd wait until the tests generate WellTyped with more than one attribute first (I'm on that right now)

Yeah I just wanted to get this up here so I can test with production data

amosr:
👍

PR #533, opened 2017-05-25
Revert "Disable put_mutable_diff test"

jystic:
Making this a PR so we don't forget, I have no intention of merging this as whatever fixes the issue will probably re-enable the test.

! @amosr @tranma

### DIFF
--- a/icicle-compiler/test/Icicle/Test/Foreign/Array.hs
+++ b/icicle-compiler/test/Icicle/Test/Foreign/Array.hs
@@ -99,8 +99,8 @@ test_copy_eq input
 -- y = array_copy (x)
 -- z = array_put_mutable (y, i, a)
 -- array_eq (x, z) == false
--- FIXME flakey segfaulty test -- the values are always as expected but it might segfault
-x_FIXME_prop_array_put_mutable_diff
+-- TODO flakey segfaulty test -- the values are always as expected but it might segfault
+prop_array_put_mutable_diff
   = forAll arbitrary $ \input -> not (univalue (inputType input)) ==>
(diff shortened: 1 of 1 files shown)


tranma:
I don't see this test segfaulting anymore, but it does fail normally from time to time

jystic:
Yes, exactly - it's not ok to have a test which fails from time to time.

tranma:
yeah, I'm just pointing out that whatever reason is causing this test to fail is unrelated to the comment about the segfault. The issue in that comment has been addressed by this change here: https://github.com/ambiata/icicle/blob/master/icicle-compiler/test/Icicle/Test/Foreign/Utils.hs#L180

We haven't been able to reproduce this failure reliably -- will come back to it.

jystic:
Oh right, sorry, my bad (I didn't read the comment, just disabled the test)

amosr:
I've started fixing the test to use unboxed arrays, and if that doesn't stop the failures, it will at least give us more information to work off

tranma:
>I've started fixing the test to use unboxed arrays, and if that doesn't stop the failures, it will at least give us more information to work off

you think the failure is because the pointers happened to be the same after the `put`, because the arrays are "boxed"?

amosr:
> you think the failure is because the pointers happened to be the same after the put, because the arrays are "boxed"?

I'm actually still convinced that the values should be unique, since they're freshly allocated pointers, but something weird is definitely happening - maybe it's accidentally freeing some pointers early, so it's reusing?

but if it fails again, eg with Time, we will at least be able to look at the values that are failing - right now the test data that's printed has nothing to do with what it's actually running on.

jystic:
@amosr I guess this can be trashed now?

amosr:
kill it

PR #534, opened 2017-05-25
Make 'data._enum.tags' start at the right offset

jystic:
! @tranma @amosr 
/jury approved @amosr

### DIFF
--- a/data/sea/54-zebra.h
+++ b/data/sea/54-zebra.h
@@ -221,7 +221,7 @@ static int64_t zebra_translate_column
             int64_t offset = zebra_translate_column_tags
                 ( column_count
                 , dst
-                , data._enum.tags );
+                , data._enum.tags + elem_start );
             for (int64_t i = 0; i != column_count; ++i) {
                 offset += zebra_translate_column
                     ( mempool


jystic:
We really need the equivalent of the corpus tests for zebra. I had a quick look at `Icicle.Test.Sea.Zebra` and it looks like it would be easy to hook up the corpus, but it doesn't actually run the program or capture the psv output?

tranma:
I'm surprised prop_zebra didn't pick this up

jystic:
> I'm surprised prop_zebra didn't pick this up

Where is `prop_zebra` ?

tranma:
I mean `prop_read_entity`. It generates a bunch of facts, convert them to zebra, and read them back using the zebra input code here. It should have picked up the difference in the facts read back and the facts generated.

tranma:
I didn't think `corpus` was necessary for this code because this is all about input, whereas `corpus` would be checking the outputs of some queries.

jystic:
> I didn't think corpus was necessary for this code because this is all about input, whereas corpus would be checking the outputs of some queries.

Yeah I agree, I meant if we could extend those tests to also run the program and check the output then corpus would be valuable.

jystic:
> I mean prop_read_entity. It generates a bunch of facts, convert them to zebra, and read them back using the zebra input code here. It should have picked up the difference in the facts read back and the facts generated.

If can figure out a way to tweak the test so it picks up this error that would be great.

tranma:
yeah I'll see if I can tweak the generators. This is the right fix anyhow.
👍

amosr:
👍

tranma:
confirmed that this bug can be triggered by prop_read_entity by making sure there are many facts for any one entity, and that the number of facts for that entity is greater than the chunk size. This change fixed it.
I'm changing the WellTyped generator to be more diverse.

Issue #385, opened 2016-04-18
Sporadic test failures: NaN /= NaN


amosr:
I got this test failure:

```
[(OutputName "animal",VDouble NaN)] /= [(OutputName "animal",VDouble NaN)]
```

The issue is that NaN /= NaN, which is stupid. Fix this one day

Issue #536, opened 2017-05-25
NaN comparison


jystic:
Do we have some special equality function for this?
```
=== prop_check_history_core_same_aval from test/Icicle/Test/Source/History.hs:136 ===
*** Failed! Falsifiable (after 594 tests): 
TestStuff {tsQwf = QueryWithFeature:
  Query: sqrt (- 0.8817318515559287)
  Now: statlernow
  Output: fozzy
  Type: String
  Name: statlerfeaturename
  Time: statlertimename
[...]

PR #537, opened 2017-05-26
Add support for checking equality modulo NaN

jystic:
This adds a new type class `NanEq` which is the same as `Eq` but where `nanEq NaN NaN = True`.

I have changed the failing test https://github.com/ambiata/icicle/issues/536 to use the new `=~=` operator, but I am sure there will be other failures in time as I've added `DoubleT` back in to the allowed types which can be generated.

The plan would be to replace `===` with `=~=` whenever we come across a failure due to `NaN`.

This fixes https://github.com/ambiata/icicle/issues/385 and https://github.com/ambiata/icicle/issues/536.

! @amosr @tranma


amosr:
wow! 👍

tranma:
Javascripty
👍

icicle-source/src/Icicle/Encoding.hs
```diff
@@ -375,74 +376,73 @@ sourceTypeOfEncoding e
...
+renderJsonValue :: Value -> Maybe Text
+renderJsonValue = \case
+   Tombstone ->
+     Nothing
+   StringValue v ->
+     pure $ renderJsonString v
+   IntValue v ->
+     pure . T.pack $ show v
+   DoubleValue v ->
+     pure . T.decodeUtf8 $ renderDouble v
```
jystic: We do the JSON rendering ourself here instead of using Aeson, so that we can use `renderDouble` from Anemone

jystic:
On top of https://github.com/ambiata/anemone/pull/53

Issue #475, opened 2016-12-27
Links to Ivory are broken


k-bx:
Readme mentions https://github.com/ambiata/ivory few times, but that repository doesn't seem to exist.

PR #538, opened 2017-05-26
Clarify that Ivory is closed source in README

jystic:
Fixes https://github.com/ambiata/icicle/issues/475

! @amosr @tranma 
/jury approved @amosr


amosr:
:+1: not too much point since code deps (zebra/piano/snapper?) are closed source anyway

PR #543, opened 2017-05-31
New error constructor + Map_insert check maximum map size

amosr:
/jury approved @jystic

### DIFF
--- a/icicle-source/src/Icicle/Source/ToCore/Prim.hs
+++ b/icicle-source/src/Icicle/Source/ToCore/Prim.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 module Icicle.Source.ToCore.Prim (
     convertPrim
+  , primInsert
   ) where
 
 import                  Icicle.Source.Query
@@ -395,8 +396,9 @@ primInsert tk tv xm xk xv = do
              [ apps (bf $ Min.PrimBuiltinMap   $ Min.PrimBuiltinKeys tk tv)
              [ CE.XVar () n' ] ]
 
+  maxMapSize <- convertMaxMapSize
   let lenchk = apps (C.PrimMinimal $ Min.PrimRelation Min.PrimRelationLt T.IntT)
-             [ len, CE.XValue () T.IntT $ V.VInt 2 ]
+             [ len, CE.XVar () maxMapSize ]
 
   return     $ CE.makeLets () [ (n', insert) ]
              $ apps (C.PrimFold C.PrimFoldBool $ tsum)
--- a/icicle-source/src/Icicle/Source/ToCore/Prim.hs
+++ b/icicle-source/src/Icicle/Source/ToCore/Prim.hs
@@ -314,17 +314,16 @@ convertPrim p ann resT xts = go p
    = return $ CE.emptyMap tk tv
    | otherwise
    = convertError $ ConvertErrorCannotConvertType ann resT
+
   gomap MapInsert
    | ((xk, _) : (xv, _) : (xm, tm) : _) <- xts
    = case valTypeOfType tm of
        Just (T.MapT tk tv)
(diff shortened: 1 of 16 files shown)


tranma:
image
does not compute

icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs
```diff
@@ -275,7 +275,16 @@ makeApps a fun (arg:rest) doWrap
             -- Bare value. Note that this is now definite, but with same (bare) type
             bare  = Var (extractValueAnnot arga) nValue
 
-        fun' <- makeApps a (App a fun bare) rest True
+        -- Check if we're calling a primitive which already returns its result wrapped.
+        -- If so, we do not need to perform rewrapping of the value.
+        let doWrap'
+              | Just (p, _, _) <- takePrimApps fun
+              , primReturnsPossibly p
```
jystic: so none of the prims returned possibly before this? (just curious)

jystic:
👍 on 🇳🇬

icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs
amosr: "box" does, but it's desugared to a simple case expression in reifyPossibilityX: https://github.com/ambiata/icicle/blob/b3884e9f6fb84550aaad3c3d6c474d695ed12185/icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs#L47
It's harder to do the same here because then we'd kind of need two primitives, one that does the check and one that doesn't. Then we'd desugar the 'checked' one to the 'unsafe' one.

amosr:
I'm wondering whether this is the right way to do this, or whether the Core primitives should also return (`Sum Error (Map k v)`). It'd mean we could say the map size is valid for all Core programs as well, not just Source programs.

jystic:
> It'd mean we could say the map size is valid for all Core programs as well, not just Source programs.

That sounds like a nice property

amosr:
Ok, I'll think about how to modify Core to check map sizes. It's more work and requires changing flattening, but I think it'll be better in the end

amosr:
Changed my mind. I'll implement the drop check on the Source-Core conversion for now, since I'll need to handle the Possibly regardless, and once this works I can move the actual check to its own Core primitive

jystic:
💥

amosr:
The builds passed but this is still failing sometimes, but should be fixed by #550

amosr:
I think this failed because it's exposing an existing bug in `distinct`, but the next PR #562 changes distinct code gen anyway. So I'm going to squash them together..

amosr:
Test failure is not a bug in distinct, but because the typechecker changes to return `Possibly` for distincts. Distinct codegen was assuming that whole distinct could only be possibly if one of the subparts was possibly, so using `k'possibly || q'possibly` instead..

amosr:
closing PR anyway in favour of #562

icicle-source/src/Icicle/Source/ToCore/Prim.hs
```diff
@@ -365,3 +365,45 @@ convertPrim p ann resT xts = go p
...
+  maxMapSize <- convertMaxMapSize
+  let lenchk = apps (C.PrimMinimal $ Min.PrimRelation Min.PrimRelationLt T.IntT)
+             [ len, CE.XVar () maxMapSize ]
+
+  return     $ CE.makeLets () [ (n', insert) ]
+             $ apps (C.PrimFold C.PrimFoldBool $ tsum)
+             [ vright, verr, lenchk ]
+ where
+  apps f xs = CE.makeApps () (CE.XPrim () f) xs
+  bf = C.PrimMinimal . Min.PrimBuiltinFun
```
tranma: this checks out

PR #544, opened 2017-05-31
Add global max_map_size

amosr:
Making this available to all the evaluators + threading through the C code is a bit of a pain. This is a big change because it touches every part of the pipeline, but it's really only adding a new value.

! @jystic @tranma 
/jury approved @jystic


jystic:
👍

💯 for splitting this out to a separate PR, made it really easy to review

amosr:
The failure is a Core-Avalanche conversion error, if you don't mind I'll spend an hour or two trying to fix that for good. Hopefully it's just a matter of freshening names

amosr:
To clarify: this does not change semantics at all

PR #545, opened 2017-05-31
Test that maps are below max size

amosr:
This should be failing consistently but isn't. The generated programs aren't interesting enough.
/jury approved @jystic

### DIFF
--- a/icicle-compiler/test/test.hs
+++ b/icicle-compiler/test/test.hs
@@ -20,9 +20,6 @@ import qualified Icicle.Test.Avalanche.Flatten
 import qualified Icicle.Test.Avalanche.Melt
 import qualified Icicle.Test.Avalanche.MeltPrim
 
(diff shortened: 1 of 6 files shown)


jystic:
would one of the corpus programs fail consistently?

amosr:
Yeah, good idea. I think it's worth having both. I'm going to leave this PR here until I've got the rest implemented

amosr:
includes corpus tests for max map size

! @jystic @tranma

jystic:
👍

icicle-compiler/test/test.hs
```diff
@@ -80,13 +77,12 @@ main
 
         , Icicle.Test.Foreign.Array.tests
 
-        -- , Icicle.Test.Dictionary.Types.tests
-
```
jystic: 🍾

Issue #528, opened 2017-05-19
prop_check_commute failed


tranma:
```
=== prop_check_commutes from test/Icicle/Test/Avalanche/CheckCommutes.hs:27 ===
*** Failed! Falsifiable (after 40 tests): 
StructT (StructType (fromList [(StructField "black",BoolT),(StructField "grey",IntT)]))
[...]

PR #546, opened 2017-06-01
Substitution should not shadow

amosr:
Substitution was introducing shadowing, which is actually a type error. 
#528 


jystic:
👍

tranma:
Right, we don't handle shadowing at all. I forgot that.

tranma:
👍 should have realised, sorry! :)

PR #547, opened 2017-06-01
Simplify group conversion in Source-to-Core

amosr:
The test changes are just renaming.

Conversion to Core had the group and distinct code more or less copied across `convertFold` and `convertQuery`, but we can just call the `convertFold` case from `convertQuery`.

! @jystic @tranma 
/jury approved @jystic


jystic:
Queue @charleso

jystic:
👍

tranma:
haha yeah, it was easier to figure out the conversion manually but afterwards it's obvious that it's the same as `convertFold`
👍

PR #548, opened 2017-06-01
Prelude: missing day_of_week

tranma:
! @jystic 
/jury approved @jystic

### DIFF
--- a/data/libs/prelude.icicle
+++ b/data/libs/prelude.icicle
@@ -23,6 +23,9 @@ is_weekend v
 week_of_day d
   = trunc (double (days d) / 7).
 
+day_of_week v =
+  = mod (days v - 4) 7
+
 # Boolean statements
 not a = case a | True -> False | False -> True end.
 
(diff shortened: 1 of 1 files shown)


data/libs/prelude.icicle
```diff
@@ -37,7 +40,7 @@ snd x = case x | (a,b) -> b end.
 group_keys k = keys (group k ~> 0).
 
 # This only works if inlining is set to with-subst (:set inline with-subst in the REPL)
-group_vals k v = vals (group k ~> v).
+# group_vals k v = vals (group k ~> v).
```
jystic: why is this removed?

tranma: see the comment, it's not always safe to use

jystic: isn't that the default now? or not?

amosr: I don't think so, I thought we decided to wait until the switch to Sorbet to enable that

jystic: ah righto, perhaps we should remove it altogether then, I'm not sure what we gain by having it in there commented out?

jystic:
👍 happy either way tbh

PR #549, opened 2017-06-02
Test with multiple attributes and entities

tranma:
This reworked the WellTyped tests to be more honest, generating input much closer to what we might expect -- with multiple entities and attributes.

! @jystic @amosr 
/jury approved @amosr @tranma


icicle-compiler/src/Icicle/Sea/IO/Zebra.hs
```diff
@@ -174,7 +174,6 @@ seaOfDefReadProgram state = vsep
   , "        /* run compute on the facts read so far */"
   , "        if (*fact_count != 0) {"
   , indent 12 $ vsep $ fmap (\i -> pretty (nameOfCompute i) <+> " (program);") computes
-  , "            *fact_count = 0;"
```
jystic: Curious why no longer gets reset to 0? I introduced that [here](https://github.com/ambiata/icicle/pull/535/files) to fix a bug is all.

jystic:
This looks fantastic, just curious about that `*fact_count` change

icicle-compiler/src/Icicle/Sea/IO/Zebra.hs
tranma: right, the original condition should really be an invariant that I should prove for the whole program, but we all know that's not gonna happen :)

the `fact_count` is the number of facts read for each iteration of `zebra_read_step`, so it should be cleared there rather than right after compute, as compute is still in the middle of this iteration.

I removed this because I check `fact_count` in the test to find out how many rows I should read back from the icicle input struct.

amosr: but if fact_count is nonzero, isn't write_outputs going to rerun over the same input data twice?

amosr: other option is to modify psv_write_outputs to not call compute - but that sounds like it would require some scary changes to psv read

tranma: > but if fact_count is nonzero, isn't write_outputs going to rerun over the same input data twice?
no, because `write_outputs` is outside of `zebra_read_step` (it's in `zebra_snapshot_step`)

> other option is to modify psv_write_outputs to not call compute - but that sounds like it would require some scary changes to psv read

NOPE NOPE NOPE 🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅🙅 NOPE NOPE NOPE

tranma: so what I mean is to move this `fact_count = 0` to outside the loop in `zebra_read_step`

amosr: oh, right. sounds good. I misread your original comment

tranma:
I think this is finally ready to roll (the `icicle-compiler` test suites are the ones of interest). I will put some notes on any interesting parts and rebase

icicle-compiler/src/Icicle/Sea/FromAvalanche/Type.hs
```diff
@@ -62,28 +69,55 @@ defDepth t
...
+    ts' =
+      Set.fromList .
+      concatMap fixup .
+      Set.toList $
+        ts
+
+  in
+    if ts' == ts
+    then return ts
+    else progress ts'
```
tranma: ❓

icicle-compiler/test/Icicle/Test/Arbitrary/Corpus.hs
```diff
@@ -50,94 +49,154 @@ inputs =
...
+    , [outputid|tombstone:filter_group|]
+    , "feature injury ~> filter tombstone ~> group location ~> count location ")
+
+  , ( [inputname|injury|]
+    , [outputid|tombstone:filter_group_group|]
+    , "feature injury ~> filter tombstone ~> group location ~> group location ~> count location ")
+
+  , ( [inputname|injury|]
+    , [outputid|tombstone:filter_group_group_group|]
+    , "feature injury ~> filter tombstone ~> group location ~> group location ~> group location ~> count location ")
```
tranma: ❓ 
bunch of sanity checking queries, for checking treatment of tombstones

icicle-compiler/test/Icicle/Test/Arbitrary/Program.hs
```diff
@@ -4,201 +4,565 @@
...
+    noOutputs =
+      flip List.filter clusterEntities $ \e ->
+        not (e `elem` dropEntities) &&
+        not (e `elem` Map.keys hasOutputs)
+
+    fillers =
+      Map.fromList . fmap (, defaultOuputsForCluster) $ noOutputs
+
+  in
+    hasOutputs <> fillers
```
tranma: ❓ 
this is the bridge between Core evaluation and C evaluation.

They are the same except some logic is embedded into the input/output code (for both PSV and Zebra) that changes what facts go through C eval. This should cover all of them so that Core eval outputs the same things as C eval.

icicle-compiler/test/Icicle/Test/IO/IO.hs
```diff
@@ -0,0 +1,91 @@
...
+getAttr :: DictionaryEntry -> Attribute
+getAttr (DictionaryEntry a _ _) = a
+
+sortD :: Dictionary -> Dictionary
+sortD (Dictionary es fs) =
+  Dictionary (sortOn getAttr es) fs
+
+return []
+tests :: IO Bool
+tests = $checkAllWith TestRunFewer checkArgs
```
tranma: hold on this module shouldn't be in this change (botched merge)

jystic:
Have checked up to this point, all looks pretty good

icicle-compiler/src/Icicle/Sea/FromAvalanche/Type.hs
tranma: @amosr

amosr: neat

icicle-compiler/src/Icicle/Sea/IO/Psv/Input.hs
```diff
@@ -586,9 +586,7 @@ seaOfReadJsonList vtype avars = do
     [ "if (*p++ != '[')"
     , "    return ierror_loc_format (p-1, p-1, \"array missing '['\");"
     , ""
-    , "char term = *p;"
-    , ""
-    , "for (iint_t ix = 0; term != ']'; ix++) {"
+    , "for (iint_t ix = 0, term = *p; term != ']'; ix++) {"
```
amosr: this doesn't actually change anything, does it? just curious how it came about

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
```diff
@@ -411,6 +415,10 @@ seaOfOutput isJSON struct structIndex outId missing env outType argTypes transfo
    seaOfOutputBase' b
      = seaOfOutputBase b mismatch
 
+seaOfNotAnError :: Doc -> Doc
+seaOfNotAnError v =
+  "((" <> v <> " == ierror_not_an_error) || (" <> v <> " == ierror_tombstone))"
```
amosr: doesn't this mean it will print the default/uninitialised values for tombstones?

amosr: confused. maybe this should just be true. shouldn't we print NA for all errors (cannot compute, fold1 no value, tombstone, etc)?

icicle-compiler/test/Icicle/Test/Arbitrary/Corpus.hs
amosr: umm, is it worth grouping by different keys too?

icicle-compiler/test/Icicle/Test/Arbitrary/Program.hs
amosr: looks good, not sure about dropping though - it only happens in PSV, not Zebra, so how do you use this in the Zebra tests?

icicle-compiler/test/Icicle/Test/IO/IO.hs
amosr: 🔥

icicle-compiler/test/Icicle/Test/Sea/Psv.hs
```diff
@@ -47,141 +45,118 @@ import           P
...
+  -- there is no point in trying again
+  validated tryCountOutputType $
+    tryGenWellTypedWithInputAndOutput d x y
+
+genWTA1 x =
+  validated tryCountInputType $
+    tryGenAttributeWithInput x
+
+prop_success_array_of_struct_input
+ | dup <- AllowDupTime
```
amosr: interesting way to encode a let binding

icicle-compiler/test/Icicle/Test/Sea/Psv.hs
```diff
@@ -285,190 +276,102 @@ compileTest wt (TestOpts _ _ inputFormat allowDupTime) = do
...
-                  (maybe Map.empty (Map.singleton n) missingValue)
-                  n
-denseDictionary _ _ = return Nothing
-
-genMissingValue :: Gen (Maybe Text)
-genMissingValue = elements [Nothing, Just "NA", Just ""]
-
 ------------------------------------------------------------------------
 
 return []
```
amosr: 💯

amosr:
👍 tests good, not sure about the PSV output `SumT ErrorT` / `== ierror_tombstone` change

icicle-compiler/test/Icicle/Test/Arbitrary/Program.hs
tranma: the facts limit is just made absurdly huge for zebra

icicle-compiler/src/Icicle/Sea/IO/Psv/Input.hs
tranma: potential shadowing

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
tranma: hm, so I made this change because I wanted this behaviour:

input:
```
homer|test|tombstone|2000-1-1
```

core:
```
test_query
STREAM test
...
KONS = False
```

output:
```
homer|test_query|False
```

icicle-compiler/test/Icicle/Test/Sea/Psv.hs
tranma: I took this habit from you!

icicle-compiler/src/Icicle/Sea/IO/Psv/Input.hs
amosr: ah, cool

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
amosr: hm ok, that sounds more like a bug in the evaluator/compute rather than output?

icicle-compiler/test/Icicle/Test/Arbitrary/Corpus.hs
tranma: just focusing on tombstones for these queries, there are groups on different keys below this

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
tranma: yes this was wrong. I reverted this and changed the "model" (the test)

tranma:
gonna push this through when all builds are green and fix any failures as they come up later

tranma:
:agile:

amosr:
sounds ok to me. maybe you could just temporarily limit the number of attributes to 1 to get it to pass?

tranma:
it can pass as-is most of the time now. I've only seen the unsafeSubstExp error sometimes.

icicle-compiler/test/Icicle/Test/Arbitrary/Program.hs
```diff
@@ -4,201 +4,594 @@
...
+
+-- FIXME
+--
+-- This only gnerates one compute kernel for each cluster.
+-- PsvFission tests multiple kernels by doing its own thing.
+-- Perhaps roll that into this as well.
+--
+data WellTypedCluster = WellTypedCluster {
+    wtCluster       :: Sea.Cluster () ()
+  , wtFactType      :: !ValType -- clusterInputType == SumT ErrorT wtFactType
```
jystic: Should this be `SumErrorFactT` ?

tranma: nah, it's the value type `t` as it would be coming from a TOML dictionary. A cluster input type is `SumT ErrorT t` and a core program stream type is `SumT ErrorT (t, TimeT`)` (or something like that)

tranma:
stack trace courtesy of ghc 8

```

=== prop_success_array_of_struct_input from test/Icicle/Test/Sea/Psv.hs:87 ===
*** Failed! (after 75 tests): 
Exception:
  IMPOSSIBLE: unsafeSubstExp name being shadowed, cannot perform substitution
  CallStack (from HasCallStack):
    error, called at src/Icicle/Core/Program/Subst.hs:77:20 in icicle-core-0.0.1-A3Zm7l0dm1cIZwtW4tMC0A:Icicle.Core.Program.Subst
ArrayT (StructT (StructType (fromList [(StructField "black",StringT),(StructField "blue",BoolT),(StructField "green",DoubleT),(StructField "orange",StringT),(StructField "purple",TimeT),(StructField "red",BoolT),(StructField "yellow",DoubleT)])))
Exception thrown while printing test case:
  IMPOSSIBLE: unsafeSubstExp name being shadowed, cannot perform substitution
  CallStack (from HasCallStack):
    error, called at src/Icicle/Core/Program/Subst.hs:77:20 in icicle-core-0.0.1-A3Zm7l0dm1cIZwtW4tMC0A:Icicle.Core.Program.Subst

Exception thrown while printing test case:
  IMPOSSIBLE: unsafeSubstExp name being shadowed, cannot perform substitution
  CallStack (from HasCallStack):
    error, called at src/Icicle/Core/Program/Subst.hs:77:20 in icicle-core-0.0.1-A3Zm7l0dm1cIZwtW4tMC0A:Icicle.Core.Program.Subst
```

jystic:
The `branches-7-10-compiler-psv` tests look like they take about 70 minutes to run, are there any which you think are redundant or could be split in to separate test suites?

It would be good if we could keep builds under 30 minutes. Ideally builds would less than 10 minutes, but I understand that's a bit tricky given how programs are currently generated.

tranma:
yeah going to split in the the next PR rn

jystic:
💯

PR #550, opened 2017-06-02
Avalanche type checker: require fresh names. Freshen after flattening.

amosr:
! @tranma @jystic


amosr:
This changes the typechecker to require fresh names

tranma:
this seems legit
👍

tranma:
but you might need to freshen after something else as well, not just flatten.

amosr:
Yeah, I'm worried about whether I'll be able to get them all, but on the other hand I think it's better to leave it as an error so we can actually find these places

amosr:
ergm

PR #552, opened 2017-06-02
Facts and features can have Ivory-style names

amosr:
Fact and feature names can be Ivory names, but need to be converted to a legal C identifier during Sea conversion.

! @jystic @amosr @HuwCampbell 
/jury approved @amosr


tranma:
👍

amosr:
I changed the name mangler but it broke everything

tranma:
actually doesn't work for some type names, e.g. `input_foo_t`
👎

tranma:
changing the sea name mangler isn't enough, we'll need to find and fix everywhere the virtual feature name is taken from the dictionary as well. ghc can probably help quite a bit if we change the text type for virtual/concrete feature names to be something special that allow dots.

tranma:
going to take over this

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -91,7 +91,7 @@ mangleToSeaName = SeaName . Text.pack . concatMap mangle . show . pretty
      | c >= 'A' && c <= 'Z'
      = [c]
      | c == '_'
-     = "__"
+     = "_"
```
amosr: it may not matter in practice, but the reason I had escaped the underscore here was to distinguish an identifier containing underscores from the underscore followed by hex: `fish_21` vs `fish_21` or `fish!`

tranma:
this is good to be reviewed now.
@jystic

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: ok, I guess it's also a good way to make sure all generated identifiers go through `mangle` (otherwise some tests will get mismatched names, e.g. `iint_t foo_bar; .... foo__bar = 0`)

tranma: although this will mean `mangle (mangle x) != mangle x`. What do you think?

tranma: Should we have to pick one, between:

`forall x. mangle (mangle x) == mangle x`

and:

`mangle x == mangle y ==> x == y`

?

HuwCampbell: Could replace `_` with `_95` (it's hex character code). Should get both then.

amosr: I'd prefer the second: it's more like escaping a string than finding a normal form.

tranma: ah yeah, let's just escape everything that's not alphanumeric

tranma: we still won't get `mangle (mangle x) == mangle x` but I suppose that's alright

amosr: Yeah actually that sounds easier.. unless there's something other than underscore we can use as an escape, which we know won't occur in the original string... `µµ`?

HuwCampbell: This is a bit contrived, but
````haskell
> showHex ( ord 'λ' ) ""
"3bb"
showHex ( ord ';' ) ""
"3b"
```
which might make
```haskell
mangle "λ" == mangle ";b"
```

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -35,25 +40,75 @@ import           Text.Printf (printf)
...
+  pretty =
+    string . Text.unpack . getSeaString
+
+takeSeaName :: SeaName -> Text
+takeSeaName = getSeaName
+
+seaNameValidHead :: Char -> Bool
+seaNameValidHead c =
+  c >= 'a' && c <= 'z'
+{-# INLINE seaNameValidHead #-}
```
jystic: are capital letters valid?

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
amosr: damn, you're right - maybe we should append "_" after the showHex too

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: nope

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: I'm just going through to make sure no name gets mangled inconsistently between its definition and mentions. It's a bit tedious but not too bad...

HuwCampbell: That's exactly what I was going to suggest

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -35,25 +40,75 @@ import           Text.Printf (printf)
...
+     | c >= '0' && c <= '9'
+     = [c]
+     | c >= 'a' && c <= 'z'
+     = [c]
+     | c >= 'A' && c <= 'Z'
+     = [c]
+     | c == '_'
+     = "_"
+     | otherwise
+     = "_" <> showHex (ord c) ""
```
jystic: I think this function needs a test, I suspect it is not generating unambiguous names.

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -35,25 +40,75 @@ import           Text.Printf (printf)
...
+    getSeaString :: Text
+ } deriving (Eq, Ord, Show)
+
+instance Pretty SeaName where
+  pretty =
+    string . Text.unpack . takeSeaName
+
+instance Pretty SeaString where
+  pretty =
+    string . Text.unpack . getSeaString
```
amosr: should this do `seaOfString` to escape and wrap in quotes..?

icicle-compiler/test/Icicle/Test/Arbitrary/NanEq.hs
```diff
@@ -117,7 +117,9 @@ instance NanEq Time where
 
 instance NanEq OutputName
 
-instance NanEq Namespace
+instance NanEq Namespace where
+  nanEq =
+    (==)
```
amosr: I'm just curious, did this need to be changed because the constructor is no longer exported?

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
jystic: i'm confused, that doesn't gel with the definition of `mangle` below, is that because these are manually introduced identifiers?

icicle-compiler/test/Icicle/Test/Arbitrary/NanEq.hs
amosr: oh - maybe because `NamespaceName` isn't deriving `Generic`

amosr:
I might be out of the loop, but I don't understand why the front-end needs to change for this. I thought it would just be changing the C pretty printer to mangle names consistently?

icicle-repl/test/cli/repl/t30.3-sum-not-error/expected
```diff
@@ -41,20 +41,20 @@ typedef struct {
...
+    iint_t           repl_24ix_241;
+    idouble_t        repl_24ix_242;
 
     /* resumables: values */
-    idouble_t        res_0_0_acc_perhaps_conv_5_simpflat_6;
-    iint_t           res_0_0_acc_perhaps_conv_5_simpflat_5;
-    ibool_t          res_0_0_acc_perhaps_conv_5_simpflat_4;
+    idouble_t        res_0_0_acc_24perhaps_24conv_245_24simpflat_246;
+    ibool_t          res_0_0_acc_24perhaps_24conv_245_24simpflat_244;
+    iint_t           res_0_0_acc_24perhaps_24conv_245_24simpflat_245;
```
jystic: it's a shame these become a bit harder to read, apparently `$` is actually accepted by `gcc` and `clang`, maybe we should pass it through? :trollface:

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: you are right. `mangle` doesn't deal with the `head` specially, it should.

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: no, that's the responsibility of the caller. SeaString is by definition, "a value that will need to be quoted in C"

tranma:
@amosr the front-end dictionaries need to change because we want to be consistent with blizzard, in that we explicitly allow Ivory-style names for facts and features only.

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma:
```
prop_mangle_name :: Property
prop_mangle_name =
  forAll genValidUtf8 $ \text1 ->
  forAll genValidUtf8 $ \text2 ->
  mangleToSeaName text1 == mangleToSeaName text2 ==>
  text1 == text2
```
^ most thorough but too many discards

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
jystic: but shouldn't capital letters be allowed then? it seems like they're not allowed in the head or the tail based on these functions

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
amosr: hm, maybe it shouldn't have a pretty instance then - if it has to be explicitly handled?

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: oh right, you can have capitals at the head of C identifiers.

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
jystic: I guess the discard problem would still be there, but maybe you could write a different generator which creates escape looking things a lot of the time:
```hs
genChar :: Jack Char
genChar =
  chr <$> chooseInt (ord minBound, ord maxBound)

genAwful :: Jack String
genAwful =
  fmap concat . listOf $ oneOf [
      (\x -> "_" <> showHex (ord x) "") <$> genChar
    , (:[]) <$> genChar
    , pure "_"
    ]
```

tranma: it's not too bad, there are lots of discards but  it still finishes in a few minutes.

icicle-compiler/test/Icicle/Test/Arbitrary/NanEq.hs
tranma: oh right

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
jystic: oh ok cool, I thought you were saying the test wasn't plausible because there were too many discards, all good

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: yeah good point. Done

tranma:
should `mangle "" == ""`?

jystic:
Yeah I think so

tranma:
right, I think so too, but then the signature `mangleToSeaName :: Pretty n => n -> SeaName` and claim that `SeaName` is a `legal C identifier` is a bit of a lie. I suppose that's not really a problem.

tranma:
this will likely pass 7-10-repl and fail 8-0-repl because the ordering of SeaNames in the HashMap changed... /more bikeshedding

jystic:
> this will likely pass 7-10-repl and fail 8-0-repl because the ordering of SeaNames in the HashMap changed... /more bikeshedding

I'm curious why this was not a problem before? Haven't we always pretty printed the names in map order?

tranma:
so it's just different on 7.10.3 (my version)

jystic:
I've been thinking, I suspect the issue is that we have a lock file for `7.10.2` and `8.0.1` which pins down the version of hashable. I bet if you made a `7.10.3` lock file with just the right version of hashable it would probably work.

eg:
```
# mafia-lock-file-version: 0
hashable == 1.2.4.0
```

tranma:
yeah good point....I should have thought of that!

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -35,25 +40,75 @@ import           Text.Printf (printf)
...
+takeSeaName :: SeaName -> Text
+takeSeaName = getSeaName
+
+takeSeaString :: SeaString -> Text
+takeSeaString = getSeaString
+
+seaNameValidHead :: Char -> Bool
+seaNameValidHead c =
+  (c >= 'a' && c <= 'z') ||
+  (c >= 'A' && c <= 'Z')
```
amosr: nit: I think '_' is valid too?

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -35,25 +40,75 @@ import           Text.Printf (printf)
...
+          | c >= '0' && c <= '9'
+          = [c]
+          | c >= 'a' && c <= 'z'
+          = [c]
+          | c >= 'A' && c <= 'Z'
+          = [c]
+          | c == '_'
+          = "__"
+          | otherwise
+          = "u_" <> showHex (ord c) ""
```
amosr: do we need '_' suffix as well, to signal the end of hex, in case next char is digit/hex?

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
erikd-ambiata: It is!

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
tranma: not necessary for valid names but will be easier to read, yeah

icicle-compiler/src/Icicle/Sea/FromAvalanche/Program.hs
```diff
@@ -188,9 +226,9 @@ seaOfStatement state compute stmt
      KeepFactInHistory _
       -> Pretty.empty
   where
-   stNew   n = "s->" <> stateInputNew (seaOfName n)
-   stRes   n = "s->" <> stateInputRes (nameOfResumable compute $ seaOfName n)
-   stHas   n = "s->" <> stateInputHas (nameOfResumable compute $ seaOfName n)
+   stNew   n = "s->" <> stateInputNew (prettyText . takeSeaName $ n)
```
amosr: why not define `seaOfName = prettyText . takeSeaName . mangleToSeaName` ?

tranma: I wanted to make it harder to accidentally mangle a c name multiple names (resulting in name mismatch between definition and uses). `takeSeaName . mangleToSeaName` hides this imo

amosr:
👍

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
jystic: curious why the `u` prefix? is that required if we have an '_' suffix?

PR #553, opened 2017-06-09
Update to new master runner with caching

jystic:
This is still experimental, but it's ready for a serious trial.

It enables mafia caching across build bots.

! @tranma

### DIFF
--- a/icicle-compiler/master.toml
+++ b/icicle-compiler/master.toml
@@ -1,11 +1,12 @@
 [master]
-  runner = "s3://ambiata-dispensary-v2/dist/master/master-haskell/linux/x86_64/20170104233312-ee8addd/master-haskell-20170104233312-ee8addd"
+  runner = "s3://ambiata-dispensary-v2/dist/master/master-haskell/linux/x86_64/20170609055811-189e18a/master-haskell-20170609055811-189e18a"
   version = 1
-  sha1 = "3b707a2ba2d74ee13fe81a658bed191a5808e6f5"
+  sha1 = "b11f877a5f5cfda3594edbb11b911ccb2d3ee90f"
 
 [build.dist-7-10]
   GHC_VERSION="7.10.2"
-  CABAL_VERSION="1.22.4.0"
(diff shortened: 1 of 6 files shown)


icicle-compiler/master.toml
```diff
@@ -1,11 +1,12 @@
...
-  runner = "s3://ambiata-dispensary-v2/dist/master/master-haskell/linux/x86_64/20170104233312-ee8addd/master-haskell-20170104233312-ee8addd"
+  runner = "s3://ambiata-dispensary-v2/dist/master/master-haskell/linux/x86_64/20170609055811-189e18a/master-haskell-20170609055811-189e18a"
   version = 1
-  sha1 = "3b707a2ba2d74ee13fe81a658bed191a5808e6f5"
+  sha1 = "b11f877a5f5cfda3594edbb11b911ccb2d3ee90f"
 
 [build.dist-7-10]
   GHC_VERSION="7.10.2"
-  CABAL_VERSION="1.22.4.0"
+  CABAL_VERSION="1.24.0.0"
```
tranma: is this cabal version change ok -- why does it need the change?

jystic: Sorry for the long speel, the reason is complicated :)

With cabal <= 1.24 it was generating ghc package keys (the hash when you `ghc-pkg dump <package-id>`) which seemed to differ depending on the box that built the binary for some packages, one example is text. It seems that this hash has been improved and doesn't change on the more recent versions of cabal, likely to help support `cabal new-build`. So in order to share binaries between boxes we need the new version of cabal.

Yesterday I updated the boris boxes so that 7.10 has the new version of cabal available, so is ok to choose either version on 7.10 builds now. You can't pick it for 7.8.4 though.

tranma: yes, you told me this once, I just completely forgot about it -- cool

tranma:
can't wait to get it rolled out
👍

jystic:
me too, I hope it actually works this time, I'm still a little cautious

PR #559, opened 2017-06-14
Use GHC z-encoding for C variable names

jystic:
Not sure that we should actually do this, but I had a crazy thought this morning and thought I would try it out.

GHC already [solves](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/SymbolNames) the problem of encoding arbitrary strings to C identifiers, so why not steal the same encoding?

One side benefit is that we'll get used to it and hence will be able to naturally read GHC linker symbols and the like. It's also neat that we get decoding for free. It's also cool that the symbol names a mnemonics, like `$` is encoded as `zd`. The `d` makes it easy to guess that the symbol is a dollar.

Not saying we should definitely do this, just wanted to see what people though?

! @tranma @amosr 
/jury approved @amosr @tranma

### DIFF
--- a/icicle-compiler/icicle.cabal
+++ b/icicle-compiler/icicle.cabal
@@ -64,6 +64,7 @@ library
                      , vector                          == 0.11.*
                      , vector-space                    == 0.10.*
                      , void                            >= 0.5        && < 0.8
+                     , zenc                            == 0.1.*
 
                      -- Used for benchmark data generation
                      , QuickCheck                      == 2.8.*
--- a/icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
+++ b/icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
@@ -11,6 +11,7 @@ module Icicle.Sea.FromAvalanche.Base (
   , asSeaName
   , mangleToSeaName
   , mangleToSeaNameIx
+  , unmangleSeaName
   , attributeAsSeaString
   , seaOfChar
   , seaOfString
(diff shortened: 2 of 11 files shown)


icicle-compiler/test/Icicle/Test/Sea/Name.hs
```diff
@@ -27,14 +29,32 @@ prop_mangle_works =
     isNothing (asSeaName text1) ==>
     isJust . asSeaName . takeSeaName . mangleToSeaName $ text1
 
-prop_mangle_unique :: Property
-prop_mangle_unique =
-  forAll genValidUtf8 $ \text1 ->
-  forAll genValidUtf8 $ \text2 ->
-  mangleToSeaName text1 == mangleToSeaName text2 ==>
-  text1 == text2
```
jystic: This test wasn't actually doing anything, I put a `trace` / `classify` in and found that the only successful test cases were the empty string.

icicle-compiler/src/Icicle/Sea/FromAvalanche/Base.hs
```diff
@@ -11,6 +11,7 @@ module Icicle.Sea.FromAvalanche.Base (
   , asSeaName
   , mangleToSeaName
   , mangleToSeaNameIx
+  , unmangeSeaName
```
amosr: unmange

amosr:
👍

icicle-compiler/icicle.cabal
```diff
@@ -64,6 +64,7 @@ library
                      , vector                          == 0.11.*
                      , vector-space                    == 0.10.*
                      , void                            >= 0.5        && < 0.8
+                     , zenc                            == 0.1.*
```
tranma: how did I not find out about this package sooner???

tranma:
💯

PR #561, opened 2017-06-15
Write the names / ordering of facts to the C dictionary

jystic:
Sorry for the two PRs in one, this adds a check that the attribute count matches what we expect and also dumps the list of attribute names and the order we expect them in as part of the C.

! @amosr @tranma 
/jury approved @amosr @tranma


amosr:
👍

tranma:
👍

Issue #555, opened 2017-06-14
Sorting mutates in-place


amosr:
Input dictionary:

```
[feature.newest]
   expression = "feature nested ~> newest value"
[feature.sorted]
   expression = "feature nested ~> sort (newest value)"
```

Input data:
```
[3,1,2]
```

Output:
```
newest: [1,2,3]
sorted: [1,2,3]
```

sorted is sorting the input array in-place

PR #563, opened 2017-06-15
Sorty

amosr:
#555 
/jury approved @jystic


amosr:
Fix sorting in-place on input arrays
! @jystic @tranma

jystic:
👍

PR #566, opened 2017-06-19
Set default function application to substitution

amosr:
/jury approved @jystic


amosr:
This changes how function application works, but the only function in the prelude it will change is `group_vals`, which is unlikely to be used since the old way is broken for it

jystic:
👍 I'm good with this, do you think we need to do any validation on real data sets?

amosr:
validated this on some real data and it's ok, but would like to check a few other dictionaries to make sure nothing is using the `group_vals` function.

PR #567, opened 2017-06-19
Write out schema whenever we run a query

jystic:
This change includes a formation/walrus schema in the compiled C dictionary. Which means we have access to it when running a query. As such, we can now write out a schema alongside the data 🎆 

! @tranma @amosr

/cc @nhibberd 
/jury approved @amosr

### DIFF
--- a/icicle-compiler/main/icicle.hs
+++ b/icicle-compiler/main/icicle.hs
@@ -78,7 +78,6 @@ pQuery =
     <*> pInputFile
     <*> pOutputFile
     <*> optional pOutputCode
-    <*> optional pOutputSchema
     <*> (pSnapshot <|> pChordPath)
     <*> pLimit
     <*> pDrop
@@ -149,7 +148,7 @@ pOutputFormat =
 
 pOutputSparse :: Parser OutputFile
 pOutputSparse =
-  fmap (OutputFile OutputSparsePsv) $
+  fmap (\path -> OutputFile OutputSparsePsv path Nothing) $
     strOption (long "output-sparse-psv" <> metavar "OUTPUT_PSV")
 
 pOutputSparseFormat :: Parser OutputFormat
@@ -158,8 +157,9 @@ pOutputSparseFormat =
 
 pOutputDense :: Parser OutputFile
 pOutputDense =
-  fmap (OutputFile OutputDensePsv) $
-    strOption (long "output-dense-psv" <> metavar "OUTPUT_PSV")
+  OutputFile OutputDensePsv
+    <$> strOption (long "output-dense-psv" <> metavar "OUTPUT_PSV")
+    <*> optional pOutputSchema
 
 pOutputDenseFormat :: Parser OutputFormat
 pOutputDenseFormat =
--- a/icicle-compiler/main/icicle.hs
+++ b/icicle-compiler/main/icicle.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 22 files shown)


icicle-compiler/main/icicle.hs
```diff
@@ -269,7 +279,19 @@ runQuery f msrc query = do
...
-  liftIO (maybe (return ()) (flip writeUtf8 (querySource query)) msrc)
+  case msrc of
+    Nothing ->
+      pure ()
+    Just src ->
+      writeUtf8 src (querySource query)
+
+  case sfOutputSchema (queryFleet query) of
+    Nothing ->
+      pure ()
```
amosr: when will this be `Nothing` - should it be an error? if you specify `--output-schema blah` and there's no schema, maybe it should warn

jystic: It will be `Nothing` when we're doing sparse output.

We always output a schema if we're doing dense, and never for sparse. The `--output-schema` option is really just to override the default path that the schema is written to. I can see what you're saying, just not sure how to surface it.

tranma: true, should be impossible to accidentally ignore the schema

jystic: Maybe we can do the check in the cli parsing instead of here? Could say that `--output-schema` isn't available if the user says `--output-sparse`?

icicle-compiler/src/Icicle/Sea/Fleet.hs
```diff
@@ -202,6 +203,25 @@ initSnapshot library piano = \case
   HasInput (FormatZebra _ _ _) _ input ->
     initZebraSnapshot library piano input
 
+getOutputSchema :: Library -> Input a -> EitherT SeaError IO (Maybe PsvSchema)
+getOutputSchema library = \case
+  NoInput ->
```
amosr: what is `NoInput` for? just curious

icicle-compiler/src/Icicle/Sea/IO/Psv/Output.hs
```diff
@@ -555,3 +598,95 @@ outputDie = "if (error) return error;"
...
+      columns0 <- concat <$> traverse schemaOfProgram programs
+
+      let
+        columns =
+          columns0 <> schemaOfLabel (outputPsvMode config)
+
+        missing =
+          PsvMissingValue $ outputPsvMissing config
+
+      pure $ PsvSchema missing columns
```
tranma: 👍

icicle-compiler/src/Icicle/Sea/Fleet.hs
jystic: This is the code path for when we do eval in the repl

tranma:
fleet --> schema --> json looks good
👍

icicle-compiler/test/cli/icicle/run
```diff
@@ -1,5 +1,16 @@
...
+: ${UPDATE:="0"}
+
+diff_update () {
+    if [ "$UPDATE" = "1" ]; then
+        echo Updating $1
+        cat $2 > $1
+    else
+        diff -u $1 $2
+    fi
+}
```
amosr: 💯

icicle-compiler/main/icicle.hs
jystic: Have changed the cli so it looks like this:
```
(--output-sparse-psv OUTPUT_PSV | --output-dense-psv OUTPUT_PSV [--output-schema OUTPUT_PSV_SCHEMA_JSON])
```

amosr: Ok, cool

amosr:
👍

PR #568, opened 2017-06-21
Launch the missiles

jystic:
! @tranma @amosr 
/jury approved @amosr

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/Eval/Base.hs
+++ b/icicle-compiler/src/Icicle/Sea/Eval/Base.hs
@@ -252,11 +252,21 @@ getCompilerOptions = do
     Just env -> return $ defaultCompilerOptions <> T.splitOn " " (T.pack env)
 
 defaultCompilerOptions :: [CompilerOption]
-defaultCompilerOptions =
-  [ "-O3"           -- 🔨
-  , "-g3"           -- include debug information (3 = as much as possible)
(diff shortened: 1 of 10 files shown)


icicle-compiler/src/Icicle/Sea/Eval/Base.hs
```diff
@@ -252,11 +252,21 @@ getCompilerOptions = do
...
+  , "-g3"
+
+  -- all optimisations valid for the current CPU (AVX512, etc)
+  , "-march=native"
+
+  -- variable declarations anywhere!
+  , "-std=c99"
+
+  -- can't segfault if we exit(1) first!
+  , "-DICICLE_ASSERT=1"
```
amosr: our segfault handler also prints out a stacktrace. would it be worthwhile doing this in the assertion code too?

amosr:
👍

amosr:
Test failure: I think the Zebra test function needs killing
```
int64_t test_check_limit (zebra_state_t *state, zebra_entity_t *entity) {
    if (zebra_limit_exceeded (state)) {
        zebra_write_dropped_entity (state, entity);
        return 1;
    }
    return 0;
}
```

```
error: read entity: In function 'test_zebra_read_entity':
read entity:46: error: implicit declaration of function 'zebra_limit_exceeded'
read entity: In function 'test_check_limit':
read entity:54: error: implicit declaration of function 'zebra_write_dropped_entity'
```

PR #569, opened 2017-06-21
Haskell interface for creating Icicle runtime arrays

jystic:
This is the start of doing the Zebra -> Icicle runtime conversion in Haskell.

Thought it would be useful to have a decent interface to Icicle's runtime array representation.

The Haskell interface is untyped, so it's equivalent to `iarray_iany_t`. The safe interface checks that the `Storable` you are writing is indeed 8-bytes wide.

! @amosr @tranma 
/jury approved @amosr

### DIFF
--- a/icicle-compiler/src/Icicle/Runtime/Array.hs
+++ b/icicle-compiler/src/Icicle/Runtime/Array.hs
@@ -1,0 +1,362 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DoAndIfThenElse #-}
+{-# LANGUAGE ForeignFunctionInterface #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+module Icicle.Runtime.Array (
+    Array(..)
+  , ArrayIndex(..)
+  , ArrayLength(..)
+  , ArrayCapacity(..)
+  , ArrayFootprint(..)
+
+  , new
+  , length
+  , read
+  , write
+  , grow
+  , fromVector
+  , fromList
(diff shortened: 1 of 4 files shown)


icicle-compiler/src/Icicle/Runtime/Array.hs
```diff
@@ -0,0 +1,352 @@
...
+{-# INLINABLE new #-}
+
+length :: Array -> IO ArrayLength
+length =
+  fmap ArrayLength . peek . unsafeLengthPtr
+{-# INLINE length #-}
+
+checkElementSize :: Storable a => a -> EitherT ArrayError IO ()
+checkElementSize x = do
+ if sizeOf x /= 8 then
```
amosr: this could be `/= elementSize`

icicle-compiler/src/Icicle/Runtime/Array.hs
```diff
@@ -0,0 +1,352 @@
...
+calculateCapacity (ArrayLength len) =
+  if len < 4 then
+    4
+  else
+    let
+      !bits =
+        64 - countLeadingZeros (len - 1)
+    in
+      ArrayCapacity (1 `shiftL` bits)
+{-# INLINE calculateCapacity #-}
```
amosr: maybe
```
forall n >= 2. calculateCapacity (2^n) == 2^n
```

amosr:
👍 
maybe `icicle-compiler/test/Icicle/Test/Foreign/Utils.hs:pokeArr` could be updated to use this, not sure how easy that would be though

icicle-compiler/src/Icicle/Runtime/Array.hs
jystic: ah!

icicle-compiler/src/Icicle/Runtime/Array.hs
jystic: done

tranma:
does it still make sense to have those strings-as-C-code tests in `Foreign.Array` or would it be better to convert them to using this interface?

jystic:
> does it still make sense to have those strings-as-C-code tests in Foreign.Array or would it be better to convert them to using this interface?

Yeah definitely, probably better to wait until I have a full replacement for the foreign test utils though.

PR #570, opened 2017-06-22
Split branches builds in to separate test suites

jystic:
Should speed up builds a bit, running tests in parallel.

! @amosr @tranma


tranma:
yes please
💯

jystic:
Not sure if these are the right splits, if we notice one build taking longer than others would be good to split again

jystic:
Worth also noting that I didn't bother splitting the master build, that one runs all the tests before publishing a new binary

icicle-compiler/test/test.hs
```diff
@@ -41,52 +38,104 @@ import qualified Icicle.Test.Sea.Text
...
+      Icicle.Test.Sea.Psv.tests
+    , Icicle.Test.Sea.Psv.Schema.tests
+    , Icicle.Test.Sea.PsvFission.tests
+    ]
+
+zebra :: TestSuite
+zebra =
+  TestSuite "zebra" [
+      Icicle.Test.Sea.Zebra.tests
+    , Icicle.Test.Sea.Seaworthy.tests
```
tranma: this is psv

icicle-compiler/test/test.hs
```diff
@@ -41,52 +38,104 @@ import qualified Icicle.Test.Sea.Text
...
+    , Icicle.Test.Sea.Name.tests
```
tranma: this is not zebra specific

PR #571, opened 2017-06-22
Data types for zebra <-> icicle runtime translation

jystic:
These data types are designed to be as close as we can get to the Icicle runtime representation.

They are designed to be easy to translate to/from zebra as well as to icicle runtime.

`Time64` for example will be the packed representation we use in C for example.

! @amosr @tranma

### DIFF
--- a/icicle-compiler/src/Icicle/Runtime/Logical.hs
+++ b/icicle-compiler/src/Icicle/Runtime/Logical.hs
@@ -1,0 +1,42 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_GHC -funbox-strict-fields #-}
+module Icicle.Runtime.Logical (
+    Value(..)
+  ) where
+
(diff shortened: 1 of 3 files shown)


tranma:
ok, I think this makes sense because it's not so much an encoding as a description of the memory layout.
👍

icicle-compiler/src/Icicle/Runtime/Logical.hs
```diff
@@ -0,0 +1,42 @@
...
+  | Time !Time64
+
+  | Left !Value
+  | Right !Value
+
+  | None
+  | Some !Value
+
+  | Error !Error64
+  | Success !Value
```
amosr: if `Success` means `Right :: a -> Sum Error a`, you probably want another constructor to distinguish between `Left Error` and raw, unadorned `Error`. unadorned errors are unlikely to occur, but I imagine we can encode them?

jystic: Ah I see, maybe it's not worth having this specialised?

jystic: Talked IRL, we're going to try without allowing unadorned `Error` and see what happens

PR #572, opened 2017-06-22
Reify possiblies: deal with primitives that return possibly

amosr:
! @tranma @jystic 
/jury approved @nhibberd

### DIFF
--- a/icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs
+++ b/icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs
@@ -284,7 +284,7 @@ makeApps a fun (arg:rest) doWrap
               | otherwise
               = True
 
-        fun' <- makeApps a (App a fun bare) rest doWrap'
+        fun' <- makeApps a (App annotWrapPrim fun bare) rest doWrap'
 
         let app'  = Case a' arg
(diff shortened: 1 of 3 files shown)


tranma:
context: before the "cannot compute" change, only the arguments in an `App` can "infect" the type of the `App` with `Possibly`. Now the function can do that too.
👍

icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs
```diff
@@ -296,7 +296,15 @@ makeApps a fun (arg:rest) doWrap
 
  -- If argument is a definitely, just apply it as usual
  | otherwise
- =  makeApps a (App a fun arg) rest doWrap
+ =  makeApps a (App annotWrapPrim fun arg) rest doWrap
```
amosr: Because there were no primitives that returned Possibly before, the return type of the application itself never needed to change - it was being unwrapped and rewrapped, so the primitives were still operating over real, definite values. Now that `map_insert` returns Possibly, we need to change the type annotation for applications to include the Sum.

nhibberd:
:+1:

tranma:
update ice when this is merged too
💯

PR #576, opened 2017-06-24
Rename SeaProgramAttribute -> Cluster and move to its own area

jystic:
Another quick cleanup, this renames `SeaProgramAttribute` -> `Cluster` and `SeaProgramCompute` -> `Kernel` and moves it to `Icicle.Sea.Data`. I also moved `SeaName` to `Icicle.Sea.Name` so it's up at the same level. Neither are really avalanche specific after all.

So the idea is that each input has a cluster of kernels which execute the relevant queries for that input.

! @tranma

/cc @amosr

### DIFF
--- a/icicle-compiler/icicle.cabal
+++ b/icicle-compiler/icicle.cabal
@@ -130,7 +130,12 @@ library
 
                        Icicle.Simulator
 
+                       Icicle.Sea.Data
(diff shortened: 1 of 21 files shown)


jystic:
Pushing this one through

tranma:
👍 
ack

PR #578, opened 2017-06-27
Support for encoding / decoding C dictionary header

jystic:
This adds support for encoding / decoding a header which we will attach to the start of C dictionaries.

The header looks a bit like:

```c
//---------------------
// ICICLE C DICTIONARY
//---------------------
//
// {
//   "version": "1",
//   ..
// }
//
//---------------------
// ICICLE C DICTIONARY
//---------------------
// .. start of C code 
```
! @tranma

### DIFF
--- a/icicle-compiler/src/Icicle/Sea/Header.hs
+++ b/icicle-compiler/src/Icicle/Sea/Header.hs
@@ -1,0 +1,557 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Icicle.Sea.Header (
+    Header(..)
+  , Fingerprint(..)
+  , renderHeader
+  , parseHeader
+
+  , HeaderDecodeError(..)
+  , renderHeaderDecodeError
+
+  , pFingerprint
+  , pStructType
+  , pValType
+  , pMeltedType
+  , pClusterId
+  , pKernelIndex
+  , pKernelId
+  , pSeaName
+  , pInputId
+  , pOutputId
+  , pMeltedOutput
+  , pNamedType
+  , pKernel
+  , pCluster
+  , pHeader
+
+  , ppFingerprint
+  , ppStructType
+  , ppValType
+  , ppMeltedType
+  , ppClusterId
+  , ppKernelIndex
+  , ppKernelId
+  , ppSeaName
+  , ppInputId
+  , ppOutputId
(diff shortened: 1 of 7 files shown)


icicle-compiler/src/Icicle/Sea/Header.hs
```diff
@@ -0,0 +1,557 @@
...
+        "serialised melted type did not match calculated melted type:" <>
+        "\n" <>
+        "\n  logical type =" <>
+        "\n    " <> show logical <>
+        "\n" <>
+        "\n  serialised melted type =" <>
+        "\n    " <> show melted0 <>
+        "\n" <>
+        "\n  calculated melted type =" <>
+        "\n    " <> show melted
```
tranma: when can this happen? looks like both the serialised melted type `melted0` and the re-computed melted type `melted` are produced by `meltType` from the same inputs at different points in time? maybe if we are reading a wrong dictionary?

icicle-compiler/src/Icicle/Sea/Header.hs
```diff
@@ -0,0 +1,557 @@
...
+    , "clusters" .=
+        ppList ppCluster (headerClusters x)
+    ]
+
+headerKeyOrder :: [Text]
+headerKeyOrder = [
+  -- Header
+    "version"
+  , "fingerprint"
+  , "clusters"
```
tranma: what's the fingerprint of the C dictionary? a hash of the TOML dictionary?

tranma:
code looks good, some q's re. specification

icicle-compiler/src/Icicle/Sea/Header.hs
jystic: Yeah it's `'impossible'` in the GHC sense, it would happen if the melting algorithm changed, I figured there is no harm checking it.

The alternative is to never store the melted types are just regenerate them.

icicle-compiler/src/Icicle/Sea/Header.hs
jystic: Just a string, I was thinking of putting the icicle version in there, but if we call this from blizzard it would be the blizzard version instead.

We can only get the version/git hash in the main executable.

icicle-compiler/src/Icicle/Sea/Header.hs
tranma: in situations like these I prefer to not store the results tbh. This is alright but I'd rather not store the melted types altogether

tranma:
discussed IRL re. the inclusion of intermediate internals such as melted types, cluster ids and where to check them...
it's ok as-is unless we have a better story for ensuring the executable that generates some C dictionary is the same as the one used to run it.
👍

PR #579, opened 2017-06-28
Update array tests to use Hedgehog

jystic:
This updates the array tests to use the shiny new state machine testing now available in Hedgehog.

Figured it was time to bite the bullet and start trying to roll it out.

! @tranma @amosr 
/jury approved @amosr

### DIFF
--- a/icicle-compiler/src/Icicle/Runtime/Data/Array.hs
+++ b/icicle-compiler/src/Icicle/Runtime/Data/Array.hs
@@ -76,7 +76,7 @@ import qualified Data.Vector.Storable.Mutable as MStorable
 import           Data.Void (Void)
 import           Data.Word (Word8)
 
-import           Foreign.C.Types (CSize(..))
+import           Foreign.C.Types (CInt(..), CSize(..))
 import           Foreign.C.String (CString)
 import           Foreign.ForeignPtr (newForeignPtr_, withForeignPtr)
 import           Foreign.Ptr (Ptr, plusPtr, castPtr)
@@ -297,9 +297,24 @@ grow pool arrayOld@(Array ptrOld) lengthNew = do
     !capacityNew =
       calculateCapacity lengthNew
 
-  if capacityOld >= capacityNew then do
+  if lengthOld > lengthNew then do
+    let
+      deadSize =
+        (lengthOld - lengthNew) * elementSize
+
+      deadPtr =
+        unsafeElementPtr arrayOld `plusPtr`
+        fromIntegral (lengthNew * elementSize)
+
+    c_memset deadPtr 0 (fromIntegral deadSize)
+    unsafeWriteLength arrayOld lengthNew
+
+    pure arrayOld
+
+  else if capacityOld >= capacityNew then do
     unsafeWriteLength arrayOld lengthNew
     pure arrayOld
+
   else do
     let
       !bytesOld =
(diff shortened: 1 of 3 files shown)


icicle-compiler/test/Icicle/Test/Runtime/Array.hs
```diff
@@ -1,96 +1,359 @@
...
+    n <- forAll $ Gen.int (Range.constant 2 62)
+    2 ^ n === Array.calculateCapacity (ArrayLength (2 ^ n))
 
 prop_roundtrip_writes :: Property
 prop_roundtrip_writes =
-  gamble (listOf (bounded :: Jack Int64)) $ \xs0 ->
-  testIO . bracket Mempool.create Mempool.free $ \pool -> do
-    array <- Array.new pool (fromIntegral (length xs0))
+  property $ do
+    xs0 <- forAll . Gen.list (Range.linear 0 100) $ Gen.int64 Range.linearBounded
```
tranma: so this is a list that grows linearly from 0 to 100, containing int64s of from minBound..maxBound?
💯 💯 💯

jystic: yep! (and the int64s grow linearly too, from [0,0] to [minBound,maxBound])

icicle-compiler/test/Icicle/Test/Runtime/Array.hs
```diff
@@ -1,96 +1,359 @@
...
+            new `List.isPrefixOf` old
+      ]
+
+------------------------------------------------------------------------
+
+prop_array_state_machine :: Property
+prop_array_state_machine =
+  property $ do
+    actions <- forAll $
+      Gen.actions (Range.linear 1 100) initialState [newArray, readArray, writeArray, growArray]
```
tranma: omg! :1st_place_medal: :1st_place_medal: :1st_place_medal:

amosr:
💯

icicle-compiler/src/Icicle/Runtime/Data/Array.hs
```diff
@@ -297,9 +297,24 @@ grow pool arrayOld@(Array ptrOld) lengthNew = do
...
+  if lengthOld > lengthNew then do
+    let
+      deadSize =
+        (lengthOld - lengthNew) * elementSize
+
+      deadPtr =
+        unsafeElementPtr arrayOld `plusPtr`
+        fromIntegral (lengthNew * elementSize)
+
+    c_memset deadPtr 0 (fromIntegral deadSize)
```
jystic: I actually found a bug, it was expected that when you grow an array, all the newly added elements start off as `0`, this wasn't the case if you shrink an array then grow it again:

```hs
Var 2 = NewArray (ArrayLength 1)
Var 4 = WriteArray (ArrayIndex 0) (-1) (Var 2)
Var 5 = GrowArray (ArrayLength 0) (Var 2)
Var 6 = GrowArray (ArrayLength 1) (Var 5)
Var 7 = ReadArray (ArrayIndex 0) (Var 6) -- reads the -1 value!
```

amosr: that's really cool! does that mean grow should zero out everything between the old count and the new count?

jystic: Yeah, so I decided the easiest way to do it was to make sure the unused memory is always initialised to zero. So when we shrink from an array length of say `10` to a length of `5` we need to zero out indices `5` to `9`.

amosr: sounds good. I don't know, do we even have a way to shrink arrays from source? is there `map_delete`?

jystic: This is really just for use from Haskell, although I suppose we could use it to test the C code itself.

PR #581, opened 2017-07-03
Max map size: clip inputs before evaluating

amosr:
Make sure the inputs are not larger than the max map size.
#575 

! @jystic @tranma 
/jury approved @jystic


icicle-compiler/test/Icicle/Test/Source/MaxMapSize.hs
```diff
@@ -11,33 +11,77 @@ import qualified Icicle.Core.Eval.Program   as PV
...
 import           Icicle.Common.Base
+import           Icicle.Data.Fact (AsAt(..))
 
 import           P
 
 import           System.IO
 
 import           Test.QuickCheck
+
+import           Control.Arrow ((***))
```
jystic: 🎩

jystic:
👍

PR #582, opened 2017-07-03
Stop torturing gcc

jystic:
! @amosr @nhibberd 
/jury approved @amosr


amosr:
👍 
would also be interesting to go through the user manual for that version of gcc and see if anything obvious that would be causing it

PR #584, opened 2017-07-03
Start on evaluator which can call kernels directly

jystic:
Not much here yet, just an idea of how compilation will look. You can see how we pull the cluster information out of the C file. It's set up so that even if we have an avalanche program we still go through the same steps for converting it to runtime executable functions.

! @amosr @tranma


jystic:
Ignore this, I'll put a PR when I'm done

PR #587, opened 2017-07-04
FoldBool: change arguments to be lambdas rather than values.

amosr:
```
- if :: a -> a -> Bool -> a
+ if :: (() -> a) -> (() -> a) -> Bool -> a
```
if was sometimes strict in both its arguments because it would anormalise and pull out any let-bindings inside the branch.

! @jystic @tranma


tranma:
yes, this makes sense
💯

PR #588, opened 2017-07-04
Typechecking changes prior to NaN

amosr:
Application: when result and argument both have mode, but expected argument type
is pure, join the result and argument modes rather than requiring
equality.  New result is the result of join.

Remove Cons Hack for fold: this was prematurely committing to Definitely, when it shouldn't have.

Perform top-level constraint discharge before checking for leftover constraints. This finds constraints like `a = PossibilityJoin b c` and rewrites `a = b = c`. This is necessary because conversion to Core doesn't deal with polymorphic modes, so if one is a non-variable mode this will solve it.

Case: return single substitution, and apply substitution to preceding alternative. When you have two alternatives, you recurse into the first alternative, then the second. If you learn something from the second alternative, you need to go back and apply that to the first alternative.

Defaults: default to Unit, Pure or Definitely based on kind. Find all the types in the program and decompose them into temporality, possibility and data. Then find the free variables of the decomposed and choose a sensible default. Before, any type variable possibilities would have been substituted to Unit. (But this probably wasn't happening often because of the Cons Hack for folds would change make it concrete)

Function: don't "clean up" the constraints on function types. Nobody looks at the function types, so it doesn't matter if they're ugly. And the "cleaning up" was over-specific.

! @jystic @tranma 
/jury approved @jystic

### DIFF
--- a/icicle-source/src/Icicle/Source/Checker/Constraint.hs
+++ b/icicle-source/src/Icicle/Source/Checker/Constraint.hs
@@ -717,7 +717,7 @@ generateP ann scrutTy resTy resTm resPs ((pat, alt):rest) env
 
 
 appType
- :: (Hashable n, Eq n)
+ :: (Hashable n)
  => a
  -> Type n
  -> (Type n, Type n)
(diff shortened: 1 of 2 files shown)


jystic:
👍 seems legit

Would be good to test this changes out and make sure nothing breaks in production

amosr:
Tested this change on real data, output was identical to master. So this one is OK to merge, but the nan one needs more work

PR #589, opened 2017-07-04
Topic/nan again

amosr:
Nan properly
/jury approved @jystic

### DIFF
--- a/icicle-source/src/Icicle/Source/Query/Prim.hs
+++ b/icicle-source/src/Icicle/Source/Query/Prim.hs
@@ -4,6 +4,7 @@
 -- So we must generate a fresh name for any forall binders.
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternGuards #-}
 module Icicle.Source.Query.Prim (
     primLookup'
   , primReturnsPossibly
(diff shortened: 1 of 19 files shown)


icicle-source/src/Icicle/Source/Transform/ReifyPossibility.hs
```diff
@@ -296,15 +296,7 @@ makeApps a fun (arg:rest) doWrap
...
- =  makeApps a (App annotWrapPrim fun arg) rest doWrap
-
- where
-  annotWrapPrim
-   | Just (p, _, _) <- takePrimApps fun
-   , primReturnsPossibly p
-   = wrapAnnotReally a
-   | otherwise
-   = a
+ =  makeApps a (App a fun arg) rest doWrap
```
amosr: This is a rejigging of https://github.com/ambiata/icicle/commit/130a02bd2d211743960724e72afd8d2da6824051
We were wrapping the application annotations with `SumT ErrorT`, but not the primitive itself

amosr:
Will create a new exception type `ExceptNotANumber` as next PR, for now this is using `ExceptCannotCompute` when it encounters a NaN

icicle-repl/test/cli/repl/t16-prelude/expected
```diff
@@ -66,20 +66,20 @@ ok, loaded test/cli/repl/data.psv, 13 rows
...
 > > -- Pearsons correlation example
 > - C evaluation:
 [homer, 0.9279829441311143
-,marge, NaN]
+,marge, tombstone]
 
 - Core evaluation:
 [homer, 0.9279829441311143
-,marge, NaN]
+,marge, tombstone]
```
jystic: 🎆 🍾

icicle-source/src/Icicle/Source/Query/Prim.hs
```diff
@@ -22,14 +23,19 @@ import                  Data.Hashable (Hashable)
 
 
 primLookup' :: Hashable n => Prim -> Fresh.Fresh n (FunctionType n)
-primLookup' p
- = case p of
-    Op (ArithUnary _)
-     -> fNum $ \at -> ([at], at)
+primLookup' prim
+ = case prim of
+    -- Negate on Doubles will not introduce NaN or Inf
```
jystic: Well, I guess that's something :)

icicle-source/src/Icicle/Source/Query/Prim.hs
```diff
@@ -156,8 +179,25 @@ primLookup' p
...
+ , DoubleT             <- dat
+ , Just PossibilityPossibly <- pos
+ = case p of
+    Op (ArithBinary _)     -> True
+    Op (ArithDouble Div)   -> True
+    Fun (BuiltinMath Log)  -> True
+    Fun (BuiltinMath Exp)  -> True
+    Fun (BuiltinMath Sqrt) -> True
+    _                      -> False
+primReturnsPossibly _ _                          = False
```
jystic: Is there any way to check that this lines up with what we expect?

jystic: (as in, if we added a new prim that needs to be here too, will a test fail?)

jystic:
👍

icicle-source/src/Icicle/Source/Query/Prim.hs
amosr: OK, I think if I write a source generator that just does expressions with primitives and values, this would catch it better than the current `convert_is_well_typed`

amosr:
This needs to wait because the reified introduces tricky cases we aren't dealing with properly. #597.

Issue #524, opened 2017-05-05
Bug: convert_ok cannot check key


tranma:
```
=== prop_convert_ok from test/Icicle/Test/Source/Convert.hs:19 ===
*** Failed! Falsifiable (after 613 tests): 
QueryWithFeature:
  Query: group `2008-03-16` ~> chefnow
  Now: chefnow
  Output: animal
  Type: Struct []
  Name: cheffeaturename
  Time: cheftimename
  Key: Left (box None)
feature cheffeaturename
~> group `2008-03-16` ~> chefnow
Left (ConvertErrorCannotCheckKey () (XApp () (XPrim () (PrimMinimal (PrimConst (PrimConstLeft UnitT UnitT)))) (XValue () (OptionT UnitT) VNone)) (ExpErrorApp (XPrim () (PrimMinimal (PrimConst (PrimConstLeft UnitT UnitT)))) (XValue () (OptionT UnitT) VNone) (FunT [FunT [] UnitT] (SumT UnitT UnitT)) (FunT [] (OptionT UnitT))))
```

tranma:
fact key -- prob. my fault, will look later

amosr:
the test wasn't calling reifyPossibility on the key, but the real dictionary parser was

PR #590, opened 2017-07-05
ReifyPossibility of Key

amosr:
#524 

Test wasn't calling ReifyPossibility on the query key. Dictionary parser is. Had to expose `reifyPossibilityX` to call from test.

! @jystic @tranma 
/jury approved @tranma @jystic


tranma:
👍

jystic:
👍

PR #591, opened 2017-07-05
Conversion of striped columns to and from runtime arrays

jystic:
This adds support for converting between striped <-> logical representation of icicle values, and between striped <-> melted array representation of icicle values.

The main thing to check is the `Icicle.Runtime.Array` module and the `toArrays` / `fromArrays` functions in `Icicle.Runtime.Data.Striped`. The rest of the stuff is just a specialisation of what zebra does.

! @tranma @amosr


icicle-compiler/src/Icicle/Runtime/Array.hs
```diff
@@ -358,5 +405,148 @@ fromList pool xs =
...
+          !n_csize =
+            fromIntegral n
+
+          !n_int =
+            fromIntegral n
+
+        dstElemPtr <- Mempool.allocBytes pool (n_csize + 1)
+
+        c_memcpy dstElemPtr (srcPtr `plusPtr` off) n_csize
+        pokeByteOff dstElemPtr n_int (0 :: Word8)
```
tranma: is `n_csize` definitely `Int64` and `n_int` might be 32?

icicle-compiler/src/Icicle/Runtime/Array.hs
```diff
@@ -358,5 +405,148 @@ fromList pool xs =
...
+  Storable.foldM_ loop 0 srcArrays
+
+  pure (ns, dstArray)
+{-# INLINE toArraySegments #-}
+
 foreign import ccall unsafe "string.h memcpy"
   c_memcpy :: Ptr a -> Ptr a -> CSize -> IO ()
+
+foreign import ccall unsafe "string.h strlen"
+  c_strlen :: CString -> IO CSize
```
tranma: this module looks ok

icicle-compiler/src/Icicle/Runtime/Data/Primitive.hs
```diff
@@ -0,0 +1,198 @@
...
+  NotAnError64 ->
+    ExceptNotAnError
+  Tombstone64 ->
+    ExceptTombstone
+  Fold1NoValue64 ->
+    ExceptFold1NoValue
+  CannotCompute64 ->
+    ExceptCannotCompute
+  _ ->
+    ExceptTombstone
```
tranma: this could be a source for mistakes if we introduce a new error tag and forget to update this? We might end up seeing spurious tombstones where there is none.

icicle-compiler/src/Icicle/Runtime/Array.hs
jystic: `n_csize :: CSize` and `n_int :: Int`, they hold the same value though

icicle-compiler/src/Icicle/Runtime/Data/Primitive.hs
jystic: Yeah is a good point, maybe I'll make it a `Maybe` instead

icicle-compiler/src/Icicle/Runtime/Data/Striped.hs
```diff
@@ -0,0 +1,594 @@
...
+  Schema.Sum x y ->
+    1 + arrayCount x + arrayCount y
+  Schema.Option x ->
+    1 + arrayCount x
+  Schema.Result x ->
+    1 + arrayCount x
+  Schema.Pair x y ->
+    arrayCount x + arrayCount y
+  Schema.Struct fields ->
+    Boxed.sum . Cons.toVector $! fmap (arrayCount . fieldData) fields
```
tranma: worth making sure the other cases don't build up thunks too?

icicle-compiler/src/Icicle/Runtime/Data/Striped.hs
```diff
@@ -0,0 +1,594 @@
...
+          vs <- fromLogical sv vs0
+
+          pure $
+            Map ns ks vs
+
+fromField :: Field Schema -> Boxed.Vector Logical.Value -> Either StripedError (Field Column)
+fromField field =
+  fmap (field $>) .
+  fromLogical (fieldData field)
+{-# INLINABLE fromField #-}
```
tranma: this looks good

tranma:
yup, looks like this runtime array representation matches icicle
📦

icicle-compiler/src/Icicle/Runtime/Data/Striped.hs
jystic: done

PR #595, opened 2017-07-05
Bare bones evaluator which allows queries over striped data

jystic:
This probably doesn't even work yet but just wanted to give an overview of the design in a piecemeal way.

I will hook it up to the corpus / well typed tests when #549 gets merged. For the moment I'm going to focus on pushing it all the way through to the repl.

! @amosr @tranma

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 dist
 cabal-dev
-gen
 *.o
 *.hi
 *.chi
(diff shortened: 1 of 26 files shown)


amosr:
looks plausible 👍

jystic:
> looks plausible 👍 

(only barely!)

PR #596, opened 2017-07-05
Move REPL to the main icicle executable

jystic:
I think having all the icicle functionality together in a single executable will make it more discoverable and nicer to use. Also we get proper command line parsing now instead of `getArgs` :)

Note that this will mean that the REPL will get deployed to boxes now as part of our normal process.

```
Usage: icicle ((-v|--version) | COMMAND)

Available options:
  -v,--version             Version information
  -h,--help                Show this help text

Available commands:
  repl                     Interactively evaluate icicle expressions.
  compile                  Compile a dictionary to its C intermediate form.
  query                    Run an icicle query over some data.
```

```
Usage: icicle repl [--init COMMAND]
  Interactively evaluate icicle expressions.

Available options:
  --init COMMAND           A command to execute in the REPL before interative
                           evaluation begins.
  -h,--help                Show this help text
```

! @amosr @tranma 
/jury approved @amosr


amosr:
👍

PR #598, opened 2017-07-06
Typecheck Core, and pretty errors

amosr:
We weren't typechecking the Core, and, badness upon badness, the Avalanche simplifier was simplifying out the ill-typed expressions

! @jystic @tranma 
/jury approved @jystic


amosr:
Running on some real data, will know whether it makes compilation take any longer soon

amosr:
master: 1297s, this: 1351s. no difference in compile time. this isn't too surprising since it's typechecking the Core before fusion, so each program will be quite small.

jystic:
👍

Issue #597, opened 2017-07-06
Case with pure scrutinee and aggregate return is broken


amosr:
If the scrutinee is pure but the return type is aggregate, the pattern variables are bound in the wrong place

```
> :set +core -core-simp

> feature salary ~> case (0,0) | (b,c) -> (fold a = b : c ~> a) end

- Core (not simplified):
Program (conv/0 : (Sum Error Int), conv/1 : FactIdentifier, conv/2 : Time, conv/3 : SNAPSHOT_TIME, conv/4 : MaxMapSize)
Precomputations:
  conv/5               = pair#@{Int, Int} (0@{Int}) (0@{Int})

Streams:
  STREAM_FOLD (a/conv/8 : Int)
    INIT:
      desugar_q/0/conv/6
    KONS:
      desugar_q/1/conv/7


Postcomputations:
  conv/10              =
                           let conv/9 = conv/5
                           let desugar_q/0/conv/6 = fst#@{Int, Int} conv/9
                           let desugar_q/1/conv/7 = snd#@{Int, Int} conv/9
                            in a/conv/8

Returning:
  repl:output          = conv/10


- Core type error:
Stream error:
    Worker function type error:
        Variable not bound: desugar_q/0/conv/6
```

amosr:
...

PR #600, opened 2017-07-06
Case with pure scrutinee and aggregate return is broken #597

amosr:
Fix in two ways. One. make temporality of pattern bindings equal to result
temporality.
So this is outlawed, because the bindings 'b' and 'c' are Aggregates,
despite the scrutinee being pure.

> feature salary ~> case (0,0) | (b,c) -> (fold a = b : c ~> a) end

This makes the type checker more restrictive, but this shouldn't affect
any existing programs because they would currently fail to typecheck in
Core.

Two. Converting an Aggregate case to Core, some of the references will show up
in the precomputations. But the pattern isn't bound until post.
So take the set of names bound by the pattern, and run through the
precomputations, deferring anything that mentions the pattern names
until post. That way it will be nested beneath the pattern bindings.


jystic:
👍

amosr:
Tested in production. Output matches master, after nan -> NA regex

PR #601, opened 2017-07-06
Add support for converting icicle striped <-> zebra striped

jystic:
Almost there! This does the conversion for inputs, outputs is almost the same but will be in the next PR.

! @amosr @tranma 
/jury approved @amosr

### DIFF
--- a/icicle-compiler/src/Icicle/Runtime/Serial/Zebra.hs
+++ b/icicle-compiler/src/Icicle/Runtime/Serial/Zebra.hs
@@ -1,0 +1,578 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Icicle.Runtime.Serial.Zebra (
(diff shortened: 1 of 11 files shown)


icicle-compiler/src/Icicle/Runtime/Serial/Zebra.hs
```diff
@@ -0,0 +1,467 @@
...
+
+  Zebra.Double _ xs ->
+    pure $ Icicle.Double xs
+
+  Zebra.Enum _ tags variants ->
+    case Cons.toList variants of
+      [Zebra.Variant "false" _, Zebra.Variant "true" _] ->
+        pure $ Icicle.Bool (decodeBoolTags tags)
+
+      [Zebra.Variant "none" _, Zebra.Variant "some" x] ->
```
amosr: should this check and fail if `none` isn't unit?

amosr:
👍

icicle-compiler/src/Icicle/Runtime/Serial/Zebra.hs
jystic: yeah, probably should :)

PR #603, opened 2017-07-07
Split PSV test suite into 4 suites

tranma:
Hoping each of these takes about 10m

! @jystic
/jury approved @jystic


jystic:
👍 good stuff

PR #604, opened 2017-07-11
Overhaul REPL, add support for zebra input

jystic:
This change really only touches the repl and the pretty printing, not anything that we're running in production.

I started adding support for zebra to the repl and got a bit carried away overhauling it. I think the result is a lot nicer, try building it and play around yourself.

Not sure that this PR is reviewable, but again, nothing critical has changed so I'm not worried about introducing any production bugs here.

/cc @tranma @amosr


icicle-compiler/src/Icicle/Repl/Base.hs
```diff
@@ -1,156 +0,0 @@
...
-
-
-terminalWidth :: HL.InputT IO (Maybe Int)
-terminalWidth
- = fmap (fmap TS.width)
- $ liftIO TS.size
-
-showFlag :: Bool -> String
-showFlag True  = "on"
-showFlag False = "off"
```
tranma: 👍

tranma:
I reviewed the list of changed files....
👍

Issue #606, opened 2017-07-13
Beta reduction should simplify lets the same as lambdas


jystic:
```
let x = y in x
==>
y
```

tranma:
not sure why we only do it when `y` is a value (https://github.com/ambiata/icicle/blob/master/icicle-data/src/Icicle/Common/Exp/Simp/Beta.hs#L38), gotta discuss with the big boss

amosr:
Yeah, so variable should be fine, but anything else you might duplicate work. `isSimpleValue` (https://github.com/ambiata/icicle/blob/master/icicle-data/src/Icicle/Common/Exp/Simp/Beta.hs#L77) is really "is simple or value" rather than simple and value.
So maybe betaToLets should just substitute simple/values in.

Is this really a problem though?

jystic:
It was just to improve the core output so it's easy to read.

amosr:
Oh ok, cool

amosr:
So is `y` a variable in this case, or an expression? For general expressions we'd want something like free variables `freevars' :: Exp n -> Map n Count` to see if it's used multiple times.
We might be able to get away with just matching on `XLet x y (XVar x)` though?

jystic:
`y` is an expression and a pretty decent one at that, that's why it would be nice to reduce.

Yeah matching on `XLet x y (XVar x)` sounds like it would work I think.

tranma:
@amos yeah I'm ok with matching on `XLet x y (XVar x)` (the conversion even has a name, so it's legit!)

amosr:
Is that a specialised form of "zeta", or does it have its own name?

amosr:
I think zeta is more general `let x = e in e' --> e'[x := e]` but I don't know. in a CBV semantics you would restrict the binding to a value, but I guess that would work in CBN?

PR #609, opened 2017-07-17
Beta reduce special case when let body is variable

amosr:
Reduce `let x = e in x --> e`; when variables same, `let x = e in x' --> x'` when variables different.

It also turns out that beta was parameterised by the `isValue` predicate, but this was only ever called with `isSimpleValue`. So I've specialised it, which sped up compilation of a test dictionary slightly from 144s to 118s.

#606 

! @jystic @tranma 

### DIFF
--- a/icicle-compiler/test/Icicle/Test/Core/Exp/Simp.hs
+++ b/icicle-compiler/test/Icicle/Test/Core/Exp/Simp.hs
@@ -29,7 +29,7 @@ import           Test.QuickCheck
 prop_beta_evaluation
  = withTypedExp
  $ \x _
- -> let x' = Beta.beta Beta.isSimpleValue x
+ -> let x' = Beta.beta x
     in  counterexample (show $ pretty x)
       $ counterexample (show $ pretty x')
        (eval0 evalPrim x === eval0 evalPrim x')
@@ -39,17 +39,7 @@ prop_beta_type
  = withTypedExp
  $ \x _
  -> typeExp0 coreFragment x == typeExp0 coreFragment
-   ( Beta.beta Beta.isSimpleValue x)
-
--- Reduce regardless of whether it's a value
-prop_beta_always_evaluation
- = withTypedExp
- $ \x _
- -> let x' = Beta.beta (const True) x
-    in  counterexample (show $ pretty x)
-      $ counterexample (show $ pretty x')
-       (eval0 evalPrim x === eval0 evalPrim x')
-
+   ( Beta.beta x)
 
 -- Converting all beta reductions to lets
 prop_betaToLets_evaluation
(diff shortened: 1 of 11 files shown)


icicle-data/src/Icicle/Common/Exp/Simp/Beta.hs
```diff
@@ -15,16 +15,16 @@ import P
...
  where
   go xx
    = case xx of
       XApp _ p q
        | XLam _ n _ x <- go p
        , v <- go q
-       , isValue v
+       , isSimpleValue v
        , Just x' <- substMaybe n v x
        -> go x'
```
tranma: so if we have `(lambda x. y) a`, this will not reduce to `y` unless `y` is just a value, right?

amosr: it's if `a` is a value, not `y`. because the substitution might duplicate `a`, which isn't a good idea if `a` is expensive

tranma: ah, sorry, I was confused. I meant to ask whether `(lambda x. x) a` would reduce if `a` is not a simple value.

amosr: ah, no it won't, good idea

icicle-compiler/test/cli/repl/t10-avalanche/expected
```diff
@@ -91,23 +91,16 @@ read c-conv-11 =
 let conv-31 =
   Sum_fold#
     (\reify-10-conv-21 ->
-      let
-        conv-22 =
-          left# reify-10-conv-21
-      in
-        conv-22)
+      left# reify-10-conv-21)
```
jystic: 🎸

jystic:
👍 looks legit to me

icicle-compiler/test/Icicle/Test/Core/Exp/Simp.hs
```diff
@@ -39,17 +39,7 @@ prop_beta_type
...
-   ( Beta.beta Beta.isSimpleValue x)
-
--- Reduce regardless of whether it's a value
-prop_beta_always_evaluation
- = withTypedExp
- $ \x _
- -> let x' = Beta.beta (const True) x
-    in  counterexample (show $ pretty x)
-      $ counterexample (show $ pretty x')
-       (eval0 evalPrim x === eval0 evalPrim x')
```
tranma: this test would still make sense with `let x' = beta x`, wouldn't it?

icicle-data/src/Icicle/Common/Exp/Simp/Beta.hs
```diff
@@ -34,8 +38,13 @@ beta isValue toplevel
...
+      XLet _ n v (XVar a' n')
+       -> if   n == n'
+          then go v
+          else XVar a' n'
+
       XLet _ n v x
-       | isValue v
+       | isSimpleValue v
        , Just x' <- substMaybe n v x
        -> go x'
```
tranma: this looks fine

icicle-compiler/test/Icicle/Test/Core/Exp/Simp.hs
amosr: yeah, there's already one like that earlier in the file

tranma:
👍

PR #612, opened 2017-07-25
Delete PSV C backend

jystic:
This removes the PSV C backend.

Doing that forced me to hook the psv part of the repl up to the new evaluator, and I also wrote a corpus test for the new evaluator.

! @tranma @amosr


icicle-compiler/src/Icicle/Compiler/Sea.hs
```diff
@@ -0,0 +1,192 @@
...
+      fmap (Entity . Text.decodeUtf8 . Runtime.unEntityId . Runtime.entityId . Runtime.snapshotEntity) .
+      Boxed.toList $
+      Runtime.outputKey output0
+
+  output2 <- hoistEither $ fromOutputs output1
+  output3 <- hoistEither $ traverse (\x -> maybeToRight (CompilerSeaValueFromCore x) $ Common.valueFromCore x) output2
+
+  pure .
+    fmap Result $
+    List.zipWith ((,)) entities output3
```
tranma: this seaEval is only used by the REPL right? If so I'd move it there just to be out of the way

icicle-compiler/src/Icicle/Runtime/Data/Logical.hs
```diff
@@ -454,3 +483,172 @@ takeMap = \case
...
+  Unit ->
+    pure VUnit
+  Bool False64 ->
+    pure $ VBool False
+  Bool _ ->
+    pure $ VBool True
+  Int x ->
+    pure $ VInt (fromIntegral x)
+  Double x ->
+    pure $ VDouble x
```
tranma: do you need to check `typ` for these cases?

icicle-compiler/src/Icicle/Runtime/Data/Logical.hs
```diff
@@ -454,3 +483,172 @@ takeMap = \case
...
+    pure $ Double x
+
+  VUnit ->
+    pure Unit
+
+  VBool False ->
+    pure $ Bool False64
+
+  VBool True ->
+    pure $ Bool True64
```
tranma: need to check ty?

tranma:
No PR has ever made me so happy

icicle-compiler/src/Icicle/Compiler/Sea.hs
jystic: Also used by Icicle.Test.Language and the conversion functions are used by Icicle.Test.Runtime.Corpus, but I think it can probably still live in the repl area

icicle-compiler/src/Icicle/Runtime/Data/Logical.hs
jystic: yeah, you're right, probably should

icicle-compiler/src/Icicle/Compiler/Sea.hs
jystic: It's supposed to have the same type more or less as `coreEval` and `avalancheEval` in `Icicle.Compiler`, that's why I put it here fwiw. I just couldn't put it in `Icicle.Compiler` because of a circular dependency.

icicle-compiler/src/Icicle/Runtime/Data/Logical.hs
jystic: Yep will do, this is your code ! 😆

jystic:
@tranma if you can think of anything that I've missed / haven't deleted that would be cool

tranma:
ok, I think you've deleted all the right stuff.
👍

PR #614, opened 2017-07-27
Calculate correct offset in to input column using all facts

jystic:
I was calculating the wrong offset in to the input data, using `ncounts_valid` instead of `ncounts_all`.

`ncounts_all` = the count of facts in the zebra file
`ncounts_valid` = the count of facts which occurred before snapshot time

! @tranma

### DIFF
--- a/icicle-compiler/test/Icicle/Test/Runtime/Evaluator.hs
+++ b/icicle-compiler/test/Icicle/Test/Runtime/Evaluator.hs
@@ -6,10 +6,12 @@ module Icicle.Test.Runtime.Evaluator where
 import           Control.Monad.IO.Class (liftIO)
 import           Control.Monad.Morph (hoist)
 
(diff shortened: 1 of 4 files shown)


icicle-compiler/test/Icicle/Test/Runtime/Evaluator.hs
```diff
@@ -119,25 +130,58 @@ fromInputColumn input =
...
+    stime <- forAll $ SnapshotTime <$> genTime64
 
     ccoptions0 <- Runtime.getCompilerOptions
 
     let
-      stime =
-        SnapshotTime . packTime $
-          UnpackedTime64 3000 1 1 0
+      input =
+        takeInput et
```
tranma: 👍

tranma:
this looks good to me
👍

PR #615, opened 2017-07-27
Runtime corpus test now uses more interesting snapshot times

jystic:
I also made the avalanche pretty printing a bit more compact / easier to read.

! @tranma

### DIFF
--- a/icicle-compiler/test/Icicle/Test/Runtime/Corpus.hs
+++ b/icicle-compiler/test/Icicle/Test/Runtime/Corpus.hs
@@ -121,12 +121,16 @@ handleDiscard :: MonadCatch m => PropertyT m a -> PropertyT m a
 handleDiscard =
   handleIf QuickCheck.isDiscard (const discard)
 
+timeEnd :: Time
+timeEnd =
+  unsafeTimeOfYMD 3000 1 1
+
 test_runtime_corpus :: CorpusId -> (InputName, OutputId, Text) -> Property
 test_runtime_corpus cid query =
   property . handleDiscard $ do
-    stime <- forAll $ Gen.arbitrary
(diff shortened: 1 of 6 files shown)


jystic:
! @amosr seeing as you're here

icicle-compiler/test/Icicle/Test/Runtime/Corpus.hs
```diff
@@ -121,12 +121,16 @@ handleDiscard :: MonadCatch m => PropertyT m a -> PropertyT m a
...
+  unsafeTimeOfYMD 3000 1 1
+
 test_runtime_corpus :: CorpusId -> (InputName, OutputId, Text) -> Property
 test_runtime_corpus cid query =
   property . handleDiscard $ do
-    stime <- forAll $ Gen.arbitrary
-    maxMapSize <- forAll $ Gen.int (Range.linear 1 100)
     wt <- forAll . Gen.quickcheck $ genWellTypedFromSource query
+    maxMapSize <- forAll $ Gen.int (Range.linear 1 100)
+    stime <- forAll $ Gen.element (timeEnd : fmap atTime (concat . Map.elems $ wtInputs wt))
```
amosr: funny thing is, I think `inputsForType` generates a date too, but `genWellTypedFromSource` is throwing it away

amosr:
pretty good 👍

icicle-compiler/test/Icicle/Test/Runtime/Corpus.hs
jystic: 😆

tranma: yeah, that was deliberate when I was generating multiple `Cluster`

PR #617, opened 2017-08-02
Remove FactIdentifiers from Avalanche

amosr:
Mainly deleting code. There is a bit of cruft left in the tests, but it's easier to leave that for the next PR, when I remove it from Core

! @jystic @tranma


tranma:
good, glad to see it gone
👍

PR #618, opened 2017-08-02
Melting relation primops: (SumT ErrorT t) special case

amosr:
Less-than etc weren't working for the too-smart `(Error,t)` representation. In a melted normal sum the tag is True for Right, which makes `left < right`. In the `(Error,t)` case, the "tag" is ExceptNotAnError==0 for Right, so `left > right`!

So I modified the melt type to distinguish between bare errors and tag errors. When we see a tag error, we can convert it to an "isRight" by using `error == ErrorNotAnError`. Then we can compare on that as before, pretending it's a real sum with that as the tag.

This *doesn't* change (Sum Error t) nested inside arrays, bufs, etc. But, the relation semantics for these are already different. Comparison of pair-of-arrays vs array-of-pairs are different. So the tests are already ignoring these.

! @jystic @tranma


tranma:
yes, fair enough.
👍

PR #621, opened 2017-08-03
Generate a bit more

amosr:
Introduce some new primitives to the Core generator, as well as small changes necessary to make pass.

Remove the ToString primitives: these weren't exposed in Source, and weren't generating C code either.

Change the type of Pow from `Num a => a -> a -> a` to only work on Doubles. This won't affect real programs because Pow wasn't generating valid code for Ints before, anyway.

! @jystic @tranma 
/jury approved @jystic


amosr:
Next PR will add interesting Double primitives to generator, and modify tests to use NanEq instead of Eq.

jystic:
🐻

PR #622, opened 2017-08-03
nan it all to hell

amosr:
I decided to move NanEq into icicle-data, because alpha equality, which is defined in icicle-data needed to use it.

! @jystic @tranma 
/jury approved @jystic


tranma:
wow a new notion of alpha equivalence

amosr:
I think I've used this pun before.
I also changed the Map type generator, so it would only generate Ord types or primitives. Basically no arrays or maps inside the key of a Map.

amosr:
#625

jystic:
👍

PR #626, opened 2017-08-10
Chords using Icicle.Runtime.Stencil

jystic:
`snapshotBlock` vs `chordBlock` is now the only difference between a snapshot and a chord.
/jury approved @tranma

### DIFF
--- a/icicle-compiler/src/Icicle/Runtime/Evaluator.hs
+++ b/icicle-compiler/src/Icicle/Runtime/Evaluator.hs
@@ -25,6 +25,7 @@ module Icicle.Runtime.Evaluator (
   , Sea.getCompilerOptions
 
   , snapshotBlock
+  , chordBlock
 
   , runQuery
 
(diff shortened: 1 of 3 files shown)


icicle-compiler/src/Icicle/Runtime/Evaluator.hs
```diff
@@ -517,36 +519,63 @@ runQuery cluster maxMapSize stencil input =
...
+    missingInputs =
+      clusters `Map.difference` inputs
+
+    missingClusters =
+      inputs `Map.difference` clusters
+  in
+    if Map.size both /= Map.size inputs then
+      Left $ RuntimeInputClusterMismatch (Map.keysSet missingInputs) (Map.keysSet missingClusters)
+    else
+      pure both
```
tranma: is this just the same as comparing `inputs` and `clusters` by their sizes?

tranma:
👍

PR #631, opened 2017-08-14
Pretty with annotations

amosr:
When prettying with annotations, we were annotating the return type of the let, instead of the binding type, which is kind of confusing. It was also printing the expressions in lets and folds without annotations, so I changed that. Also, there's not much point annotating monomorphic primitives. `True:{Bool}` is boring, but `None:{Option Double}` is interesting.

Example below. It's not as pretty because of the annotations inside the case, now, but I think it's more useful.

```
feature "injury"
~> let v:{ (Bool, Int) } =
     ((,):{ (Bool, Int) } True 1)
~> fold s2:{ Bool } =
     False : case:{ Element (Definitely Bool) } case:{ Element (Definitely Bool) } v:{ (Bool, Unit) }
     | (desugar_q-0, desugar_q-1) ->
         desugar_q-0:{ Element Bool }
     end
     | True ->
         False
     | False ->
         True
     end
~> s2:{ Aggregate (Definitely Bool) }
```

! @jystic @tranma 
/jury approved @jystic

### DIFF
--- a/icicle-source/src/Icicle/Source/PrettyAnnot.hs
+++ b/icicle-source/src/Icicle/Source/PrettyAnnot.hs
@@ -83,19 +83,23 @@ instance (Pretty a, Pretty n) => Pretty (PrettyAnnot (Context a n)) where
       Filter a x ->
         prettyAnnotK "filter" a <+> pretty (PrettyAnnot x)
 
-      LetFold a f ->
+      -- For fold and let-bindings, the annotation holds the *return* type, but this isn't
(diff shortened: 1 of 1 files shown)


tranma:
so I think we decided to print without annotations in lets and folds in particular because they were very hard to read. You read this way more often though, so up to you.

Still would be nice to have an option to turn them off.

tranma:
ah right, cool

jystic:
👍 on 🍇

Issue #625, opened 2017-08-10
prop_flatten_simp_commutes_value failure


jystic:
```
  ✗ prop_flatten_simp_commutes_value failed after 94 tests and 50 shrinks.

       ┏━━ test/Icicle/Test/Avalanche/Flatten.hs ━━━
    83 ┃ prop_flatten_simp_commutes_value = property $ do
    84 ┃  t <- forAll genInputType
       ┃  │ UnitT
    85 ┃  o <- forAll genOutputType
       ┃  │ ArrayT (ArrayT (SumT ErrorT BoolT))
    86 ┃  p <- forAll (programForStreamType t o)
       ┃  │ Program
       ┃  │   { inputType = UnitT
       ┃  │   , factValName =
       ┃  │       Name
       ┃  │         { nameHash = 353389297757728241
       ┃  │         , nameBase = NameBase (Var "input" 3)
       ┃  │         }
       ┃  │   , factTimeName =
       ┃  │       Name
       ┃  │         { nameHash = -9206499228160308544
       ┃  │         , nameBase = NameBase (Var "factTime" 1)
       ┃  │         }
       ┃  │   , snaptimeName =
       ┃  │       Name
       ┃  │         { nameHash = -2168093946657691001
       ┃  │         , nameBase = NameBase (Var "snapshotTime" 2)
       ┃  │         }
       ┃  │   , maxMapSize =
       ┃  │       Name
       ┃  │         { nameHash = -728873976287134335
       ┃  │         , nameBase = NameBase (Var "maxMapSize" 0)
       ┃  │         }
       ┃  │   , precomps =
       ┃  │       [ ( Name
       ┃  │             { nameHash = -5974379687266208255
       ┃  │             , nameBase = NameBase (Var "american_curl" 4)
       ┃  │             }
       ┃  │         , XVar
       ┃  │             ()
       ┃  │             Name
       ┃  │               { nameHash = -2168093946657691001
       ┃  │               , nameBase = NameBase (Var "snapshotTime" 2)
       ┃  │               }
       ┃  │         )
       ┃  │       ]
       ┃  │   , streams =
       ┃  │       [ SFold
       ┃  │           Name
       ┃  │             { nameHash = -5974379687266208256
       ┃  │             , nameBase = NameBase (Var "american_curl" 5)
       ┃  │             }
       ┃  │           TimeT
       ┃  │           (XVar
       ┃  │              ()
       ┃  │              Name
[...]

PR #633, opened 2017-08-14
Generate Ord type based on original type. Generate map primitives with Ord

amosr:
Don't generate Map primitives with arrays as the keys.
#625 

! @jystic @tranma 
/jury approved @jystic


tranma:
> Doubles are not 'Ord' because NaN

YOUR MATHEMATICS IS POWERLESS HERE

amosr:
I had disabled generating `PrimWindow` because I saw a test failure involving it, but I've re-enabled it and ran several tests, as well as looking at the conversion, and it seems to be fine. The failure I saw was probably NaN or Ord and just happened to also contain PrimWindow.

jystic:
☝️ 👆

PR #634, opened 2017-08-14
show original query in Core check error

amosr:
Should help debug the Core check error
! @jystic @tranma 
/jury approved @jystic


jystic:
🚀🎸

PR #635, opened 2017-08-15
Remove entities from output which have no facts before the snapshot time

jystic:
! @amosr @tranma 
/jury approved @amosr

### DIFF
--- a/icicle-compiler/icicle.cabal
+++ b/icicle-compiler/icicle.cabal
@@ -154,6 +154,7 @@ library
                        Icicle.Runtime.Data.Array
                        Icicle.Runtime.Data.IO
                        Icicle.Runtime.Data.Logical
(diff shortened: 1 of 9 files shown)


amosr:
scattered segment descriptors!

it's a bit of a shame that for a nested array output, it's O(sum nested elements) rather than O(outer n), regardless of how many mask bits are set. probably not a real issue since the output is so much smaller than the input.

amosr:
👍

jystic:
> it's a bit of a shame that for a nested array output, it's O(sum nested elements) rather than O(outer n), regardless of how many mask bits are set. probably not a real issue since the output is so much smaller than the input.

Yeah I agree, it does bug me, but I decided to just go for the simple approach and see if it causes any issues.

Issue #586, opened 2017-07-04
Array index is unsafe


amosr:
At the moment in Source,
```
index :: Array a -> Int -> a
```
but this should be
```
index :: Array a -> Int -> Possibly a
```
and the conversion to Core should check that it's a valid index, similar to `map_insert`.

PR #637, opened 2017-08-15
Safe array indexing

amosr:
#586 
New exception tags: `NotANumber` and `IndexOutOfBounds`.

Make `index` return Possibly in Source. This is converted to a straightforward safe indexing in Core, which returns `Sum Error t`. Having safe indexing in Core means we can generate random programs using that. Then when we flatten, it introduces explicit bounds checks before converting to unsafe indexing.

The array and map primitives in Flatten are using unsafe indexing, so this won't affect them.

! @jystic @tranma 
/jury approved @jystic

### DIFF
--- a/data/sea/00-includes.h
+++ b/data/sea/00-includes.h
@@ -35,6 +35,8 @@ static const ierror_t ierror_not_an_error           = 0;
 static const ierror_t ierror_tombstone              = 1;
 static const ierror_t ierror_fold1_no_value         = 2;
 static const ierror_t ierror_cannot_compute         = 3;
+static const ierror_t ierror_not_a_number           = 4;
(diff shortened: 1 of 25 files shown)


tranma:
very good, it's long overdue 😁

amosr:
I'm going to hold off merging this one until we've updated payload to master. Have tested on one production dictionary but would prefer to test everywhere, since it does affect the front-end language.

jystic:
👍

amosr:
I don't entirely remember why I said above I would wait until merging, but I think it was so this change didn't get confused with other changes, in the unlikely case it breaks an existing dictionary. this is no longer relevant, so I'll just merge

PR #639, opened 2017-08-17
Switch submodules to https for builds so we can build without auth

jystic:
All of icicle's dependencies are now open source.

! @amosr @tranma 
/jury approved @amosr


tranma:
amazing
👍

amosr:
💯

PR #640, opened 2017-08-18
Fix days_in calculation

jystic:
It was giving the wrong answer before in many situations.

! @tranma @amosr


tranma:
right, it was just plain wrong before?

tranma:
ok
👍

PR #643, opened 2017-08-29
Relax C-string constraint: allow null in strings

amosr:
Some input data contains nulls, so we can't outlaw it. Eventually it
would be nice to support these, but for now we can just copy it over.
The evaluator will only see the prefix of the string up to the first
null.

! @jystic @tranma 
/jury approved @jystic


amosr:
Ok, it passes tests here and a couple of times locally. I think all the generators must be using muppets, so no NUL in strings.

jystic:
👍 Looks good, how sad :(

PR #645, opened 2017-08-30
trigonometry

amosr:
I'm tempted to add modulus, but I'm not sure whether int and double modulus are the same polymorphic operation, or if it should be `mod` and `fmod`

! @jystic @tranma


data/sea/20-simple.h
```diff
@@ -31,6 +31,7 @@ static idouble_t INLINE iint_extend   (iint_t    x)              { return x; }
 static iint_t    INLINE iint_neg      (iint_t    x)              { return -x; }
 static iint_t    INLINE iint_add      (iint_t    x, iint_t    y) { return x + y; }
 static iint_t    INLINE iint_sub      (iint_t    x, iint_t    y) { return x - y; }
+static iint_t    INLINE iint_mod      (iint_t    x, iint_t    y) { return x % y; }
```
jystic: looks like you did add modulus?

amosr: only to the backend yet. heh.

amosr:
I'll just add some sanity tests in a sec

data/sea/20-simple.h
jystic: what's the difference in semantics between the integer and double variants?

amosr:
interesting tidbit: ages ago, when classifying double operations into whether they produced nans or not, atan, atan2, cos, sin, tan and tanh are all "clean". it feels more consistent to just say all double operations are possibly, though. any preference?

amosr:
I'm not confident enough in my double testing code to say for sure that all tans are valid. I believe that `tan (pi/2) = sin (pi/2) / cos (pi/2) = 1 / 0`, so it *should* be able to produce infinity, even though for me `cos (pi/2) = 6e-17`.

I don't think we lose anything by making them Possibly, so I'll keep it.

amosr:
I realised integer modulo is a somewhat larger job, because it would need to be in Possibly (1 `mod` 0). the machinery for this is already in place for doubles, but would be a bit different for integers.

so, given we can already encode modulo clumsily, I will leave both modulo operations for later

amosr:
going to hold off on merging this, because if anybody is using any of these as variable names it will break. I had a look but couldn't see any.

PR #646, opened 2017-10-19
Add separated parsers for QueryTop

HuwCampbell:
! @jystic @tranma 

Had a chat with @don41382 this morning regarding parsing
icicle expressions outside of icicle dictionaries (such as in a
web UI). There was a struggle with source locations due to
`feature blah ~>` being artificially prepended before the parse
and type check.

I pointed him to the fact that the parser can actually parse
these separately. But adding these two functions will make it
much easier in that it can be done without any knowledge of,
or imports from Parsec.

@tranma  @charleso  as discussed.
/jury approved @amosr

### DIFF
--- a/icicle-source/src/Icicle/Source/Parser.hs
+++ b/icicle-source/src/Icicle/Source/Parser.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 module Icicle.Source.Parser (
     parseQueryTop
+  , parseQuery
+  , parseTop
   , parseFunctions
   , prettyParse
   , ParseError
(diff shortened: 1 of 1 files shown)


icicle-source/src/Icicle/Source/Parser.hs
```diff
@@ -37,6 +39,17 @@ parseQueryTop name inp
  = let toks = lexer "" inp
    in  runParser (consumeAll $ top name) () "" toks
 
+parseTop :: Text -> Either ParseError UnresolvedInputId
+parseTop inp
```
amosr: can we call this "parseFactName" or "parseInputId" instead? the fact name is what makes it a top-level query, but the name isn't the top-level in itself

amosr:
👍 good idea

icicle-source/src/Icicle/Source/Parser.hs
amosr: `QueryTop` is probably a bad name too, maybe we should call it something else. shrug.

HuwCampbell:
No worries `parseFactName` it is.

amosr:
I'm guessing you can't merge, so I will.

PR #650, opened 2017-10-30
repl cflags: plumbing

amosr:
tedious
/jury approved @jystic

### DIFF
--- a/icicle-compiler/src/Icicle/Repl.hs
+++ b/icicle-compiler/src/Icicle/Repl.hs
@@ -94,7 +94,7 @@ getInitialState = do
               UseColor ->
                 [FlagColor]
 
-      , stateCFlags = "-O0 -DICICLE_ASSERT=0"
+      , stateCFlags = "-O0 -UICICLE_ASSERT"
       }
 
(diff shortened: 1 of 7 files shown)


icicle-compiler/src/Icicle/Repl.hs
```diff
@@ -93,6 +93,8 @@ getInitialState = do
                 []
               UseColor ->
                 [FlagColor]
+
+      , stateCFlags = "-O0 -UICICLE_ASSERT"
```
jystic: Could this be `-DICICLE_ASSERT=0` instead?

jystic:
👍 looks good

icicle-compiler/src/Icicle/Repl.hs
amosr: nah, gcc 4.7 barfs if you pass the same identifier multiple times as arguments like "-DA=1 -DA=2"

jystic: wow, that's delightful

amosr: real helpful

PR #651, opened 2017-10-31
simplify anormal pass

amosr:
/jury approved @jystic

### DIFF
--- a/icicle-data/src/Icicle/Common/Exp/Simp/ANormal.hs
+++ b/icicle-data/src/Icicle/Common/Exp/Simp/ANormal.hs
@@ -60,23 +60,25 @@ anormalAllVars :: (Hashable n, Eq n) => a -> Exp (Ann a n) n p -> Fresh n (Exp (
 anormalAllVars a_fresh xx
  = do   (bs, x)  <- pullSubExps a_fresh xx
 
-        (ret,_) <- foldM insertBinding (x,Set.empty) (reverse bs)
+        (ret,_) <- foldM insertBinding (x, snd $ annotOfExp x) (reverse bs)
         return ret
 
  where
   insertBinding (x,bs) (n,b)
-   | n `Set.member` snd (annotOfExp x)
+   | n `Set.member` bs
    = do n' <- fresh
-        let u = Set.union (snd $ annotOfExp x) (snd $ annotOfExp b)
-        let a' = (a_fresh, u)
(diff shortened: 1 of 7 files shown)


jystic:
Seems reasonable, what's the story here?

amosr:
the conversion to ANF is renaming way more variables than it needs to, and we should be able to take advantage of the fact that the input program never has shadowing, in order to limit renaming variables.

but this implementation is wrong, I just found a case which fails:

```
let a = (let b = 0 in b + 1)
\b. a + b

==>

let b = 0
let a = b + 1
\b. a + b
```

amosr:
actually, it wasn't the renaming that was taking so long, it was the fact that the bookkeeping for what to rename was done in a separate pass, as well as converting a list to a set on every iteration.

amosr:
A fact with 200 int fields
```
feature ints200 ~> group a1 ~> latest 1 ~> fields


45.2s -> 3.44s
```
!!!

jystic:
> 45.2s -> 3.44s

Wow!!

amosr:
yeah, it's crazy!

icicle-data/src/Icicle/Common/Exp/Simp/ANormal.hs
```diff
@@ -59,65 +59,26 @@ anormal a_fresh xx
...
-  -- If the same name is used in multiple places, we'll introduce shadowing, which is
-  -- outlawed by the type checker.
-  --
-  -- So, we look at each binding, and if it would shadow something, we rename it.
-  renames _ [] seen  x
-   = return (seen, x)
-
-  renames allNames ((n,b):bs) seen x
-   -- Check if n is used anywhere else, including as a later binding name
-   |  n `Set.member` (allNames <> Set.fromList (fmap fst bs))
```
amosr: I think this was a large part of it - reconstructing this set every iteration, just to check membership. I wonder whether you could write a rewrite rule: `member n (a <> b) = member n a || member n b`

jystic:
I can only imagine how bad it is for a 500 field struct

amosr:
let's find out!

amosr:
918.87s -> 31.56s

tranma:
unreal

jystic:
Winning, going to make a huge difference

jystic:
:100:

PR #653, opened 2017-11-01
Flatten/Constructor: don't duplicate

amosr:
A while ago I noticed some duplicate work in the flattened avalanche:

```
feature salary ~> fold y = 1 : 1 ~> y / 0

Avalanche
----------
y-conv-4 =r acc-y-conv-4
anf-1 = div# y-conv-4 0.0
anf-0 = doubleIsValid# anf-1
flat-0 =i Left ExceptNotAnError : Sum Error Double

if (anf-0)
{
    flat-0 =w right# anf-1
}

Flattened, simplified
--------------------
y-conv-4 =r acc-y-conv-4
anf-1 = div# y-conv-4 0.0
anf-0 = doubleIsValid# anf-1
flat-0-simpflat-0 =i ExceptNotAnError : Error
flat-0-simpflat-1 =i 0.0 : Double

if (anf-0)
{
    flat-0-simpflat-0 =w ExceptNotAnError
    flat-0-simpflat-1 =w div# y-conv-4 0.0
}
```

The `div#` gets duplicated into two places! The constructor transform is putting every let-binding it sees in the environment, and when trying to melt `right# anf-0` looks up the arguments in the environment, so it ends up inlining the definition of `anf-0`.

I think it only needs to inline things which are going to cause a melt/unmelt transform to fire: so values, melt prims, unmelt prims, as well as variables which might point to any of those. Is that sufficient?

### DIFF
--- a/icicle-compiler/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
+++ b/icicle-compiler/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
@@ -129,36 +129,41 @@ constructor a_fresh statements
       -> FixT (Fresh n) ( ExpEnv    (Ann a n) n Prim
                         , Statement (Ann a n) n Prim )
   goS env s
-   = let !env'  = updateExpEnv s env
-         ret s' = return (updateExpEnv s' env', s')
+   = let ret s' = return (env, s')
 
          goWith x s'
-          = do x' <- goX env' x
+          = do x' <- goX env x
                ret $! s' x'
 
      in  case s of
           If x t e
            -> goWith x $ \x' -> If x' t e
           Let n x ss
-           -> goWith x $ \x' -> Let n x' ss
+           -> do x' <- goX env x
+                 let !env'
+                        | pertinent x
+                        = Map.insert n x' env
+                        | otherwise
+                        = env
+                 return (env', Let n x' ss)
+
           ForeachInts t n from to ss
-           -> do from' <- goX env' from
-                 to'   <- goX env' to
+           -> do from' <- goX env from
+                 to'   <- goX env to
                  ret $ ForeachInts t n from' to' ss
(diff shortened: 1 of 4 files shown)


icicle-compiler/test/cli/repl/t30-sea/expected
```diff
@@ -1082,8 +1057,7 @@ if (eq#
         if (doubleIsValid# conv-171)
         {
             flat-118-simpflat-291 =w ExceptNotAnError
-            flat-118-simpflat-292 =w sqrt#
-                                       flat-110-simpflat-288
+            flat-118-simpflat-292 =w conv-171
```
amosr: this is an important one - duplicating `sqrt`

amosr:
A lot of the duplicated work is just floating point operations. So it's not going to make a big difference in runtime. Perhaps even slightly slower, depending on register pressure. But making the environment smaller should speed up `constructor` a bit, as well as producing smaller output code.

amosr:
! @jystic @tranma 
forgot jury

icicle-compiler/src/Icicle/Avalanche/Statement/Simp/Constructor.hs
```diff
@@ -446,6 +451,19 @@ constructor a_fresh statements
...
+  pertinent x
+   | XValue{} <- x
+   = True
+   | XVar{} <- x
+   = True
+   | Just (PrimMelt{},_) <- takePrimApps x
+   = True
+   | otherwise
+   = False
+
```
tranma: 👍

tranma:
ok, this made sense.
👍

icicle-compiler/test/cli/repl/t30-sea/expected
jystic: :savage:

PR #654, opened 2017-11-01
a big bucket of small optimisations

amosr:
1. A while ago I changed `meltType` to return a "logical melt" which distinguishes between ints as values and ints as tags (as well as errors as tags). This is because relation primops need to sort them differently. Turns out that most of the uses of `meltType` don't care and convert straight back to the representation type, so it's silly to construct the intermediate logical type. So I hand-specialised the function to generate the types.

2. simpEvalX was trying to deconstruct and take the arguments of a primitive application at every point. If you had a multiple arguments you would end up trying `takePrimApps`, then recursing down into the function half, finding another application, and trying to `takePrimApps` again. Even though the previous one failed. So instead, introduce a function for recursing into applications, which takes the list of arguments you've already seen. Then when you get to the bottom, if it's a primitive you can use the argument list you've built and simplify. Otherwise just rebuild the application.

```
feature ints500 ~> group a1 ~> latest 1 ~> fields

Flatten Avalanche only:
30.07s -> 25.54s
```

! @jystic @tranma

### DIFF
--- a/icicle-compiler/src/Icicle/Avalanche/Prim/Flat.hs
+++ b/icicle-compiler/src/Icicle/Avalanche/Prim/Flat.hs
@@ -228,8 +228,8 @@ meltLogical t
 
     OptionT a   -> rep BoolT <> meltLogical a
 
-    ArrayT a -> nested ArrayT   (meltLogical a)
-    BufT i a -> nested (BufT i) (meltLogical a)
+    ArrayT a -> nested ArrayT   (meltType a)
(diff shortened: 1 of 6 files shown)


amosr:
hold on

amosr:
I accidentally pushed this to master, but I was going to merge it anyway. so I'm going to leave it as-is rather than bumblingly trying to fix it.

amosr:
yolo

PR #655, opened 2017-11-02
Time test

amosr:
Can't properly test primitives for tempus? Fudge it. The problem with time primitives such as "minus months" is it doesn't take many months to overflow. So the number of months has to be small. If we use an arbitrary int expression it could be anything. The trick here is to only generate a *constant* number of months, between zero and thirteen.

The other problem is that "minus days" was throwing away the time of day, and the tests for that weren't catching it because they were only generating days without time. I have modified that test to generate real time of day, but not other tests, because some of them (eg roundtripping date) break on times.

! @jystic @tranma

### DIFF
--- a/data/sea/21-time.h
+++ b/data/sea/21-time.h
@@ -67,7 +67,7 @@ static iint_t INLINE itime_to_epoch_seconds (itime_t x)
   return s + 60*n + 3600*h + (365*y + y/4 - y/100 + y/400 + (m*306 + 5)/10 + (d - 1)) * 86400;
 }
 
-static itime_t INLINE itime_from_epoch_days (iint_t g)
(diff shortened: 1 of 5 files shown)


amosr:
The `itime_minus_days` fix: perhaps the right way is to modify `itime_to_epoch_days` to return a 4-tuple of day, hours, minutes, seconds. But, `itime_to_gregorian` seems cheap enough that it's unlikely to cost more even if it doesn't get CSEd. Is it nicer?

amosr:
@jystic I'd like your eyes on the `itime_minus_days` implementation, just in case. Dates scare me, and it's probably wrong for DST but not sure what to do instead. You can merge this whenever you like.

PR #656, opened 2017-11-28
needed to update megaparsec for corba

don41382:
!@charleso @nhibberd @jystic 

update for corba was required

see
> https://github.com/ambiata/technology/issues/552
/jury approved @jystic

### DIFF
--- a/icicle-compiler/icicle.cabal
+++ b/icicle-compiler/icicle.cabal
@@ -88,7 +88,7 @@ library
                      -- a bit cleaner.
                      -- As an added benefit, parsec will give us nicer error messages.
                      , attoparsec                      >= 0.12       && < 0.14
-                     , megaparsec                      == 5.0.*
+                     , megaparsec                      >= 5.0        && < 5.2
                      ,     parsec                      == 3.1.*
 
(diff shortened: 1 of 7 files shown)


icicle-compiler/icicle.cabal
```diff
@@ -88,7 +88,7 @@ library
                      -- a bit cleaner.
                      -- As an added benefit, parsec will give us nicer error messages.
                      , attoparsec                      >= 0.12       && < 0.14
-                     , megaparsec                      == 5.0.*
+                     , megaparsec                      == 5.1.*
```
jystic: Can we just increase the upper bound here? bumping the lower bound could break things that depend on icicle.

```megaparsec                      >= 5.0        && < 5.2```

don41382: I'll do that

don41382:
@jystic are you fine with this?

jystic:
👍 yep, all good now

etorreborre:
Can I ask what is `corba`? This doesn't seem to be a public Ambiata project and the name brings some painful memories :-).

PR #658, opened 2018-01-31
Add a icicle check command

HuwCampbell:
@charleso as mentioned, this could help us stay up to date in CI.
/jury approved @erikd-ambiata

### DIFF
--- a/icicle-compiler/src/Icicle/Command/Compile.hs
+++ b/icicle-compiler/src/Icicle/Command/Compile.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 module Icicle.Command.Compile (
     Compile(..)
+  , Check(..)
 
   , Fingerprint(..)
   , MaximumQueriesPerKernel(..)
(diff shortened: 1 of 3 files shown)


icicle-compiler/src/Icicle/Command/Compile.hs
```diff
@@ -150,3 +158,18 @@ icicleCompile compile = do
...
+----------------
+
+icicleCheck :: Check -> EitherT CompileError IO ()
+icicleCheck check = do
+  dictionary <- loadDictionary (checkInputDictionary check)
+  let
+    summary
+      = prettyDictionarySummary dictionary
+    render
+      = Text.pack (show (Pretty.pretty summary))
```
erikd-ambiata: Wat? Does that really need `show` *and* `Text.pack`?

HuwCampbell: Not really, I could just use `print` on the output of Pretty.pretty.

I did this to keep in line with the error message rendering higher up in the file.

HuwCampbell: I'll change it to putDoc.

erikd-ambiata: What is the type of `Pretty.pretty`? We usually try to avoid `Show` instances for anything other than printing stuff in GHCi and because `Show` instances are required in quickcheck properties.

erikd-ambiata:
👍

PR #659, opened 2018-01-31
Add a icicle check command

HuwCampbell:
Same as #658, but with builds

! @charleso 
/jury approved @charleso


charleso:
👍 

Thanks Huw.

PR #660, opened 2018-02-23
Use pattern guards for `filter` in sorbet.

HuwCampbell:
A common use case has been
```
filter
  is_some x
in
  let
    x' = box x
  in
    mean x'
```
which isn't very pleasant. I see no reason why we shouldn't write
```
filter
  Some x' <- x
in
  mean x'
```


justanotherdot:
So `sorbet` had some work done it already?

HuwCampbell:
The `bnf` and concrete parser have been written.

PR #661, opened 2018-03-12
Topic/readme

HuwCampbell:
I think it's time this was updated.
/jury approved @olorin


francoisfernando:
Nice work @HuwCampbell Like the inclusion of some documentation around streaming variance

olorin:
:+1: Nice.

